PR #26, opened 2022-09-14
pass config for projection load count

unixslayer:


dgafka:
Custom configuration released with `1.42.1`.
Read more [here](https://docs.ecotone.tech/modelling/event-sourcing/executing-and-managing/running-projections#custom-configuration)

PR #29, opened 2022-09-23
projection setup configuration defined in attribute gets overwritten

unixslayer:
projection setup configuration defined in attribute gets overwritten with default values from projection running configuration


dgafka:
:pray:

PR #35, opened 2022-10-03
all projection options can be passed using ProjectionRunningConfiguration

unixslayer:
all projection options can be passed using ProjectionRunningConfiguration
however, options set by attribute with projection class are overwritten


dgafka (approved):
All good, thanks :)

PR #36, opened 2022-10-05
allow `ProjectionRunningConfiguration` to overwrite any setup options

unixslayer:


packages/Ecotone/src/Modelling/SaveAggregateService.php
```diff
@@ -120,7 +120,6 @@ public function save(Message $message, array $metadata): Message
                             : [];
         $aggregateIds = $aggregateIds ?: $this->getAggregateIds($aggregateIds, $aggregate, $this->isEventSourced);
 
-        unset($metadata[MessageHeaders::REPLY_CHANNEL], );
```
dgafka, line 123: If we will republish this event or distribute it outside, it may create situation, where after handling message we will want to reply it back. And this reply channel is valid only within saving flow.

Why we want to keep it?

packages/PdoEventSourcing/src/ProjectionRunningConfiguration.php
```diff
@@ -12,28 +12,29 @@ class ProjectionRunningConfiguration
     public const OPTION_INITIALIZE_ON_STARTUP = 'initializeOnStartup';
     public const DEFAULT_INITIALIZE_ON_STARTUP = true;
 
-    public const OPTION_AMOUNT_OF_CACHED_STREAM_NAMES = 'amountOfCachedStreamNames';
+    public const OPTION_AMOUNT_OF_CACHED_STREAM_NAMES = 'cache_size';
```
dgafka, line 15: Thanks for cleaning this up :)

packages/Ecotone/src/Modelling/SaveAggregateService.php
unixslayer: we do not want to keep it however, this unset is duplicated on line 134 and `phpstan` is failing there. I removed this one so unset of `MessageHeaders` will be grouped together for cleaner reading.

dgafka (approved):

dgafka:
Thanks :) 
Will be released with `1.44.3`.

PR #39, opened 2022-10-14
Add array object converter for mixed scalar array

dgafka:


dgafka:
https://docs.ecotone.tech/modules/jms-converter#array-deserialization

dgafka:
To be released with `1.46.1`

PR #51, opened 2022-11-03
move revision header setup to SaveAggregateService

unixslayer:


packages/Ecotone/src/Modelling/SaveAggregateService.php
```diff
@@ -133,6 +134,8 @@ public function save(Message $message, array $metadata): Message
         unset($metadata[MessageHeaders::REPLY_CHANNEL]);
         unset($metadata[MessageHeaders::CONTENT_TYPE]);
 
+        $metadata[MessageHeaders::REVISION] = $this->getRevision($message);
```
dgafka, line 137: We are passing $message here, which is `Ecotone\Messaging\Message`.
Whatever is returned from aggregate's command handlers will be in `payload` of the `Message`.
So in case of `Event Sourcing Aggregate`, it will be `array of events`.

In that case we need to resolve the revision for each of the event which is in the payload.

packages/Ecotone/src/Modelling/SaveAggregateService.php
```diff
@@ -133,6 +134,8 @@ public function save(Message $message, array $metadata): Message
...
+
```
dgafka, line 138: [Current interface for Event Sourced Aggregate](https://github.com/ecotoneframework/ecotone-dev/pull/51/files#diff-da60ae3b5e86b2cf0bd311293b808582191e7b831c1d6b50763a3ed94b6ce85cL154) take all the `events` as one parameter and `metadata` as another.

This means that in case metadata differs we will need to call it more than one time. :) 
Otherwise if metadata is the same, let's stick with single call, as `each call to the repository is separate SQL` at the end.

dgafka: Let me know, if you need help if tests to cover that :)

dgafka:
Looks good, I will add some test for SaveAggregateService. 
Thanks :)

PR #56, opened 2022-11-24
projection status IDLE is for situation when projection awaits for new events

unixslayer:
without being handled separately resetting projection will stop after second run if load limit is set

### DIFF
--- a/packages/PdoEventSourcing/src/Config/InboundChannelAdapter/ProjectionEventHandler.php
+++ b/packages/PdoEventSourcing/src/Config/InboundChannelAdapter/ProjectionEventHandler.php
@@ -59,11 +59,12 @@ class ProjectionEventHandler
             $messagingEntrypoint->send([], $this->projectionSetupConfiguration->getProjectionLifeCycleConfiguration()->getDeleteRequestChannel());
         }
 
-        do {
-            $this->lazyProophProjectionManager->run($this->projectionSetupConfiguration->getProjectionName(), $this->projectionSetupConfiguration->getProjectionStreamSource(), $projectionExecutor, array_keys($this->projectionSetupConfiguration->getProjectionEventHandlerConfigurations()), $this->projectionSetupConfiguration->getProjectionOptions());
-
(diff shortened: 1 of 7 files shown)


unixslayer:
@dgafka projection will be set to IDLE right after lock is released. additional information should be added into documentation that limiting load count won't work for sync/polling projections by now. if any, this should be done internally inside of Prooph components. I will look if something can be done in that manner so maybe this is still good idea to have IDLE in Ecotone. Please decide whether close this PR or not.

packages/PdoEventSourcing/src/Config/InboundChannelAdapter/ProjectionEventHandler.php
```diff
@@ -59,12 +59,11 @@ public function execute(MessagingEntrypointWithHeadersPropagation $messagingEntr
...
-        $this->lazyProophProjectionManager->run($this->projectionSetupConfiguration->getProjectionName(), $this->projectionSetupConfiguration->getProjectionStreamSource(), $projectionExecutor, array_keys($this->projectionSetupConfiguration->getProjectionEventHandlerConfigurations()), $this->projectionSetupConfiguration->getProjectionOptions());
-        while ($status->getStatus() === ProjectionStatus::REBUILDING) {
-            $status = $this->lazyProophProjectionManager->getProjectionStatus($this->projectionSetupConfiguration->getProjectionName());
-
+        do {
             $this->lazyProophProjectionManager->run($this->projectionSetupConfiguration->getProjectionName(), $this->projectionSetupConfiguration->getProjectionStreamSource(), $projectionExecutor, array_keys($this->projectionSetupConfiguration->getProjectionEventHandlerConfigurations()), $this->projectionSetupConfiguration->getProjectionOptions());
-        }
+
+            $status = $this->lazyProophProjectionManager->getProjectionStatus($this->projectionSetupConfiguration->getProjectionName());
+        } while (in_array($status->getStatus(), [ProjectionStatus::REBUILDING, ProjectionStatus::RUNNING], true));
```
dgafka, line 66: `test_running_in_memory_based_projection_twice_with_reset` test is using `synchronous projection fixture`  for running test with In Memory implementation of Event Store and Projection Manager.

The difference is that before it was executed at least once, no matter of the status.
Rigth now with `do while` it seems to be never executed (the status is different?) and the projection is not created.

unixslayer: with `do...while` nothing changes and it is still run at least one.

dgafka:
Amazing work, thanks :)

PR #59, opened 2022-11-30
triggering projection

unixslayer:


packages/PdoEventSourcing/src/Config/EventSourcingModule.php
```diff
@@ -72,6 +72,7 @@ class EventSourcingModule extends NoExternalConfigurationModule
     public const ECOTONE_ES_RESET_PROJECTION = 'ecotone:es:reset-projection';
     public const ECOTONE_ES_DELETE_PROJECTION = 'ecotone:es:delete-projection';
     public const ECOTONE_ES_INITIALIZE_PROJECTION = 'ecotone:es:initialize-projection';
+    public const ECOTONE_ES_RUN_PROJECTION = 'ecotone:es:run-projection';
```
dgafka, line 75: This is not really running the projection, as projection may be running in async consumer :) 
Let's change name to `ecotone:es:trigger-projection`

dgafka:
Will be released with `1.58.0`, thanks :)

PR #76, opened 2023-01-20
redis integration

unixslayer:


packages/Redis/composer.json
```diff
@@ -0,0 +1,68 @@
...
+    "autoload-dev": {
+        "psr-4": {
+            "Test\\Ecotone\\Redis\\": [
+                "tests"
+            ]
+        }
+    },
+    "require": {
+        "ecotone/ecotone": "^1.68",
+        "enqueue/redis": "^0.10.9"
```
dgafka, line 31: This [fails](https://github.com/ecotoneframework/ecotone-dev/actions/runs/3969565653/jobs/6804237214) as `ecotone/enqueue` is missing. 
This is only visible in split testing (which test each package seperately), as monorepo has all the dependencies

packages/Redis/tests/Integration/RedisBackedMessageChannelTest.php
```diff
@@ -0,0 +1,44 @@
...
+    public function test_sending_and_receiving_message(): void
+    {
+        $queueName = Uuid::uuid4()->toString();
+        $messagePayload = Uuid::uuid4()->toString();
+
+        $ecotoneLite = EcotoneLite::bootstrapForTesting(
+            containerOrAvailableServices: [
+                RedisConnectionFactory::class => $this->getConnectionFactory(),
+            ],
+            configuration: ServiceConfiguration::createWithDefaults()
```
dgafka, line 26: If you take a look on stack trace of error from loading [DbalConnectionFactory](https://github.com/ecotoneframework/ecotone-dev/actions/runs/3969565655/jobs/6804231787), it shows that Dbal module was loaded. 

We want only Redis Module to load up, that's why we need to add:

```
   ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::SQS_PACKAGE]))
```

dgafka: This was missing in docs, docs are now updated :)

dgafka (commented):
Looks good. I've added information why errors have been thrown :) 

Can you add one more test, that was included recently to the documentation?
This scenario tests running given Message Channel with Message Handler, to test out most common flow:
https://docs.ecotone.tech/messaging/contributing-to-ecotone/demo-integration-with-sqs/inbound-and-outbound-adapters-and-message-channel#message-channel-with-consumer

PR #81, opened 2023-02-04
Reactive systems

dgafka:
- introduce instant retries
- allow deduplication on command bus
- add quick start refactor example to reactive system


dgafka:
https://docs.ecotone.tech/modelling/error-handling#command-bus-instant-retries

https://docs.ecotone.tech/modules/dbal-support#deduplication-idempotent-consumer

PR #85, opened 2023-02-13
Making use of Event (with metadata)

unixslayer:
- wrapping aggregate events with Event resolves issue with MessageId where saved events had different MessageId than one published and in case of having multiple events published, they all had the same MessageId

### DIFF
--- a/packages/Ecotone/src/Modelling/SaveAggregateService.php
+++ b/packages/Ecotone/src/Modelling/SaveAggregateService.php
@@ -19,6 +19,7 @@ use Ecotone\Messaging\Support\InvalidArgumentException;
 use Ecotone\Messaging\Support\MessageBuilder;
 use Ecotone\Modelling\Attribute\NamedEvent;
 use Ecotone\Modelling\Config\BusModule;
+use Ramsey\Uuid\Uuid;
 
 /**
  * Class SaveAggregateService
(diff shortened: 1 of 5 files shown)


packages/Ecotone/src/Modelling/SaveAggregateService.php
```diff
@@ -76,6 +77,19 @@ public function __construct(
...
+        unset($metadata[AggregateMessage::AGGREGATE_OBJECT]);
+        unset($metadata[AggregateMessage::TARGET_VERSION]);
+        unset($metadata[BusModule::COMMAND_CHANNEL_NAME_BY_NAME]);
+        unset($metadata[BusModule::COMMAND_CHANNEL_NAME_BY_OBJECT]);
+        unset($metadata[BusModule::EVENT_CHANNEL_NAME_BY_NAME]);
+        unset($metadata[BusModule::EVENT_CHANNEL_NAME_BY_OBJECT]);
+        unset($metadata[BusModule::QUERY_CHANNEL_NAME_BY_NAME]);
+        unset($metadata[BusModule::QUERY_CHANNEL_NAME_BY_OBJECT]);
+        unset($metadata[MessageHeaders::REPLY_CHANNEL]);
+        unset($metadata[MessageHeaders::CONTENT_TYPE]);
```
dgafka, line 91: Can we extract `OutboundMessageConverter::unsetEnqueueMetadata`, to `MessageHeaders` and reuse it too?
And use `DistributedMetadata::unsetDistributionKeys`, `MessageHeaders::unsetAsyncKeys($metadata)`, so things from `EventSourcingRepository#111`.

packages/PdoEventSourcing/src/EventSourcingRepository.php
```diff
@@ -108,27 +108,28 @@ public function save(array $identifiers, string $aggregateClassName, array $even
...
                 [
-                    MessageHeaders::MESSAGE_ID => Uuid::uuid4()->toString(),
                     LazyProophEventStore::AGGREGATE_ID => $aggregateId,
                     LazyProophEventStore::AGGREGATE_TYPE => $aggregateType,
                     LazyProophEventStore::AGGREGATE_VERSION => $versionBeforeHandling + $eventNumber,
                 ]
             );
 
-            $eventsWithMetadata[] = Event::create($event, $metadata);
+            $eventsWithMetadata[] = Event::create($event->getPayload(), $metadata);
```
dgafka, line 132: Maybe we could expose some method on the `Event::enrichMetadata: Event` to simplify code a bit :)

dgafka (approved):
So small code improvement proposals. 
If you're willing to add it, it would be awesome :)

dgafka:
Released with `1.73.0`, thanks :)

packages/PdoEventSourcing/src/EventSourcingRepository.php
unixslayer: Yes, I was thinking about adding some methods like `withMetadata` or `withAddedMetadata` to Event making it also immutable

PR #86, opened 2023-02-14
code improvements

unixslayer:
- code cleanup
- extracting metadata cleanup methods into MessageHeaders
- EventSourcedRepository will wrap incoming events with Event object in case it was not done already
- MessageHeaders::$headers are always an array
- Event can have added or replaced metadata using `withAddedMetadata` and `withMetadata` keeping it immutable
- Event should not allow to set its ID from outside due it may cause multiple events having the same ID in some cases


packages/Ecotone/src/Messaging/MessageHeaders.php
```diff
@@ -167,6 +169,62 @@ public static function unsetAsyncKeys(array $metadata): array
         return $metadata;
     }
 
+    public static function unsetEnqueueMetadata(array $metadata): array
+    {
+        unset(
+            $metadata[self::DELIVERY_DELAY],
+            $metadata[self::TIME_TO_LIVE],
+            $metadata[self::CONTENT_TYPE],
+            $metadata[self::CONSUMER_ACK_HEADER_LOCATION],
```
dgafka, line 178: When you unset this first then line #184 will not be able to find it. 
We can move it after and tests should pass :)

unixslayer: You're right. I somehow missed that.

dgafka:
Looks good, thanks :)

Issue #89, opened 2023-03-03
Inheritance support for Aggregates


CipherdevNL:
**Ecotone version(s) affected**: 1.74

**Description**  
For a project we've some processes/ event flows that are similar between two different concepts. To prevent drift in the process (e.g. AggregateA behaves different than AggregateB) we wanted to use inheritance on those aggregates.

This way we can keep the core data/ flow of those processes the same and reuse the handlers and extend the behaviour inside the child aggregates. 

In the first try we encountered the following error:
```
$ bin/console c:c

Channel name `Project\CommandX` should be unique, but is used in multiple handlers: Project\AggregateA:handleX Project\AggregateB:handleX Project\BaseAggregate:handleX
```

After digging through the code we found the following attribute `#[NotUniqueHandler]` that was able to solve this error, but after applying it another error popped up:

```
$ bin/console c:c

You have requested a non-existent service "Project\BaseAggregate".
```

Most likely this problem occurs as Ecotone tries to create a channel for the handler it found on the abstract base class. But failed as such classes are not existing in the service container. By manually patching some central place in the code I was able to get rit of the message and have a working inheritance setup.

The patch that we applied is to filter out all the abstract classes from the result that the `FileSystemAnnotationFinder.php` is giving. But I'm aware that this patch is far from ideal as it's leaking Ecotone concerns into the AnnotationFinder part of the framework. But I'm not that familiar with the codebase to know which patches must be applied inside the Ecotone part to exclude non-constructible classes from getting into the CommandHandler preparation part.

This is the dirty patch in `FileSystemAnnotationFinder.php` I created so far to continue the development of the project:
```php
    private function init(string $rootProjectDir, array $namespacesToUse, string $catalogToLoad, AutoloadNamespaceParser $autoloadNamespaceParser, array $systemClassesToRegister, array $userClassesToRegister, bool $isRunningForTesting)
    {
        if (! $catalogToLoad && $namespacesToUse == [] && $userClassesToRegister == [] && ! $isRunningForTesting) {
            throw ConfigurationException::create('Loading catalog was turned off and no namespaces were provided. Please provide namespaces manually via configuration or turn on catalog loading. Read related Module section at https://docs.ecotone.tech');
        }

        $registeredClasses = array_merge($systemClassesToRegister, $userClassesToRegister);
        if ($catalogToLoad) {
            $registeredClasses = array_merge(
                $registeredClasses,
[...]

PR #90, opened 2023-03-05
Do not lookup abstract classes

dgafka:
In case annotation is put on abstract classes, it's looked up like any other class. 
As abstract classes can't be initialized on their own, they should be skipped, as they're meaningful only in case other class extends them.

This related to #89


dgafka:
Released with `1.74.1`

PR #95, opened 2023-03-27
Fix Symfony bundle factory definition for SymfonyExpressionEvaluationAdapter service

eKsiK:
While linting symfony container I'm getting error: `"Invalid service "expressionEvaluationService": method "Ecotone\Messaging\Handler\SymfonyExpressionEvaluationAdapter::createWithExternalExpressionLanguage()" does not exist."`

In fact, this method is not existing. This PR fixes the issue above.


dgafka:
Thank you for contribution. Released with `1.77.1` :)

PR #100, opened 2023-04-16
Subscription based amqp consumer

dgafka:
In order to limit resources used during consumption of messages from RabbitMQ and to speed up consumption process, this PR introduces subscription based consumption.

This also open possibility for using [RabbitMQ Streams](https://www.rabbitmq.com/streams.html)


packages/Amqp/src/AmqpQueue.php
```diff
@@ -46,6 +48,16 @@ public static function createWith(string $queueName): self
         return new self($queueName);
     }
 
+    public static function createStreamQueue(string $queueName): self
```
dgafka, line 51: Used only for testing purposes, not meant to be used in production, API may change.

PR #103, opened 2023-04-21
fix EventSourcingSaga attribute

jlabedo:
It is not possible to register an EventSourcingSaga as stated in the doc


jlabedo:
Hey @dgafka , I would be happy to add a test with some indications from on how to do it (unit, behat, ...)
I also thought of hooking into `InterfaceToCall` with a method like `hasCompatibleClassAnnotation`. What do you think ?

dgafka:
Hey @jlabedo,

Thanks for the fix. 
Yes, please add the high level test to provide it works. By high level test I mean using `Ecotone Lite`, this is exactly the same way you can write test in your application :).
  
You may add it to `ModellingEcotoneLiteTest` and take a look on tests there for reference :) 
In `tests/Modelling/Fixture/*`, you may add specific example.

jlabedo:
Hello @dgafka , a test has been added, let me know

packages/Ecotone/tests/Modelling/Unit/ModellingEcotoneLiteTest.php
```diff
@@ -116,4 +119,19 @@ public function test_calling_asynchronous_command_handler_with_combined_message_
...
+            [OrderDispatch::class, OrderDispatchRepository::class],
+            [new OrderDispatchRepository()],
+            ServiceConfiguration::createWithDefaults()
+                ->withSkippedModulePackageNames(ModulePackageList::allPackages())
+        );
+
+        $ecotoneLite->getEventBus()->publish(new OrderWasCreated("1"));
+        $this->assertEquals(
+            "1", "1"
+        );
```
dgafka, line 135: If Saga would not trigger in that case, the test would still pass :P 

Can you add [QueryHandler](https://docs.ecotone.tech/modelling/query-handling/handling-queries#name-routing) on Saga to expose the status:
```php
#[QueryHandler('order_saga.getStatus')]
public function getStatus(): string
{
  return $this->status;
}
```
 and after calling `EventBus` confirm the status :
    
```php
$this->assertTrue(
      'new',
      $ecotoneLite
          ->publishEvent(new OrderWasCreated('1'))
          ->sendQueryWithRouting('order_saga.getStatus', metadata: ['aggregate.id' => '1'])  
  );
```

dgafka: To make use of the test API that I suggested we need to do [above comment first](https://github.com/ecotoneframework/ecotone-dev/pull/103#discussion_r1185722406)

dgafka:
Thank you @jlabedo, I've added two comments :)

packages/Ecotone/tests/Modelling/Unit/ModellingEcotoneLiteTest.php
jlabedo: 😵
Corrected thanks !

dgafka:
Thanks, merged :)

Issue #104, opened 2023-04-21
Event sourced projection event is not correctly handled in complex flows


jlabedo:
**Ecotone version(s) affected**: 1.79.0

**Description**  
In the context of 2 event sourced aggregates, both aggregates having projections on their stream,
Then, with a synchronous flow being command (Agg1_**id1**)=>Event(Agg2_id1)=>Command(Agg1_**id1**), the events from Agg2 handled in its projection are not handled in the **context** of a projection handler, but as a "common" event handler.
This breaks the projection, as the projection state is not updated and persisted, and the projection is not initialized.
There is no bug in the case of different ids on Agg1: (Agg1_**id1**)=>Event(Agg2_id1)=>Command(Agg1_**id2**) works as expected

**How to reproduce**  
See modified EventSourcing example in:
https://github.com/jlabedo/ecotone-dev/tree/failing-es-projection/quickstart-examples/EventSourcing
The example show that `SagaProjection` is not initialized when we create a product with Id=3

**Context**
The case is pretty awkward, and we found it in a badly designed use case. We fixed it by redesigning correctly our aggregates

dgafka:
Hey,

Can you provide PR with test case (you may copy paste your example into `package/PdoEventSourcing/tests/Fixture/*`) using `Ecotone Lite` with support for testing [Event Sourced Application](https://docs.ecotone.tech/testing-messaging-architecture/testing-event-sourcing-applications#testing-aggregates-with-event-store)?

jlabedo:
Hello @unixslayer, I got another use case that can be even more problematic : if the event sourced aggregate records 2 events and both events are listened by the same projection, the projection will fail in the same manner. Hope that helps, I may try to add a test for that

unixslayer:
@jlabedo currently both with @dgafka we are working on an issue around metadata propagation in https://github.com/ecotoneframework/ecotone-dev/pull/134. This may be related to or have an impact on your case which I'll get back to as soon as we finish with that one.

dgafka:
This is now fixed with [1.87.0](https://github.com/ecotoneframework/ecotone-dev/releases/tag/1.87.0).

unixslayer:
[...]

PR #112, opened 2023-05-04
Add failing test with synchronous event sourced saga

jlabedo:
Test to reproduce #104 

The test is failing because the SagaProjection is not initialized. In fact, the event handler is executed as a "normal" event handler and not in the context of a "projection" handler


dgafka:
Amazing :) 

How important is this for you? From what I understood in the issue, you changed the flow, so no hurry with this one?

dgafka:
Hey @unixslayer :) 
Will you find time to take a look on this?

unixslayer:
> Hey @unixslayer :) 
> Will you find time to take a look on this?

Sure. I'll find some time after this weekend.

dgafka:
Great thanks @unixslayer :)

unixslayer:
Model fails in two different circumstances and setups.

In Behat test, initialization of SagaProjection is not triggered even it is configured and registered correctly. Although, I've noticed that `BasketWasCreated` gets handled twice: first time after it is recorded and second time when Saga is done which leasd to resetting basket and removing product in read model.

I've also written unit test using `Ecotone Test Support` and it shows that basket projection is not handling `ProductWasAddedToBasket` after Saga finish. However, here `SagaProjection` is initialized properly. 

@jlabedo if you grant me access to your fork I can push additional commit with changes I've made.

packages/Ecotone/src/Modelling/Config/BusRoutingModule.php
```diff
@@ -290,22 +290,10 @@ public static function getEventBusByNamesMapping(AnnotationFinder $annotationReg
...
-                $type = TypeDescriptor::create($chanelName);
-                if ($type->isUnionType()) {
-                    foreach ($type->getUnionTypes() as $type) {
-                        $namedEventHandlers[$type->toString()][] = $chanelName;
-                        $namedEventHandlers[$type->toString()]   = array_unique($namedEventHandlers[$type->toString()]);
-                    }
-                } else {
-                    $namedEventHandlers[$chanelName][] = $chanelName;
-                    $namedEventHandlers[$chanelName]   = array_unique($namedEventHandlers[$chanelName]);
-                }
```
dgafka, line 308: This is small improved to avoid registering event handler by class name routing by name, if not defined explicitly.

dgafka:
@jlabedo @unixslayer  This is solved by #139 :)

PR #117, opened 2023-05-16
message ID should not be overwritten.

unixslayer:
this happens when Event is loaded from existing stream and sent to projection

### DIFF
--- a/packages/Ecotone/src/Modelling/Event.php
+++ b/packages/Ecotone/src/Modelling/Event.php
@@ -19,10 +19,8 @@ class Event
         $this->eventType = $eventType;
         $this->payload = $payload;
 
-        $metadata[MessageHeaders::MESSAGE_ID] = Uuid::uuid4()->toString();
-        if (! array_key_exists(MessageHeaders::TIMESTAMP, $metadata)) {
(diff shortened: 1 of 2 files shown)


packages/Ecotone/src/Modelling/Event.php
```diff
@@ -19,10 +19,8 @@ private function __construct(string $eventType, array|object $payload, array $me
         $this->eventType = $eventType;
         $this->payload = $payload;
 
-        $metadata[MessageHeaders::MESSAGE_ID] = Uuid::uuid4()->toString();
-        if (! array_key_exists(MessageHeaders::TIMESTAMP, $metadata)) {
-            $metadata[MessageHeaders::TIMESTAMP] = (int)round(microtime(true));
-        }
+        $metadata[MessageHeaders::MESSAGE_ID] = $metadata[MessageHeaders::MESSAGE_ID] ?? Uuid::uuid4()->toString();
+        $metadata[MessageHeaders::TIMESTAMP] = $metadata[MessageHeaders::TIMESTAMP] ?? (int)round(microtime(true));
```
dgafka, line 23: Can we have test for this? 
It can be only for this particular class, to check if message id is not overwritten :)

PR #118, opened 2023-05-17
when aggregate is being saved, events gets the same message ID as message transporting aggregate

unixslayer:


docker-compose.yml
```diff
@@ -1,6 +1,33 @@
 version: "3.8"
 
 services:
+  app8_2:
```
dgafka, line 4: Can you bump up `app` service to php 8.2 instead?

quickstart-examples/EmittingEventsFromProjection/composer.json
```diff
@@ -9,7 +9,11 @@
     ],
     "autoload": {
         "psr-4": {
-            "App\\": "src/"
+            "App\\": "src/",
+            "Ecotone\\": "../../packages/Ecotone/src",
+            "Ecotone\\EventSourcing\\": "../../packages/PdoEventSourcing/src",
+            "Ecotone\\JMSConverter\\": "../../packages/JmsConverter/src",
+            "Ecotone\\Lite\\": "../../packages/LiteApplication/src"
```
dgafka, line 16: Why this change? :thinking:

packages/Amqp/composer.json
```diff
@@ -40,7 +40,7 @@
     "require": {
         "ext-amqp": "*",
         "ecotone/enqueue": "^1.82",
-        "enqueue/amqp-ext": "^0.10.4",
+        "enqueue/amqp-ext": "^0.10.18",
```
dgafka, line 43: :+1:

quickstart-examples/EmittingEventsFromProjection/composer.json
unixslayer: so `quickstart-examples` will run the current version of ecotone packages instead of latest one which is not the same. my changes from yesterday introduced a bug which was caught after releasing new version

PR #121, opened 2023-05-23
feat(Symfony): boot messaging system on service injection

jlabedo:
This PR will create the messaging system when injecting the messaging system and not on kernel boot anymore, improving performance when the messaging system is not necessary

Edit : I am waiting for #122 to be validated before continuing

### DIFF
--- a/packages/Amqp/tests/Integration/AmqpChannelAdapterTest.php
+++ b/packages/Amqp/tests/Integration/AmqpChannelAdapterTest.php
@@ -267,7 +267,7 @@ class AmqpChannelAdapterTest extends AmqpMessagingTest
      */
     private function receiveOnce(AmqpInboundChannelAdapterBuilder $inboundAmqpGatewayBuilder, QueueChannel $inboundRequestChannel, ChannelResolver $channelResolver, ReferenceSearchService $referenceSearchService): ?Message
     {
-        return $this->receiveWithPollingMetadata($inboundAmqpGatewayBuilder, $inboundRequestChannel, $channelResolver, $referenceSearchService, PollingMetadata::create('someId')->setExecutionAmountLimit(1));
+        return $this->receiveWithPollingMetadata($inboundAmqpGatewayBuilder, $inboundRequestChannel, $channelResolver, $referenceSearchService, PollingMetadata::create('someId')->setExecutionAmountLimit(1)->setExecutionTimeLimitInMilliseconds(100));
     }
 
     /**
--- a/packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
+++ b/packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
@@ -4,6 +4,7 @@ namespace Ecotone\SymfonyBundle\DepedencyInjection\Compiler;
 
 use Ecotone\Lite\PsrContainerReferenceSearchService;
 use Ecotone\Messaging\Config\Configuration;
+use Ecotone\Messaging\Config\ConfiguredMessagingSystem;
 use Ecotone\Messaging\Config\MessagingSystemConfiguration;
 use Ecotone\Messaging\Config\ProxyGenerator;
 use Ecotone\Messaging\Config\ServiceConfiguration;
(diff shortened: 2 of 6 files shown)


packages/Ecotone/src/Messaging/Config/LazyConfiguredMessagingSystem.php
```diff
@@ -95,6 +95,6 @@ public function replaceWith(ConfiguredMessagingSystem $messagingSystem): void
 
     private function getConfiguredSystem(): ConfiguredMessagingSystem
     {
-        return $this->container->get(LazyConfiguredMessagingSystem::class);
+        return $this->container->get(ConfiguredMessagingSystem::class);
```
jlabedo, line 98: @dgafka , this change makes a lot of sense to me: the lazy configured messaging system calls the real messaging system when needed, but I think it may break the Laravel package, but I never used Laravel, so not easy to add a test. 

In Symfony, with this PR, when the `ConfiguredMessagingSystem` reference is requested from the `SymfonyReferenceSearchService` it returns a `LazyConfiguredMessagingSystem` so there is no circular reference.

In Laravel, I understand that the `LaravelReferenceSearchService` returns directly the real `ConfiguredMessagingSystem` so I think it should break when it builds the container because of the circular reference. Easy to fix, but I would prefer having a failing test first. What do you think ?

dgafka:
@jlabedo can you rebase?

packages/Symfony/composer.json
```diff
@@ -37,6 +37,7 @@
         "phpstan/phpstan": "^1.8",
         "phpunit/phpunit": "^9.5",
         "symfony/doctrine-messenger": "^5.4|^6.0",
+        "symfony/expression-language": "^6.0",
```
dgafka, line 40: I wonder, if we should make it obligated? 
People actually can make use of Ecotone without using expression language ever :) 

If we need to do that, then let's make it `^5.4` compatible too

packages/Amqp/tests/Integration/AmqpChannelAdapterTest.php
```diff
@@ -267,7 +267,7 @@ private function createAmqpInboundAdapter(string $queueName, string $requestChan
      */
     private function receiveOnce(AmqpInboundChannelAdapterBuilder $inboundAmqpGatewayBuilder, QueueChannel $inboundRequestChannel, ChannelResolver $channelResolver, ReferenceSearchService $referenceSearchService): ?Message
     {
-        return $this->receiveWithPollingMetadata($inboundAmqpGatewayBuilder, $inboundRequestChannel, $channelResolver, $referenceSearchService, PollingMetadata::create('someId')->setExecutionAmountLimit(1));
+        return $this->receiveWithPollingMetadata($inboundAmqpGatewayBuilder, $inboundRequestChannel, $channelResolver, $referenceSearchService, PollingMetadata::create('someId')->setExecutionAmountLimit(1)->setExecutionTimeLimitInMilliseconds(100));
```
dgafka, line 270: What does this fixes?

packages/Ecotone/src/Messaging/Config/LazyConfiguredMessagingSystem.php
dgafka: We have tests for Laravel Application now. 
You can find them in `Test\Ecotone\Laravel\Application\Execution\ApplicationTest`

packages/Amqp/tests/Integration/AmqpChannelAdapterTest.php
jlabedo: I think it should, but not sure, hard to reproduce a flaky test.

packages/Symfony/composer.json
jlabedo: It is a dev dependency, it was required to make phpstan pass (if you don't like, only the DepedencyInjection directory is checked by phpstan. So I can just move the code around, but phpstan checks are great :) )

dgafka: I've missed the part that it was dev dependency. Then all good, we can of course have it :)

dgafka:
@jlabedo the tests are failing now due to circular dependency

jlabedo:
Work in progress

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
```diff
@@ -167,5 +169,36 @@ public function process(ContainerBuilder $container)
...
+        $container->setDefinition(ConfiguredMessagingSystem::class, $definition);
+
+        $this->setUpExpressionLanguage($container);
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     * @return void
+     */
+    private function setUpExpressionLanguage(ContainerBuilder $container): void
```
dgafka, line 192: We probably should have class_exists in here, to avoid setting up expression language if it's not installed. 
wdyt?

dgafka (approved):
Amazing :1st_place_medal:

dgafka:
> Sharing benchmark that simulate before and after this PR with the very simple App in Symfony package (I guess a huge impact with a real app):
> ```
> class KernelBootBench
> {
>     /**
>      * @Revs(10)
>      * @Iterations(5)
>      * @Warmup(1)
>      */
>     public function benchKernelBoot()
>     {
>         $kernel = new Kernel('test', false);
>         $kernel->boot();
>     }
> 
>     /**
>      * @Revs(10)
>      * @Iterations(5)
>      * @Warmup(1)
>      */
>     public function benchKernelBootAndLoadMessagingSystem()
>     {
>         $kernel = new Kernel('test', false);
>         $kernel->boot();
>         $kernel->getContainer()->get(ConfiguredMessagingSystem::class);
>     }
> }
> ```

Is this something you could commit to Ecotone, so we can compare the times between PRs?

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
jlabedo: Symfony removes non public dependencies from container if unused, so I think it is not necessary.

PR #122, opened 2023-05-23
EcotoneLite is not able to run console command from configured messaging system

jlabedo:
While working on #121, I had a look at how EcotoneLite was handling circular reference and it seems broken: the `StubConfiguredMessagingSystem` is injected while building the messaging system to protect against circular reference, but it fails (as expected by the stub)

- [X] Add a failing test
- [x] Replace `StubConfiguredMessagingSystem` by `LazyConfiguredMessagingSystem`


packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
```diff
@@ -111,10 +112,13 @@ public function process(ContainerBuilder $container)
...
         $definition = new Definition();
-        $definition->setClass(SymfonyReferenceSearchService::class);
+        $definition->setClass(PsrContainerReferenceSearchService::class);
         $definition->setPublic(true);
         $definition->addArgument(new Reference('service_container'));
-        $container->setDefinition('symfonyReferenceSearchService', $definition);
+        $container->setDefinition(ReferenceSearchService::class, $definition);
+
+        // TODO: Remove this alias used for backward compatibility ?
+        $container->setAlias('symfonyReferenceSearchService', ReferenceSearchService::class)->setPublic(true);
```
dgafka, line 121: Yes, we can remove this, it was only used in [Ecotone Symfony Bundle](https://github.com/ecotoneframework/ecotone-dev/pull/122/files#diff-7c887f66a9090fd68e66da41234a4dc2581c826ebc0bd95e50418678608b40b7L70)

packages/Ecotone/src/Lite/PsrContainerReferenceSearchService.php
```diff
@@ -51,4 +59,14 @@ public function has(string $referenceName): bool
 
         return true;
     }
+
+    public function setConfiguredMessagingSystem(ConfiguredMessagingSystem $configuredMessagingSystem): void
```
dgafka, line 63: I really like this solution. It pushes handling this back to the Ecotone instead of trying to handle it on Symfony/laravel and Lite layers :)

jlabedo: Happy you like it, I can push it forwards by moving the call to this function in the messaging system build phase instead of calling it in Lite/Symfony/Laravel

dgafka: Ye, sure :)

aidan-casey:
@dgafka - It looks like this PR is causing Laravel applications to break in a big way. Laravel's container returns `false` to the `has` method if a concrete implementation is not specifically bound. Laravel will still attempt to auto-wire the class when the `get` method is used from dependencies in the container though.

This means that instances like the following:

```php
final class OrganizationService
{
    public function __construct(private readonly OrganizationRepository $repository)
    {}

    #[CommandHandler]
    public function create(CreateOrganizationCommand $command): void
    {
        // 
    }
}
```

No longer work because the command bus is trying to load the service class using the container, but failing the `has` check and throwing a `ReferenceNotFoundException`.

dgafka:
Hey @aidan-casey,
Can you provide PR with failing test case for that in Laravel module?

PR #123, opened 2023-05-26
perf(Ecotone): avoid unnecessary work on boot when cache is available

jlabedo:
Early return if cached configuration is available.

The impact of this PR can be huge on large codebase, as it avoids loading the attributes from catalog when cache is available.

Example with `package/Ecotone/src` as a catalog I got these results for boostraping Ecotone:
```
    before.........................I4 - Mo163.687ms (±4.31%)
    after..........................I4 - Mo14.700ms (±2.71%)
```

Now booting Ecotone from cache will be stable (14ms in my case) whatever the size of codebase.


tworzenieweb (approved):
I applied exactly the same measures. Looking good

dgafka (approved):

dgafka:
This is amazing finding :)

Thank you @tworzenieweb and @jlabedo :)

dgafka:
@jlabedo I wonder if you could add such time benchmarks as part of Ecotone's CI process, so we can measure optimization of given PR?

PR #127, opened 2023-05-29
tests: add symfony benchmark

jlabedo:


packages/Symfony/tests/Benchmark/KernelBootBench.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * @Revs(10)
+     * @Iterations(5)
+     * @Warmup(1)
+     */
+    public function benchKernelBootAndLoadMessagingSystem()
+    {
+        $kernel = new Kernel('test', false);
```
dgafka, line 28: Can we add for `prod` env too?

dgafka:
Thanks Jean :)

Issue #128, opened 2023-05-30
Symfony cache:clear error with message `You have requested a non-existent service "expressionEvaluationService".`


jlabedo:
**Ecotone version(s) affected**: 1.85.2

**Description**  
1.85.2 is affected, 1.85.1 is not

**How to reproduce**  
bin/console cache:clear when symfony/expression-language is not installed

PR #129, opened 2023-05-30
fix: symfony bundle failing without symfony/expression-language

jlabedo:
Fix for #128 

Not sure it is the way to go, but `ExpressionEvaluationService::REFERENCE` is defined as a required reference so it tries to create a public alias in symfony container to a non existent service if `symfony/expression-language` is not installed.


jlabedo:
Other possible solutions:
- ExpressionEvaluationService::REFERENCE should not be a required reference
- define ExpressionEvaluationService::REFERENCE in container even if it is not used
- drop all aliases in container ?

dgafka:
Closed in favour of https://github.com/ecotoneframework/ecotone-dev/pull/130

PR #131, opened 2023-05-30
add php benchmarks

dgafka:
Add benchmarks for different cases


jlabedo:
@dgafka, xdebug is enabled in CI and it has a big impact on benchmarks. We could use a dedicated job without xdebug loaded for benchmarks

PR #132, opened 2023-05-31
refactor gateway proxy factories

jlabedo:
Simplify gateways building during messaging system build phase.

This PR is linked to #133


packages/Ecotone/src/Messaging/Handler/Gateway/GatewayProxyBuilder.php
```diff
@@ -334,17 +336,15 @@ public function build(ReferenceSearchService $referenceSearchService, ChannelRes
         $proxyFactory = $referenceSearchService->get(ProxyFactory::REFERENCE_NAME);
 
         if ($this->withLazyBuild) {
```
dgafka, line 338: If we are not setting up this, what is this the scenario in which it's used now?
https://github.com/ecotoneframework/ecotone-dev/pull/132/files#r1211723152

packages/Ecotone/src/Messaging/Config/MessagingSystem.php
```diff
@@ -188,38 +190,31 @@ private static function createChannelResolver(array $channelInterceptorBuilders,
...
+                $nonProxyCombinedGatewaysMethods[$proxyBuilder->getRelatedMethodName()] =
+                    $proxyBuilder->buildWithoutProxyObject($referenceSearchService, $channelResolver);
+            }
 
-            if (count($preparedGatewaysForReference) === 1) {
-                $gatewayProxyBuilder = $preparedGatewaysForReference[0]
-                    ->withLazyBuild($isLazyConfiguration);
-                $nonProxyCombinedGateways[$referenceName] = NonProxyCombinedGateway::createWith($referenceName, [$gatewayProxyBuilder->getRelatedMethodName() => $gatewayProxyBuilder->buildWithoutProxyObject($referenceSearchService, $channelResolver)]);
-                $gateways[$referenceName] = GatewayReference::createWith(
+            $nonProxyCombinedGateways[$referenceName] = NonProxyCombinedGateway::createWith($referenceName, $nonProxyCombinedGatewaysMethods);
```
dgafka, line 208: Maybe the implementation of `NonProxyGateway` should be thrown inside `ProxyWithCombinedGateway` to decrease amount of delegations. Wdyt?

jlabedo: Not sure to understand. Do you mean `NonProxyCombinedGateway` code thrown into `GatewayProxyAdapter` (renamed from `ProxyWithCombinedGateways`) ?

dgafka: nvm, got confused :)

packages/Ecotone/tests/Messaging/Fixture/Handler/DumbGatewayBuilder.php
```diff
@@ -40,14 +40,6 @@ public static function create(): self
         return new self();
```
dgafka, line 40: I think we can throw away `DumbGatewawyBuilder` and `GatewayBuilder interface` completely.
`DumbGatewayBuilder` is used in single test, where we can drop it completely or replace with real implementation.
And `GatewayBuilder` is unnecessary abstraction.

docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services:
       - "host.docker.internal:host-gateway"
     # 172.17.0.1
     environment:
+      XDEBUG_CONFIG: >-
+        client_host=host.docker.internal
```
dgafka, line 19: xdebug is configured differently. It should find your ip address based on IP Routing table.
If that's needed for you, you may set up it in `.env` file :)

dgafka:
I think we lost some performance along the way for the scenario where Ecotone is executed.

[Running Ecotone on Master Branch](https://github.com/ecotoneframework/ecotone-dev/actions/runs/5137833987/jobs/9246436137#step:13:183)
[Running Ecotone Application in this PR](https://github.com/ecotoneframework/ecotone-dev/actions/runs/5137833987/jobs/9246436137#step:13:183)

I think what happens right now is `MessagingSystem::configureGateways` is called on each execution. Which means we are creating proxies even if we are not using them at all. Before we had this lazy load, which delayed the creation of Gateway till the moment it's actually executed.

I would love this PR to be merged as it simplifies the code and lower the barrier for understanding what actually happens.
However we need to tackle this lazy creation.
Maybe instead of building it on boot, we can move the creation to `getGatewayByName` and build only the gateway that is requested. Wdyt?

PS. Your benchmarks are doing the job ;)

dgafka:
> Maybe instead of building it on boot, we can move the creation to `getGatewayByName` and build only the gateway that is requested. Wdyt?
> 
> PS. Your benchmarks are doing the job ;)

You may take a look on how `pollingConsumerBuilders` are built. Only the one chosen to be run is built and then built consumer is cached, so in case of reruning it will be reused.

jlabedo:
> Maybe instead of building it on boot, we can move the creation to getGatewayByName and build only the gateway that is requested. Wdyt?

This is actually my final goal, this PR was just for understanding the `MessagingSystem` and clean the room before moving forward. 

From my actual understanding, `MessagingSystem` is mainly a mix between a container builder and a container. I think it may be dropped for a classic container implementation, left to Lite/Symfony/Laravel responsibility, and some of the code thrown to `MessagingSystemConfiguration`

> I think what happens right now is MessagingSystem::configureGateways is called on each execution. Which means we are creating proxies even if we are not using them at all. Before we had this lazy load, which delayed the creation of Gateway till the moment it was actually executed.

I should have looked at it ! I missed something with this lazy boot.

dgafka:
>The problem with benchmarks on CI is that you cannot trust the hardware the benchmark is running on (it may be shared computing power, or even different cpus).

Ah true, really good point :) 

> From my actual understanding, MessagingSystem is mainly a mix between a container builder and a container. I think it may be dropped for a classic container implementation, left to Lite/Symfony/Laravel responsibility, and some of the code thrown to MessagingSystemConfiguration

MessagingSystem is actually connected messaging system. 
As MessagingConfiguration holds the configurations, yet it does not resolve them anyhow. 
This part of docs explains it in [more details](https://docs.ecotone.tech/messaging/contributing-to-ecotone/how-ecotone-works-under-the-hood).
In messaging system everything is decoupled and may be connected together by [Message Channels](https://docs.ecotone.tech/messaging/messaging-concepts/message-channel) and [Messages](https://docs.ecotone.tech/messaging/messaging-concepts/message), that is also why it's so easily to change the flow adding [interceptors](https://docs.ecotone.tech/modelling/interceptors) for example.
In case of Messaging System if you will take a look on [`eventDrivenConsumers`](https://docs.ecotone.tech/messaging/messaging-concepts/consumer#event-driven-consumer) they are booted in order to connect to each other, which allows for the message flow. 
Right now Ecotone have great optimalization, but in ideal implementation we could go even further. 
As right now each Event Driven Handler is booted (and Lazy Config is configured), but we could boot only the Event Driven Consumers that are related to given message flow. You can imagine that this could lead to extremely efficient system.

That's why when you say you would love to see more optimalization in place, I am leaving the space for you to explore. 
Let's see along the way where we can reach with that and I will share the knowledge with you along the way.

PR #134, opened 2023-05-31
remove headers once they are propagated

unixslayer:
in case of asynchronous projection when first event has a header which does not exist in the next one, header will remain


packages/PdoEventSourcing/tests/Fixture/MetadataPropagationWithAsyncProjection/OrderProjectionConfig.php
```diff
@@ -0,0 +1,29 @@
...
+use Ecotone\EventSourcing\ProjectionRunningConfiguration;
+use Ecotone\Messaging\Attribute\ServiceContext;
+use Ecotone\Messaging\Channel\SimpleMessageChannelBuilder;
+use Prooph\EventStore\Pdo\Projection\PdoEventStoreReadModelProjector;
+use Prooph\EventStore\Projection\ReadModelProjector;
+
+class OrderProjectionConfig
+{
+    #[ServiceContext]
+    public function eventDrivenAuditPlanActionLogProjection(): ProjectionRunningConfiguration
```
dgafka, line 16: I guess this configuration has no effect on the test, so we can drop to make the scenario more explicit?

dgafka:
@unixslayer final review, if looks legit? ;)

PR #137, opened 2023-06-02
Running tests without debug. Resolves #135

### LINKED ISSUE #135: Improve benchmarks by running without debug
**Description**  
Currently we are running benchmarks for Ecotone. 
However all the tests are running with xdebug enabled, which affect the shown performance. 
As the xdebug is not really needed for running the tests and it slows the test suite, we can disable it globally.

unixslayer:


dgafka (approved):
Thanks :)

PR #144, opened 2023-06-07
Fullapp benchmark

jlabedo:
Common "real world" app (from RefactorToReactiveSystem Stage3) with Lite, Laravel and Symfony integration, so we can compare performance on dedicated app.


Monorepo/Benchmark/FullAppBenchmark.php
```diff
@@ -0,0 +1,101 @@
...
+use PhpBench\Benchmark\Metadata\Annotations\Warmup;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * @Revs(10)
+ * @Iterations(5)
+ * @Warmup(1)
+ */
+class FullAppBenchmark
```
dgafka, line 21: I think we should leave only the full app benchmark and run it against production. 
Right now there are many variables in the result set, so it's easy to skip the valuable information. 
If we will keep it small, it will be easy to take a quick look to see, if given PR makes the difference or not.

Running the whole app actually covers most of the cases. What we could add it from the other tests is just booting the kernel and doing nothing besides that, so the tests we have in the other benchmarks.
That would ensure that the boot time have not increased due some changes.

Monorepo/Benchmark/FullAppBenchmark.php
```diff
@@ -0,0 +1,101 @@
...
+        $messagingSystem = $container->get(ConfiguredMessagingSystem::class);
+        $orderController = $container->get(OrderController::class);
+        $configuration = $container->get(Configuration::class);
+
+        $this->execute($messagingSystem, $orderController, $configuration->productId());
+
+        $kernel->shutdown();
+    }
+
+    public function bench_symfony_without_ecotone()
```
dgafka, line 38: Interesting idea of testing without Ecotone. 
What would you like to observe thanks to that?

Monorepo/ExampleApp/Common/UI/OrderController.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Monorepo\ExampleApp\Common\UI;
```
dgafka, line 5: Big applause for making it under common. It will be much easier to keep it up to date and change :heart:

dgafka: I need to think about this in case of [tutorials](https://docs.ecotone.tech/tutorial-php-ddd-cqrs-event-sourcing), as they are separate repos having copy paste code.
I wanted to add Saga and Event Sourcing to tutorials too, however as that was too much hassle I dropped the idea for now.

Monorepo/Benchmark/FullAppBenchmark.php
jlabedo: I am actually using those benchmarks for an upcoming PR, the goal is to see the impact of ecotone itself without "client side code". I mean if you have for example a long db call, it will impact performance but not because of ecotone.

dgafka: Do you think we could make a subtraction between the result and present only the time Ecotone took?

jlabedo: Not to my knowledge

.github/workflows/benchmark-pr.yml
```diff
@@ -84,13 +84,20 @@ jobs:
         id: baseline
         continue-on-error: true
         run: |
-          vendor/bin/phpbench --profile=opcache_enabled --retry-threshold=5 --tag=main.opcache_enabled
-          vendor/bin/phpbench --profile=opcache_disabled --retry-threshold=5 --tag=main.opcache_disabled
+          vendor/bin/phpbench run --profile=opcache_enabled --retry-threshold=5 --tag=main.opcache_enabled
+          vendor/bin/phpbench run --profile=opcache_disabled --retry-threshold=5 --tag=main.opcache_disabled
```
dgafka, line 88: Right now the formatting for baseline and PR's benchmarks are a bit different. It would be easier to compare them, if they would be the same.
Can we run baseline with same formatting?

.github/workflows/benchmark-pr.yml
```diff
@@ -84,13 +84,20 @@ jobs:
         id: baseline
         continue-on-error: true
         run: |
-          vendor/bin/phpbench --profile=opcache_enabled --retry-threshold=5 --tag=main.opcache_enabled
-          vendor/bin/phpbench --profile=opcache_disabled --retry-threshold=5 --tag=main.opcache_disabled
+          vendor/bin/phpbench run --profile=opcache_enabled --retry-threshold=5 --tag=main.opcache_enabled
```
dgafka, line 87: I wonder, if there is a value in keeping opcache enabled benchmarks. We will be mostly picking one or another and comparing it each to other. From my understanding as long as the application fits within opcache memory, the results should be the same.

Do you feel it should stay and if so what are the reasons?

Monorepo/ExampleApp/Common/Infrastructure/Messaging/MessageChannelConfiguration.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        /**
+         * This is dbal asynchronous channel (ecotone/dbal), which provides us with Outbox Pattern.
+         * https://docs.ecotone.tech/modelling/asynchronous-handling
+         */
+        return InMemoryRepositoryBuilder::createForSetOfStateStoredAggregates([Order::class]);
+    }
+
+    #[ServiceContext]
+    public function asynchronousErrorHandling()
```
dgafka, line 51: We can remove this. Not part of the benchmark

Monorepo/ExampleApp/Common/Infrastructure/Messaging/MessageChannelConfiguration.php
```diff
@@ -0,0 +1,79 @@
...
+
+    #[ServiceContext]
+    public function configuration()
+    {
+        return ServiceConfiguration::createWithDefaults()
+            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE]));
+    }
+
+    #[ServiceContext]
+    public function asynchronousChannel()
```
dgafka, line 31: As we are not running asynchronous consumers and it would be good to run some `Event Handlers` within benchmark, let's remove this and `Asynchronous` attributes.

Monorepo/ExampleApp/Common/Infrastructure/Messaging/MessageChannelConfiguration.php
```diff
@@ -0,0 +1,79 @@
...
+
+final class MessageChannelConfiguration
+{
+    const ASYNCHRONOUS_CHANNEL = "asynchronous";
+
+    #[ServiceContext]
+    public function configuration()
+    {
+        return ServiceConfiguration::createWithDefaults()
+            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE]));
```
dgafka, line 27: When we will make [Event Handler synchronous](https://github.com/ecotoneframework/ecotone-dev/pull/144/files#r1223903740), we can switch this to: `ModulePackageList::allPackages()`.

Monorepo/ExampleApp/Common/Infrastructure/Messaging/MessageChannelConfiguration.php
dgafka: We need to remember that we also need to call those Event Handler manually from non-ecotone test then

Monorepo/Benchmark/FullAppBenchmark.php
dgafka: What we are trying to achieve here is to see how Ecotone performs against given PR, if the timing was increased or decreased and this can be answered having only Ecotone based tests.
`Non-ecotone` tests could be useful as something optional that we can trigger manually, and even so I think in those cases we will be rather looking for specific optimalization, so some profiler will be needed (which will actually answer this question).

In that sense can we drop the support for `non-ecotone benchmarks`?

Monorepo/ExampleApp/Common/Infrastructure/Messaging/MessageChannelConfiguration.php
dgafka, line 27: This won't work, as it's a global config which is [not merged](https://github.com/ecotoneframework/ecotone-dev/blob/main/packages/Ecotone/src/Messaging/Config/ServiceConfiguration.php#L63).
We need to set it up in used framework 

- [symfony](https://docs.ecotone.tech/modules/symfony-ddd-cqrs-event-sourcing#skippedmodulepackagenames).
- [laravel](https://docs.ecotone.tech/modules/laravel-ddd-cqrs-event-sourcing#skippedmodulepackagenames)
- [Lite](https://docs.ecotone.tech/modules/ecotone-lite#withskippedmodulepackagenames)

And in that sense we can skip [all packages](https://github.com/ecotoneframework/ecotone-dev/blob/35280d2eb0f2628ebba5cd8d90f7794116540dde/packages/Ecotone/src/Messaging/Config/ModulePackageList.php#L38), when we will drop [asynchronocity](https://github.com/ecotoneframework/ecotone-dev/pull/144/files#r1223903740).

Monorepo/ExampleApp/Common/Infrastructure/Messaging/MessageChannelConfiguration.php
```diff
@@ -0,0 +1,79 @@
...
+        return ErrorHandlerConfiguration::createWithDeadLetterChannel(
+            "errorChannel",
+            RetryTemplateBuilder::fixedBackOff(10)
+                ->maxRetryAttempts(2),
+            "dbal_dead_letter"
+        );
+    }
+
+    #[ServiceContext]
+    public function retryStrategy()
```
dgafka, line 66: We can remove this. Not part of the benchmark

Monorepo/ExampleApp/Common/Domain/Order/Order.php
```diff
@@ -0,0 +1,63 @@
...
+final class Order
+{
+    use WithAggregateEvents;
+
+    private function __construct(#[AggregateIdentifier] private UuidInterface $orderId, private UuidInterface $userId, private ShippingAddress $shippingAddress, private ProductDetails $productDetails, private \DateTimeImmutable $orderAt)
+    {
+        $this->recordThat(new OrderWasPlaced($this->orderId));
+    }
+
+    #[Deduplicated('orderId')]
```
dgafka, line 30: We can remove `Deduplication`. Not part of the benchmark

Monorepo/ExampleApp/Common/Domain/Article/Publication.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Ecotone\Modelling\Attribute\Aggregate;
+use Ecotone\Modelling\Attribute\AggregateIdentifier;
+use Ecotone\Modelling\Attribute\CommandHandler;
+use Monorepo\ExampleApp\Common\Domain\Article\Command\ChangeContent;
+use Monorepo\ExampleApp\Common\Domain\Article\Command\ChangeTitle;
+use Monorepo\ExampleApp\Common\Domain\Article\Command\CreatePublication;
+
+#[Aggregate]
+class Publication
```
dgafka, line 13: `Article` namespace seems to not be used within benchmark. We can remove it

Monorepo/ExampleApp/Common/Domain/Notification/NotificationSubscriber.php
```diff
@@ -0,0 +1,31 @@
...
+use Ecotone\Messaging\Attribute\Asynchronous;
+use Ecotone\Modelling\Attribute\EventHandler;
+
+final class NotificationSubscriber
+{
+    public function __construct(private readonly OrderRepository $orderRepository,
+        private readonly UserRepository $userRepository, private readonly NotificationSender $notificationSender)
+    {}
+
+    #[Asynchronous(MessageChannelConfiguration::ASYNCHRONOUS_CHANNEL)]
```
dgafka, line 20: We can test it synchronous. So we can remove `Asynchronous` attribute.

Monorepo/ExampleApp/Common/Domain/Shipping/ShippingSubscriber.php
```diff
@@ -0,0 +1,31 @@
...
+use Ecotone\Messaging\Attribute\Asynchronous;
+use Ecotone\Modelling\Attribute\EventHandler;
+
+final class ShippingSubscriber
+{
+    public function __construct(private readonly OrderRepository $orderRepository, public readonly ShippingService $shippingService)
+    {
+    }
+
+    #[Asynchronous(MessageChannelConfiguration::ASYNCHRONOUS_CHANNEL)]
```
dgafka, line 19: We can test it synchronous. So we can remove Asynchronous attribute.

Monorepo/ExampleApp/Common/Domain/User/User.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace Monorepo\ExampleApp\Common\Domain\User;
+
+use Ecotone\Modelling\Attribute\Aggregate;
+use Ecotone\Modelling\Attribute\AggregateIdentifier;
+use Ramsey\Uuid\UuidInterface;
+
+#[Aggregate]
+final class User
```
dgafka, line 12: Not sure if this is used I guess not. If so we can remove

.github/workflows/benchmark-pr.yml
jlabedo: We can have a different impact with opcache enabled or not, but I think it's not very common. I was trying a different caching method for the messaging system configuration (php serialize vs dumping a php file with symfony VarExporter component) , and it makes a great performance boost with opcache enabled, but decrease a little bit perf with opcache disabled.

dgafka: I see, so there would be a case where having this enabled would give us information if this PR make a difference in case of opcache being enabled. 

That kind of improvements will be rather rate, so let's leave opcache out of the scope. As for those PRs, we may run it locally and verify the results. 
My intention behind this is to keep the result as small and understable for everyone as possible. This way not only the reviewer will be able to verify it, but also the contributor.

Monorepo/ExampleApp/Common/Domain/Article/Publication.php
jlabedo: This has an impact on the benchmark as it is actually loaded by the messaging system. I would like to add more and more aggregates, event handlers, etc. to benchmark on a real world like app.
Still in an upcoming PR :), I lazy load every messaging component so I expect my PR to show a performance boost because it won't load unused components.
Wdyt ?

dgafka: Ah cool, let's keep it then :)

.github/workflows/benchmark-pr.yml
jlabedo: Sure, this is because there is not the same benchmark files in main and PR 😵

dgafka:
This is amazing addition, thank you :)

PR #145, opened 2023-06-09
cache type descriptors and media types

jlabedo:


packages/Ecotone/src/Messaging/Conversion/MediaType.php
```diff
@@ -36,6 +36,8 @@ final class MediaType
 
     private const TYPE_PARAMETER = 'type';
 
+    private static array $parsedMediaTypes = [];
```
dgafka, line 39: Is using enum having the same effect or enum is always new instance?

dgafka: I would not, just curious :P

PR #146, opened 2023-06-13
Message Tracing - Open-Telemetry Integration

dgafka:
# Why

This PR will provide initial implementation for Open-Telemetry integration. The integration is not yet finished and will not be [tagged yet](https://github.com/ecotoneframework/ecotone-dev/pull/146/files#diff-22c8e8b30551117a8c9cf2f77a663e7c6b45c6eb740d0fbcc2b9ffc0639abc53R65). 
If you wish to make use of it, you should use `dev-master` version. However be aware that the implementation is likely to be changed, therefore breaking changes are possible.

## Tracing and Debugging

Messaging architecture decouples component from each other which provides us with isolation, self healing and easy to change and refactor small building blocks. Yet this comes with the cost of visibility, as the more we decouple things, the less process may be actually visible to us, especially when it starts to branch.
To tackle this tracing provides easy to follow way to visualize flows within our system. This provides insights of how the flows are actually looking in reality (instead of outdating manual diagrams). And provides developer with a tool to understand the performance impact each component have on the system.

Screenshot from 2023-06-12 20-40-22

## Open Telemetry abstraction layer

Ecotone could be integrated directly with specific 3rd part provider for tracing like Jaeger, Zipkin, DataDog etc. 
Yet at this moment of time it makes low value, as currently we are close to official `1.0.0` release of Open-Telemetry, abstraction layer that replaces Open-Tracing.  
Open Telemetry provides set of components that abstract away given 3rd providing model for tracing that is compatible will different external provides. 
This means, that if we use Jaeger for tracing and we want to switch to Data Dog, we will be simply switching implementation without any additional work needed.  
Open Telemetry provides abstraction for metrics and logging too, which can be integrated later to provide even more insights.

## New experience for PHP Developers

Together with [Ecotone Pulse](https://docs.ecotone.tech/modules/ecotone-pulse) it will provide new experience in PHP world.
As Ecotone provides full resilient messaging, which means that architecture can be configured in a way that no message will be lost, it means that we can achieve what I call resiliency circle. 
Resiliency circle guarantees that messages are kept within circle (our system) till the moment they are properly consumed. If error happens, message will try to self-heal and if it can't it will land in Ecotone Pulse. 
From there and using OpenTelemetry we will be able to view the flows, see where it stopped, why it stopped and what are the effects of this. After fixing the issue, replay the message from Ecotone Pulse Dashboard to end the circle. 

# Status

This PR is currently in work. There are multiple scenarios to cover including passing the context in asynchronous scenarios on service and between service level. This way we will be able to view not only single service flow, but the flow of our whole system.


packages/Ecotone/src/Messaging/MessageHeaders.php
```diff
@@ -182,6 +182,39 @@ public static function unsetTransportMessageKeys(array $metadata): array
         return $metadata;
     }
 
+    public static function propagateContextHeaders(array $context, array $headers)
```
dgafka, line 185: Important change: 
Correlation and parent id, becomes first class-citizens supported by the framework. 
This way we will be able to look up for messages in tracing systems.

dgafka:
@jlabedo @unixslayer I am merging this, yet this is not production ready. Still have to cover different flows before it will go live.
However current implementation is enough to keep it in main branch and run tests against it, to secure that incoming changes will not break it.

Issue #133, opened 2023-05-31
Rely more heavily on a PSR ContainerInterface to build Ecotone Messaging System


jlabedo:
**Description**  
Rely more heavily on a PSR ContainerInterface to build Ecotone Messaging System
This will make possible to use existing containers implementations that have a lot of optimisation to build services required by the messaging system (Symfony or Laravel)
The building of messaging components (gateways, channels, etc.) could be lazy built, performance improved, and maybe reduce ecotone codebase size

PR #148, opened 2023-06-15
PoC: Refactor messaging system with PSR container interface

jlabedo:
Many different things to discuss in this PR.
The first goal was to lazy load every messaging components (gateway, channels, etc.) thanks to already existing container implementation, and this PR uses exclusively Symfony for the proof of concept. It is related to issue #133 .

From my understanding, Ecotone is built around:

- `MessagingSystemConfiguration`: where all messaging components are configured, but I think not as much as it should (some "Builders" leak elsewhere in the code). This configuration is dumped for performance. I think this component looks a lot like a `ContainerBuilder` in PHP-DI or Symfony world. 
- `MessagingSystem`: where all built components are registered (and build in the static function `buildFromMessagingConfiguration` but that is just moving code around). This component looks a lot like a `Container`
- `Modules`: where you can hook into the `MessagingSystemConfiguration` (the `ContainerBuilder`), this looks a lot like a `Bundle` in Symfony world.

I think it could be possible for Ecotone to rely on external `ContainerBuilder` and `Container` implementations and propose a default one in `EcotoneLite`. This could lead to a lot of simplification and optimization of Ecotone.
A common interface could be that all "Builders" have a `build` method that accept a single parameter `EcotoneContainerBuilder`. Instead of returning an actual object, "Builders" build method could configure the `EcotoneContainerBuilder` on *how* to build the object using dependency injection, like a ContainerBuilder in Symfony of PHP-DI.

The downside is that there is a necessity to maintain multiple container implementation (Symfony/Laravel/Lite), and as we rely more heavily on them, we may have to run all test suites on each container implementation.

Another downside is that it is not compatible with EcotoneLite testing framework which rely on replacing the `ConfiguredMessagingSystem` inside the provided container (the replaceWith method). I think this is not a problem, as the `replaceWith` method is bit of a hack to me. Moreover, I think it is a strange to propose to build a new container for each test case as it is with EcotoneLite (the `$containerOrAvailableServices` parameter). It makes sense for testing Ecotone, but for a real world app, you have a single container in production, and you want to make all your tests against this one or something very near from it. And not configure a specific container for each test case. If it makes sense for some unit tests , we could use EcotoneLite without any configured container, so the problem is solved.

While building a container with Ecotone, it also required pushing a maximum of code that could be done only at compilation time vs code that should be dynamic (that depends on the `Message`), improving performance.
There is quite a big impact on performance with this PR when opcache is enabled, mainly due to how the configuration is dumped as a php file instead of using `unserialize`. But there are other perf improvements.
Here are my figures (you should look only at Symfony which implements the new system) which shows a x3 performance boost:

This PR (opcache enabled):

| benchmark           | subject            | revs | its | mem_peak | mode      | rstdev |
| ------------------- | -------------------| ---- | --- | -------- | --------- | ------ |
| PlaceOrderBenchmark | bench_symfony_prod | 10   | 5   | 5.069mb  | 763.504μs | ±2.08% |
| PlaceOrderBenchmark | bench_symfony_dev  | 10   | 5   | 5.166mb  | 798.032μs | ±0.71% |
| PlaceOrderBenchmark | bench_laravel_prod | 10   | 5   | 8.869mb  | 68.192ms  | ±1.07% |
| PlaceOrderBenchmark | bench_laravel_dev  | 10   | 5   | 8.869mb  | 69.248ms  | ±1.61% |
| PlaceOrderBenchmark | bench_lite_prod    | 10   | 5   | 12.879mb | 3.503ms   | ±0.42% |
| PlaceOrderBenchmark | bench_lite_dev     | 10   | 5   | 7.717mb  | 69.470ms  | ±1.75% |

Main (opcache enabled):
| benchmark           | subject            | revs | its | mem_peak | mode     | rstdev |
| --------------------| ------------------ | ---- | --- | -------- | -------- | ------ |
| PlaceOrderBenchmark | bench_symfony_prod | 10   | 5   | 13.148mb | 2.310ms  | ±0.45% |
| PlaceOrderBenchmark | bench_symfony_dev  | 10   | 5   | 13.261mb | 2.284ms  | ±1.18% |
| PlaceOrderBenchmark | bench_laravel_prod | 10   | 5   | 8.936mb  | 69.296ms | ±1.64% |
| PlaceOrderBenchmark | bench_laravel_dev  | 10   | 5   | 8.936mb  | 69.075ms | ±0.94% |
| PlaceOrderBenchmark | bench_lite_prod    | 10   | 5   | 12.822mb | 3.289ms  | ±2.71% |
| PlaceOrderBenchmark | bench_lite_dev     | 10   | 5   | 7.779mb  | 68.737ms | ±0.72% |

I think this PR is should be regarded as a proof of concept, but we can peak some features not directly related to the Container refactoring thing. Here are the one i am thinking of:
- Dumping a `PreparedConfiguration` object as a php file for opcache optimization (using VarExporter component that is already included with ecotone symfony bundle)
- Using a simpler code generation for proxies without proxymanager package but using directly laminas-code, because we are just proxyfying interfaces only which is simple (see packages/Symfony/SymfonyBundle/Proxy/ProxyFactory.php and an example snapshot of a proxy: packages/Symfony/tests/phpunit/ProxyFactoryTest.snapshot)
- Symfony bundle is now unregistering the autoloader on kernel shutdown, which removes a memory leak. We may consider having a shutdown function in EcotoneLite to unregister the autoloader.
- Transform the `MessagingSystem` to a simple container. The only thing that is done in the `MessagingSystem` is to `run` event driven consumers. This PR makes event driven consumer subscribe when its input channel is loaded.


dgafka:
I will find time to review this on saturday/sunday, as this is interesting idea and I want to spend some time on rethinking this.

dgafka:
btw the times with opcache are impressive, it's like we would simply execute clear PHP code.

dgafka:
I would love to see separate PRs or discuss those in separate topics, as there is a lot of going on here :smile: 
This makes is a bit hard to follow, as I am not sure what change references to what topic. 
I've started [Github Discussions in ecotone-dev](https://github.com/ecotoneframework/ecotone-dev/discussions), so if you want to discuss given topic we can also do it there from now.
But let's try to keep one topic at time, as the changes you proposed can be separately discussed and merge regardless of each other :)

Giving feedback accordingly to what I learned from the PR:

- Using var exporter
I would love to see this as separate PR. I also wonder, what are the results without the opcache for this one. 
If there is general improvement in booting Ecotone with and without opcache I think we can make it part of core Ecotone package and replace native php serialization.
- Using laminas-code to generate proxies
What is the reason for maintaining this, does it the speed run time somehow? If so I am happy to see that as separate PR
- Autoloader registered and deregistered only when needed
Yep would love to see a separate PR working for Lite and Laravel too :)
- Subscribe event driven consumer subscribe when its input channel is loaded.
I would also love to see that as separate PR. I agree with the approach of executing only the components that are necessary in given flow

---

> It makes sense for testing Ecotone, but for a real world app, you have a single container in production, and you want to make all your tests against this one or something very near from it. And not configure a specific container for each test case. If it makes sense for some unit tests , we could use EcotoneLite without any configured container, so the problem is solved.

Well this is one is actually challenging. If you consider that you want to reuse the configuration from your Symfony App, then you will bring configuration of all components with you.
The reason to boot new Ecotone Lite instance is to bring context specific classes to the scenario without any additional stuff.
This allows for testing components in full isolation, like only aggregate, which does not trigger Event Handlers (because they are not even registered), switch from synchronous to asynchronous code, providing stubs on the places where it does real integration etc.
You can still use Ecotone directly from Symfony DI Container by fetching and executing Command Bus. Personally I actually found it low value because the context is just too big and tests on Ecotone Lite are actually near to the production usage.

If you feel like you there is value in this, let's discuss separately on github discussions.

> While building a container with Ecotone, it also required pushing a maximum of code that could be done only at compilation time vs code that should be dynamic (that depends on the Message), improving performance.

I fully agree. Ideal is that the execution code, executes and builds only Message Handlers/Channels that are being triggered.
There is a lot of potential to improve thanks to decoupled nature of the components.

> I think it could be possible for Ecotone to rely on external ContainerBuilder and Container implementations and propose a default one in EcotoneLite

If you take a look on how Symfony or Laravel are used, there is minimalistic amount of the code, even Console Commands are done on Ecotone side and just re-registered in Symfony/Laravel.
The reason for this is maintainece and higher assurence that testing on lower level components will also work in higher level environments. 
There is a lot of cool projects that failed on this, including Prooph which had really hard time maintaining the changes that were done in Symfony and keeping it up to date.
With current approach, new versions of Symfony or Laravel are not really a problem, as the space of the integration is really small. Thanks to that there is high assurance that new versions of those framework will work without any or small amount of changes.

That's why there would need to be really strong argument for pushing this outside of Ecotone as current implementation is for a reason :) 
If you want we can discuss this one separately too.

---

I am really glad that you want to tackle so many improvements and I see a lot of potential in those. This may actually push Ecotone with performance to new level, thank you :)

jlabedo:
Thank you for taking time reviewing this. I was really throwing some ideas while understanding how ecotone is built and I guess it was not an easy one to review! I think we can close this PR and discuss the main subject of, let's say, "Messaging System compilation" in a github discussion as you suggest.

Regarding proxy generation, I dropped proxymanager and used laminas-code directly (which is also used by proxymanager) because I had difficulties understanding some parts of proxying code. I can remember of:
- the warmup phase that was doing strange things to me (like warming an empty proxy or something like that ?)
- the proxy Configuration object that was rebuilt many times and that have side effects by looking at proxymanager code
- if you dump the messaging configuration and warmup cache on your local computer, then move the cache to another folder, it will try to regenerate proxies in the the proxy path of your local computer (eg. /Users/jean/ecotone-dev/var/cache) instead of the folder where code is executed (eg. /app/var/cache). This is especially problematic in a lambda where you can write files only in the /tmp folder
With a bunch of lines, I think it is a lot easier to understand what is being done in a Gateway Proxy. I did not do this with performance in mind. I don't think this will have a noticeable impact on performance.
 
I will propose a PR with this change only but if we choose not to go this way, I will be happy to propose another PR the  autoloader registration/deregistration issue globally for Ecotone.

dgafka:
@jlabedo ok closing and waiting for the PRs :)

PR #150, opened 2023-06-21
Reusing events between aggregates

unixslayer:


dgafka:
Long time no see work on this and you were other changes in the meantime. So I guess this one is not really needed or waiting for better time. So closing as far now, to clean things up.

PR #151, opened 2023-06-22
when EventSourcedAggregate factory method does not return events, SaveAggregateService should not try to save it

unixslayer:
in addition
- removing AggregateFactory attribute due its deprecation


packages/Ecotone/src/Modelling/SaveAggregateService.php
```diff
@@ -210,4 +195,31 @@ private function getAggregateIds(array $aggregateIds, object $aggregate, bool $t
 
         return AggregateIdResolver::resolveArrayOfIdentifiers(get_class($aggregate), $aggregateIds);
     }
+
+    private function resolveEvents(Message $message, array $metadata): array
```
dgafka, line 199: docblock for typehint please :)

packages/Ecotone/src/Modelling/SaveAggregateService.php
```diff
@@ -77,32 +77,17 @@ public function save(Message $message, array $metadata): Message
...
-        $events = [];
+        $events = $this->resolveEvents($message, $metadata);
 
-        if ($this->isEventSourced) {
-            $events = $message->getPayload();
-            Assert::isIterable($events, "Return value Event Sourced Aggregate {$this->aggregateInterface} must return array of events");
-        } elseif ($this->aggregateMethodWithEvents) {
-            $events = call_user_func([$aggregate, $this->aggregateMethodWithEvents]);
+        if ($this->isEventSourced && $events === []) {
+            return MessageBuilder::fromMessage($message)->build();
```
dgafka, line 83: As far as I remember we return events from calling Event Sourcing Aggregate. 
If that's the case, we should return change the payload to empty array.

Can you verify?

unixslayer: @dgafka this case won't appear. Event Sourcing Aggregates methods are forced to return array.

unixslayer: this way Message already contains empty array

dgafka (approved):

dgafka:
Ok, let's go :)

PR #163, opened 2023-07-16
Poll consumer while there are messages to consume

jlabedo:
Add `--finishWhenNoMessages` option when running asynchronous endpoints. The execution will be stopped when there are no messages to handle

### DIFF
--- a/packages/Ecotone/tests/Messaging/Unit/Endpoint/Poller/PollingConsumerBuilderTest.php
+++ b/packages/Ecotone/tests/Messaging/Unit/Endpoint/Poller/PollingConsumerBuilderTest.php
@@ -17,8 +17,10 @@ use Ecotone\Messaging\Handler\InMemoryReferenceSearchService;
 use Ecotone\Messaging\Handler\InterfaceToCallRegistry;
 use Ecotone\Messaging\Handler\Recoverability\RetryTemplateBuilder;
 use Ecotone\Messaging\Handler\ServiceActivator\ServiceActivatorBuilder;
+use Ecotone\Messaging\Message;
 use Ecotone\Messaging\MessageHeaders;
 use Ecotone\Messaging\MessagingException;
+use Ecotone\Messaging\PollableChannel;
 use Ecotone\Messaging\Support\MessageBuilder;
 use InvalidArgumentException;
 use RuntimeException;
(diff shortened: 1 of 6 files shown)


packages/Ecotone/tests/Messaging/Unit/Endpoint/Poller/PollingConsumerBuilderTest.php
```diff
@@ -393,6 +395,60 @@ public function test_throwing_exception_and_requeing_when_stop_on_failure_define
         $this->assertNull($errorChannel->receive());
     }
 
+    public function test_finish_when_no_messages(): void
+    {
+        $inputChannelName = 'inputChannel';
+        $inputChannel = new class implements PollableChannel {
```
dgafka, line 401: You can switch this to `QueueChannel::create()`, as this is default In Memory Pollable implementation :)

packages/Ecotone/tests/Messaging/Unit/Endpoint/Poller/PollingConsumerBuilderTest.php
```diff
@@ -393,6 +395,60 @@ public function test_throwing_exception_and_requeing_when_stop_on_failure_define
...
+            $messageHandlerBuilder,
+            PollingMetadata::create('some')
+                ->setFinishWhenNoMessages(true)
+        );
+
+        $inputChannel->send(MessageBuilder::withPayload('some')->build());
+        $inputChannel->send(MessageBuilder::withPayload('some')->build());
+        $pollingConsumer->run();
+
+        $this->assertSame(0, $inputChannel->queueSize());
```
dgafka, line 448: When you switch to In Memory channel, then in here you can run `->receive()` and assert for `null`, as everything was already consumed.

packages/Ecotone/tests/Messaging/Unit/Endpoint/Poller/PollingConsumerBuilderTest.php
jlabedo: I used this anonymous class to add the `queueSize` method, to check the number of messages left in the queue

packages/Ecotone/tests/Messaging/Unit/Endpoint/Poller/PollingConsumerBuilderTest.php
jlabedo: ok thanks

PR #167, opened 2023-07-25
MessageGateway accepts only simple types in payload

unixslayer:


dgafka:
Will be handled with https://github.com/ecotoneframework/ecotone-dev/pull/174

PR #168, opened 2023-07-25
fix error when using --finishWhenNoMessages and when failFast is disabled

jlabedo:
The error was:
`There is problem with configuration. Interface to call Ecotone\\Messaging\\Endpoint\\Interceptor\\FinishWhenNoMessagesInterceptor:postSend was never registered via related interfaces.`

There was no error when failFast was enabled


dgafka:
Ok, thanks :)

Issue #175, opened 2023-08-10
Add support of older versions of Doctrine/Dbal library


tworzenieweb:
**Description**  
In order to provide wider support of this library for legacy projects, we should add support for legacy versions of `doctrine/dbal` library. In particular versions `^2.12.0|^3.0` have a bit different `QueryBuilder` methods available.

The other small changes are related to accessing the schema manager.

**Example**  
The implementation requires a proxy that will detect which version of QueryBuilder was provided. For the old one, it needs to add the missing shortcut fetch methods.

In order to not patch the doctrine code we need to provide a wrapper for query builder chain methods, providing the compatible base type but calling wrapped instance and extending with missing features.

dgafka:
Solved with https://github.com/ecotoneframework/ecotone-dev/pull/176 https://github.com/ecotoneframework/ecotone-dev/pull/179

PR #176, opened 2023-08-10
Add support of older versions of Doctrine/Dbal library #175

tworzenieweb:
In order to provide wider support of this library for legacy projects, we should add support for legacy versions of doctrine/dbal library. In particular versions ^2.12.0|^3.0 have a bit different QueryBuilder methods available.

The other small changes are related to accessing the schema manager.

### DIFF
--- a/packages/Dbal/src/Compatibility/QueryBuilderProxy.php
+++ b/packages/Dbal/src/Compatibility/QueryBuilderProxy.php
@@ -1,0 +1,224 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ecotone\Dbal\Compatibility;
+
+use Doctrine\DBAL\Query\QueryBuilder;
+use Doctrine\DBAL\Result;
+
+/**
+ * @package Ecotone\Dbal\Compatibility
+ * @author Łukasz Adamczewski <tworzenieweb@gmail.com>
+ *
+ * Simple proxy class to keep the QueryBuilder API compatible with Doctrine DBAL 2.10 and 3.0
+ * All the parent methods need to be implemented in order to intercept and pass to the wrapped instance.
+ * Class supports execution of various fetch methods directly from query object that was added in version 3.1
+ *
+ * @see https://github.com/doctrine/dbal/blob/3.6.x/UPGRADE.md#upgrade-to-31
+ */
+final class QueryBuilderProxy extends QueryBuilder
+{
+    public function __construct(private QueryBuilder $queryBuilder)
+    {
+    }
+
+    // override all public methods from parent class with empty body
+    public function select($select = null)
+    {
+        $this->queryBuilder->{__FUNCTION__}(...func_get_args());
+
+        return $this;
+    }
+
+    public function from($from, $alias = null)
+    {
+        $this->queryBuilder->{__FUNCTION__}($from, $alias);
+
+        return $this;
+    }
+
+    public function addSelect($select = null)
+    {
+        $this->queryBuilder->{__FUNCTION__}(...func_get_args());
+
+        return $this;
+    }
+
+    public function delete($delete = null, $alias = null)
+    {
+        $this->queryBuilder->{__FUNCTION__}($delete, $alias);
+
+        return $this;
+    }
+
+    public function update($update = null, $alias = null)
+    {
+        $this->queryBuilder->{__FUNCTION__}($update, $alias);
+
+        return $this;
+    }
+
+    public function set($key, $value)
+    {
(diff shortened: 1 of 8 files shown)


tworzenieweb:
I got one concern for testing, locally I am able to shuffle the doctrine/dbal version between 2 and 3 by modifying the composer.json, but in CI/CD not sure how we could achieve that

dgafka:
> I got one concern for testing, locally I am able to shuffle the doctrine/dbal version between 2 and 3 by modifying the composer.json, but in CI/CD not sure how we could achieve that.

@tworzenieweb we are covered with this.
 CI runs with prefer-lowest and prefer-stable, to ensure lowest and highest possible stable versions :) 
Screenshot from 2023-08-11 08-30-25

dgafka:
Can you [fix the compatibility in main `composer.json`](https://github.com/ecotoneframework/ecotone-dev/actions/runs/5825999144/job/15809483211?pr=176#step:7:20)? This allows to run Monorepo tests with same range of versions as each package has :)

packages/Dbal/src/Compatibility/QueryBuilderProxy.php
```diff
@@ -0,0 +1,224 @@
...
+ * @package Ecotone\Dbal\Compatibility
+ * @author Łukasz Adamczewski <tworzenieweb@gmail.com>
+ *
+ * Simple proxy class to keep the QueryBuilder API compatible with Doctrine DBAL 2.10 and 3.0
+ * All the parent methods need to be implemented in order to intercept and pass to the wrapped instance.
+ * Class supports execution of various fetch methods directly from query object that was added in version 3.1
+ *
+ * @see https://github.com/doctrine/dbal/blob/3.6.x/UPGRADE.md#upgrade-to-31
+ */
+final class QueryBuilderProxy extends QueryBuilder
```
dgafka, line 20: Do we need to extend `QueryBuilder`? If they will put `final` on the class, this will break. 

I am fine with Ecotone's abstraction which will hide the QueryBuilder inside :)

tworzenieweb: sadly there is no interface to implement to duck type, because some tests are creating QB instance and got a return type defined to be QueryBuilder from Doctrine/Dbal. We could of course change that if you like.

```
    private function getDocumentsFor(string $collectionName): \Doctrine\DBAL\Query\QueryBuilder
    {
        return (new QueryBuilderProxy($this->getConnection()->createQueryBuilder()))
            ->select('document', 'document_type')
            ->from($this->getTableName())
            ->andWhere('collection = :collection')
            ->setParameter('collection', $collectionName, Types::TEXT);
    }
```

tworzenieweb:
> > I got one concern for testing, locally I am able to shuffle the doctrine/dbal version between 2 and 3 by modifying the composer.json, but in CI/CD not sure how we could achieve that.
> 
> @tworzenieweb we are covered with this. CI runs with prefer-lowest and prefer-stable, to ensure lowest and highest possible stable versions :) Screenshot from 2023-08-11 08-30-25

nice, wasn't aware of that :+1:

packages/Dbal/src/Compatibility/QueryBuilderProxy.php
dgafka: Ok, I see, let's keep it this way then. 
I count on Dbal, that they will provide interface in case of making it final :)

tworzenieweb: When I think about it now, it might be also a nice starting point if doctrine/dbal 4.0 will ever be released.

dgafka:
Thanks for contribution :)

marmichalski:
> > I got one concern for testing, locally I am able to shuffle the doctrine/dbal version between 2 and 3 by modifying the composer.json, but in CI/CD not sure how we could achieve that.
> 
> @tworzenieweb we are covered with this. CI runs with prefer-lowest and prefer-stable, to ensure lowest and highest possible stable versions :) Screenshot from 2023-08-11 08-30-25

CI runs against 3.x only (lowest as 3.6.0 to be precise), becauseof dependency on doctrine-bundle in version 2.9. DoctrineBundle 2.9.0 requires [dbal 3.6.0](https://github.com/doctrine/DoctrineBundle/blob/c09bfc30e18ad9ad3f5e1d967888437b916f0254/composer.json#L34). 😎

dgafka:
Thanks @marmichalski for taking a look on that :)

I've fixed that with https://github.com/ecotoneframework/ecotone-dev/pull/179

PR #185, opened 2023-08-17
allow custom timeout for deduplication

unixslayer:


packages/Dbal/src/Configuration/DbalConfiguration.php
```diff
@@ -218,6 +221,11 @@ public function isDeduplicatedEnabled(): bool
         return $this->deduplicatedEnabled;
     }
 
+    public function minimumTimeToRemoveMessageFromDeduplication(): bool
```
dgafka, line 224: Return type is bool, which cast to 1 ms and removes messages immediately. Changing it to `int` resolves broken tests :)

unixslayer: a shit. damn copy-paste

dgafka (approved):

PR #190, opened 2023-08-30
Clearing object manager on command bus in a synchronous flow

jlabedo:
**Ecotone version(s) affected**: 1.99.0

**Description**  
My tests with synchronous event handlers are failing because of duplicate aggregates primary key when upgrading to 1.99.0 because of the new default configuration clearing the object manager on command bus.
The new default configuration is flushing and clearing the object manager after handling commands. In the case of a synchronous event handler dispatching another command (Command A => Event => Command B), command A will fetch Aggregate A, the event handler will be executed, then Aggregate B will be fetched and saved by Command B, the object manager is cleared, making Aggregate A untracked. When Aggregate A is saved, the object manager think it is a new object and will insert it instead of updating it.

**Possible Solution**  
Flush on every command but clear only after the topmost command.

### DIFF
--- a/packages/Dbal/src/ObjectManager/ObjectManagerInterceptor.php
+++ b/packages/Dbal/src/ObjectManager/ObjectManagerInterceptor.php
@@ -18,6 +18,8 @@ class ObjectManagerInterceptor
      */
     private $connectionReferenceNames;
 
+    private int $depthCount = 0;
+
     public function __construct(array $connectionReferenceNames)
     {
         $this->connectionReferenceNames = $connectionReferenceNames;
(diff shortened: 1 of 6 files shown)


dgafka:
> The new default configuration is flushing and clearing the object manager after handling commands. In the case of a synchronous event handler dispatching another command (Command A => Event => Command B), command A will fetch Aggregate A, the event handler will be executed, then Aggregate B will be fetched and saved by Command B, the object manager is cleared, making Aggregate A untracked. When Aggregate A is saved, the object manager think it is a new object and will insert it instead of updating it.

You really have crazy flows there :D 
Will release that with `1.102.1`

jlabedo:
> You really have crazy flows there :D

Are you talking about the particular test case that is crazy ? Because yes, it was the quickest way to reproduce the bug, and it is crazy :)
Or are you talking about the general synchronous Command => Event => Command flow that is crazy ?

dgafka:
> Because yes, it was the quickest way to reproduce the bug, and it is crazy :)

Ye, the flow under test :)

PR #191, opened 2023-08-30
Skip test package in laravel and symfony by default

dgafka:
Currently symfony and laravel loads test package, which is not meant to run on production.


dgafka:
There should be a separate option in the config that enables that module. But yea with current solution, not possible.

dgafka:
@jlabedo Are you up to add supporting PR for that?

PR #196, opened 2023-09-08
Fix dev environment for mac arm64

jlabedo:
On mac arm64, I was getting this error while running sqs test suite :
`Aws\Sqs\Exception\SqsException : Error executing "CreateQueue" on "http://localstack:4576"; AWS HTTP error: cURL error 7: Failed to connect to localstack port 4576: Connection refused`

Without being sure, I think it is related to mac arm64 architecture, newer version of localstack providing an arm64 docker image.

It also fixes `WARN[0000] network default: network.external.name is deprecated. Please set network.name with external: true`

### Note
I had to `docker-compose down` before `docker-compose up -d` for the network change to take effect.


dgafka:
Nice. I tried version 1.0 some time ago and did not work, cool that 2.0 looks fine :)

PR #197, opened 2023-09-08
Rewrite around method parameter resolution

jlabedo:
The behavior is not exactly the same: before, the argument resolution can be overridden at a later stage, now it returns immediately after resolution. Am I breaking something ?


dgafka:
I guess you provided this change, because it broke something on your side.
Can you provide an test case for that?

jlabedo:
Nope, just walking through the code and trying to understand what is going on :) 
My end goal is to optimize some parts of the method invocation, which has some perf penalty, and maybe reduce the stack size which is becoming huge in our app

dgafka:
> Nope, just walking through the code and trying to understand what is going on :)

Ah nice, in that case I am merging the PR

> My end goal is to optimize some parts of the method invocation

Consider refactor safe as long as tests pass. That's why for bug fixes I ask for test cases, so on later stage, we can do refactor and be confident that everything still works :)

> reduce the stack size which is becoming huge in our app

Well then I need to tell you a secret, I recently reduced the stacktrace because it was even bigger :D 
But yea, I am sure we still have places that we can improve here ;)

PR #201, opened 2023-09-20
Refactor method invocation 2

jlabedo:
Cleaning responsibilities of method invocation :
- around interceptors moved to the responsibility of RequestReplyProduces (because of #102)
- conversion service is moved from MethodInvoker to PayloadConverter (makes sense that only the payload can be converted)
- Lower stack trace by removing ServiceActivatorHandler, Splitter, Enricher handlers (wich are RequestReplyProducers)


unixslayer:
@jlabedo please rebase this locally

packages/Ecotone/src/Messaging/Handler/ServiceActivator/ServiceActivatorBuilder.php
```diff
@@ -38,6 +39,8 @@ final class ServiceActivatorBuilder extends InputOutputMessageHandlerBuilder imp
     private bool $shouldPassThroughMessage = false;
     private bool $shouldWrapResultInMessage = true;
 
+    private bool $isCompiled = false;
```
dgafka, line 42: Hmm, how is this used?

dgafka:
nice refactor @jlabedo :)

packages/Ecotone/src/Messaging/Handler/ServiceActivator/ServiceActivatorBuilder.php
jlabedo: > Hmm, how is this used?

Oups wrong rebase

PR #203, opened 2023-09-22
Allow to define custom persistence strategy per stream.

### LINKED ISSUE #202: Allow to have multiple different Persistence Strategies in one application
Current Event Store implementation allows you to have only one global Persistence Strategy in the whole application. However, there may be a reason to have different strategies depending on Aggregate.

unixslayer:
If custom strategy is not defined, the one default will be used.

resolves #202

### DIFF
--- a/packages/PdoEventSourcing/src/EventSourcingRepository.php
+++ b/packages/PdoEventSourcing/src/EventSourcingRepository.php
@@ -132,7 +132,10 @@ class EventSourcingRepository implements EventSourcedRepository
             $streamName =  $this->aggregateStreamMapping->getAggregateToStreamMapping()[$aggregateClassName];
         }
 
-        if ($this->eventSourcingConfiguration->isUsingAggregateStreamStrategy()) {
+        if (
+            $this->eventSourcingConfiguration->isUsingAggregateStreamStrategy()
+            || $this->eventSourcingConfiguration->isUsingAggregateStreamStrategyFor($streamName)
+        ) {
             $streamName = $streamName . '-' . $aggregateId;
         }
 
--- a/packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
+++ b/packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
@@ -1,0 +1,97 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Test\Ecotone\EventSourcing\Integration;
+
+use Ecotone\EventSourcing\EventSourcingConfiguration;
+use Ecotone\EventSourcing\EventStore;
+use Ecotone\EventSourcing\Prooph\LazyProophEventStore;
+use Ecotone\Lite\EcotoneLite;
+use Ecotone\Messaging\Config\ModulePackageList;
+use Ecotone\Messaging\Config\ServiceConfiguration;
+use Test\Ecotone\EventSourcing\EventSourcingMessagingTestCase;
+use Test\Ecotone\EventSourcing\Fixture\MultiplePersistenceStrategies\Basket;
+use Test\Ecotone\EventSourcing\Fixture\MultiplePersistenceStrategies\BasketCreated;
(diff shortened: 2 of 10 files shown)


packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
```diff
@@ -0,0 +1,97 @@
...
+            classesToResolve: [Order::class, Basket::class, BasketProjection::class, OrderProjection::class],
+            containerOrAvailableServices: [
+                new EventsConverter(),
+                new BasketProjection($this->getConnection()),
+                new OrderProjection($this->getConnection()),
+                $this->getConnectionFactory(),
+            ],
+            configuration: ServiceConfiguration::createWithDefaults()
+                ->withEnvironment('prod')
+                ->withSkippedModulePackageNames([ModulePackageList::AMQP_PACKAGE, ModulePackageList::JMS_CONVERTER_PACKAGE])
```
dgafka, line 36: This test may start to fail, when we will add more Modules, as they may require configuration. 
Can you use white list strategy instead?

```php 
withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::EVENT_SOURCING_PACKAGE]))
```

packages/PdoEventSourcing/src/EventSourcingRepository.php
```diff
@@ -132,7 +132,10 @@ private function getStreamName(string $aggregateClassName, mixed $aggregateId):
             $streamName =  $this->aggregateStreamMapping->getAggregateToStreamMapping()[$aggregateClassName];
         }
 
-        if ($this->eventSourcingConfiguration->isUsingAggregateStreamStrategy()) {
+        if (
+            $this->eventSourcingConfiguration->isUsingAggregateStreamStrategy()
+            || $this->eventSourcingConfiguration->isUsingAggregateStreamStrategyFor($streamName)
```
dgafka, line 137: We can hide the global check `$this->eventSourcingConfiguration->isUsingAggregateStreamStrategy()` inside isUsingAggregateStreamStategy. In a way, if there is no strategy for given stream, use global.

This way we would only use `$this->eventSourcingConfiguration->isUsingAggregateStreamStrategyFor($streamName)`

packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
```diff
@@ -0,0 +1,97 @@
...
+                ->withExtensionObjects([
+                    EventSourcingConfiguration::createWithDefaults()
+                        ->withSimpleStreamPersistenceStrategy()
+                        ->withPersistenceStrategyFor(Order::STREAM, LazyProophEventStore::AGGREGATE_STREAM_PERSISTENCE)
```
dgafka, line 40: Can you add docs for this? :)

packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
```diff
@@ -0,0 +1,97 @@
...
+                ])
+                ->withNamespaces([
+                    'Test\Ecotone\EventSourcing\Fixture\MultiplePersistenceStrategies',
```
dgafka, line 43: Can you also add an Aggregate with custom persistence strategy to the test?

dgafka (approved):
This is cool :)

packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
dgafka: Ok, I see, you already did that

packages/PdoEventSourcing/tests/Integration/MultiplePersistenceStrategiesTest.php
dgafka: Ok, it won't work with custom. Can we therefore assert for possible options, to disallow this?

PR #205, opened 2023-09-22
create aggregate from another aggregate

### LINKED ISSUE #204: Allow to create an aggregate from another aggregate
We may have a business case that requires to have two aggregates where one is in charge of creating a second one. 

Currently, Ecotone does not provide that possibility.

- calling State-based Aggregate can only return an instance of aggregate being called
- calling Event Sourcing Aggregate can only return an array that contains events related to aggregate being called

Following behavior is required:

- State-based Aggregate method can return new instance of another State-based Aggregate
- State-based Aggregate which was called should still be able to record events using `#[AggregateEvents]`
- Event Sourcing Aggregate method can return new instance of State-based Aggregate
- Event Sourcing Aggregate which was called should still be able to record its own events using `#[AggregateEvents]`. Those events should also be persisted with aggregate being called.

The last scenario should be considered as valid as long as both Aggregates (ES and State-based) are persisted in the same transaction.

unixslayer:
resolves #204

- Event Sourcing Aggregate with Internal Recorder can be created by another Aggregate - both ES and State-based
- State-based Aggregate (with optional Internal Recorder) can be created by another Aggregate - both ES and State-based
- called aggregate can record its own Events using Internal Recorder
- events of result and called Aggregates will be published


packages/Dbal/tests/Integration/CreateAggregateByStateBasedAggregateTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $calendarId = '1';
+        $meetingId = '1';
+
+        $ecotone
+            ->withStateFor(new Calendar($calendarId))
+            ->sendCommand(new ScheduleMeeting($calendarId, $meetingId))
+        ;
+
+        self::assertEquals(new Meeting($meetingId), $ecotone->getAggregate(Meeting::class, $meetingId));
+    }
```
dgafka, line 36: Next phase, modifying `Calendar` and returning `new Meeting` scenario to cover.

dgafka (dismissed):

packages/Ecotone/src/Modelling/CallAggregateService.php
```diff
@@ -50,8 +51,9 @@ class CallAggregateService
...
+    private bool $isReturningAggregate;
 
-    public function __construct(InterfaceToCall $interfaceToCall, bool $isEventSourced, ChannelResolver $channelResolver, array $parameterConverterBuilders, array $aroundMethodInterceptors, ReferenceSearchService $referenceSearchService, PropertyReaderAccessor $propertyReaderAccessor, PropertyEditorAccessor $propertyEditorAccessor, bool $isCommand, bool $isFactoryMethod, private EventSourcingHandlerExecutor $eventSourcingHandlerExecutor, ?string $aggregateVersionProperty, bool $isAggregateVersionAutomaticallyIncreased, ?string $aggregateMethodWithEvents)
+    public function __construct(InterfaceToCall $interfaceToCall, bool $isEventSourced, ChannelResolver $channelResolver, array $parameterConverterBuilders, array $aroundMethodInterceptors, ReferenceSearchService $referenceSearchService, PropertyReaderAccessor $propertyReaderAccessor, PropertyEditorAccessor $propertyEditorAccessor, bool $isCommand, bool $isFactoryMethod, private EventSourcingHandlerExecutor $eventSourcingHandlerExecutor, ?string $aggregateVersionProperty, bool $isAggregateVersionAutomaticallyIncreased, ?string $aggregateMethodWithEvents, bool $isReturningAggregate)
```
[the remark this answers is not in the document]
unixslayer: I've changed the behavior of `SaveAggregateService` so it takes the type of returned value into consideration. In order to save also called aggregate, also `CallAggregateService` behavior must be changed a bit. I'll try to refactor its constructor as well ;)

packages/Ecotone/src/Modelling/AggregateMessage.php
```diff
@@ -2,16 +2,57 @@
...
-    public const AGGREGATE_ID     = 'ecotone.modelling.aggregate.id';
-    public const TARGET_VERSION   = 'ecotone.modelling.aggregate.target_version';
+    public const AGGREGATE_ID = 'ecotone.modelling.aggregate.id';
+    public const TARGET_VERSION = 'ecotone.modelling.aggregate.target_version';
+    public const AGGREGATE_EVENTS = 'ecotone.modelling.aggregate_events';
+
+    /**
+     * @return array<Event>
+     */
+    public static function resolveEvents(Message $message, array $metadata, string $calledInterface): array
```
dgafka, line 31: This class was meant to be used only as provider for Message Headers related with this Modelling Module. 
So I think we can move this somewhere else

packages/Ecotone/src/Modelling/CallAggregateService.php
```diff
@@ -1,151 +1,12 @@
...
-use Ecotone\Messaging\Support\Assert;
-use Ecotone\Messaging\Support\MessageBuilder;
 
-/**
- * Class CallAggregateService
- * @package Ecotone\Modelling
- * @author  Dariusz Gafka <dgafka.mail@gmail.com>
- */
-class CallAggregateService
+interface CallAggregateService
```
dgafka, line 9: I don't see this being used anywhere. 
We are binding things not by references, but by messaging, so I think we can drop this.

packages/Dbal/tests/Integration/CreateAggregateByStateBasedAggregateTest.php
```diff
@@ -0,0 +1,73 @@
...
+                new CalendarCreated($calendarId),
+                new MeetingScheduled($calendarId, $meetingId),
+            ],
+            $ecotone->getRecordedEvents()
+        );
+
+        self::assertEquals(new Meeting($meetingId), $ecotone->getAggregate(Meeting::class, $meetingId));
+        self::assertEquals([$meetingId], $ecotone->sendQueryWithRouting('calendar.meetings', metadata: ['aggregate.id' => $calendarId]));
+    }
+}
```
dgafka, line 73: Can we add one more scenario, where Returned Aggregate has internal recorded and publishes an event?

packages/Ecotone/src/Modelling/CallEventSourcingAggregateService.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $events = $result;
+        if ($this->aggregateMethodWithEvents) {
+            $events = call_user_func([$this->isFactoryMethod ? $result : $calledAggregate, $this->aggregateMethodWithEvents]);
+        }
+
+        if ($this->isFactoryMethod) {
+            $resultMessage->setHeader(AggregateMessage::RESULT_AGGREGATE_OBJECT, $this->eventSourcingHandlerExecutor->fill($events, null));
+        } elseif ($this->isReturningAggregate) {
+            $resultMessage->setHeader(AggregateMessage::RESULT_AGGREGATE_OBJECT, $result);
```
dgafka, line 96: What about Query Handler returning an Aggregate?
I think this header is not really useful in that scenario and we could skip, right?

packages/Ecotone/src/Modelling/SaveEventSourcingAggregateService.php
```diff
@@ -0,0 +1,139 @@
...
+use Ecotone\Messaging\Store\Document\InMemoryDocumentStore;
+use Ecotone\Messaging\Support\Assert;
+use Ecotone\Messaging\Support\InvalidArgumentException;
+use Ecotone\Messaging\Support\MessageBuilder;
+use Ramsey\Uuid\Uuid;
+
+final class SaveEventSourcingAggregateService implements SaveAggregateServiceInterface
+{
+    public const NO_SNAPSHOT_THRESHOLD = 0;
+    public const SNAPSHOT_COLLECTION = 'aggregate_snapshots_';
```
dgafka, line 24: I think this split of Services for Event Sourced and State-Based brings a lot of sense. As we had to do if branching logic before and there were things that made sense only in one Context like snapshoting for Event Sourcing Aggregates.

packages/Ecotone/src/Modelling/SaveAggregateServiceBuilder.php
```diff
@@ -144,34 +148,146 @@ public function compile(MessagingContainerBuilder $builder): Definition
...
+            $returnTypeInterface = $interfaceToCallRegistry->getClassDefinitionFor(TypeDescriptor::create($interfaceToCall->getReturnType()->toString()));
+            if (
+                $returnTypeInterface->hasClassAnnotation(TypeDescriptor::create(Aggregate::class))
+                && $returnTypeInterface->getClassType() !== $aggregateClassDefinition->getClassType()
+            ) {
+                $this->isReturningAggregate = true;
+                $this->returnAggregateClassName = $interfaceToCall->getReturnType()?->toString();
+
+                [$this->returnAggregateIdentifierMapping, $this->returnAggregateIdentifierGetMethods] = $this->resolveAggregateIdentifierMapping($returnTypeInterface, $interfaceToCallRegistry);
+                [$this->returnAggregateVersionProperty, $this->isReturnAggregateVersionAutomaticallyIncreased]  = $this->resolveAggregateVersionProperty($returnTypeInterface);
```
dgafka, line 178: What about instead of resolving all of this and making `SaveAggregateService` aware of identifiers and versions of both Aggregates, we would construct here another instance of SaveAggregateService that would make use of identifier and version only for single aggregate?

That would potentially allow us for enriching this solution with saving multiple returning aggregates later.

packages/Ecotone/src/Modelling/CallAggregateService.php
unixslayer: @dgafka By splitting the original class into two there became an indirect relation between them so I wanted to keep an abstraction. With the next part for handling the flow of Pure ES aggregate being called, I may introduce another `CallAggregateService`. If you don't mind, I would keep it, and if it turns out redundant I will remove it.

packages/Ecotone/src/Modelling/CallEventSourcingAggregateService.php
unixslayer: I did not cover that scenario yet. The reason for this header is to keep both called and resulted aggregate in the flow separately for the case when two of those are involved due it will require handling them independently.

packages/Ecotone/src/Modelling/SaveEventSourcingAggregateService.php
unixslayer: yes, that was my intention. previous `CallAggregateService` and `SaveAggregateService` were messy and hard to read.

packages/Ecotone/src/Modelling/SaveAggregateServiceBuilder.php
unixslayer: that would be the case after having some kind of `CompoundResult`. For now, flow covers the scenario of max two aggregates involved in one flow, and before I introduce something more complex I want to cover Pure ES Aggregate case.

dgafka: Ok, let's focus on single aggregate within this PR then :)

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
```diff
@@ -563,9 +570,18 @@ private function registerAggregateCommandHandler(Configuration $configuration, I
                     $baseEventSourcingConfiguration->getDocumentStoreReferenceFor($aggregateClassDefinition->getClassType()->toString())
                 )
                 ->withInputChannelName($saveChannel)
-                ->withOutputMessageChannel($annotation->getOutputChannelName())
+                ->withOutputMessageChannel($publishChannel)
                 ->withAggregateRepositoryFactories($aggregateRepositoryReferenceNames)
             );
+            $configuration->registerMessageHandler(
+                PublishAggregateEventsServiceBuilder::create(
```
dgafka, line 577: What is the reason behind splitting it out?

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
```diff
@@ -691,9 +708,9 @@ private function registerSaveAggregate(ClassDefinition $aggregateClassDefinition
         /** @TODO do not require method name in save service */
         $methodName = $aggregateClassDefinition->getPublicMethodNames() ? $aggregateClassDefinition->getPublicMethodNames()[0] : '__construct';
 
-
         $configuration->registerMessageHandler(
             $chainMessageHandlerBuilder
+                ->chain(new ResolveAggregateEventsServiceBuilder())
```
dgafka, line 713: What is the reason behind splitting it out?

packages/Ecotone/src/Lite/Test/FlowTestSupport.php
```diff
@@ -160,23 +160,23 @@ public function withEventsFor(string|object|array $identifiers, string $aggregat
             [
                 AggregateMessage::OVERRIDE_AGGREGATE_IDENTIFIER => is_object($identifiers) ? (string)$identifiers : $identifiers,
                 AggregateMessage::TARGET_VERSION => $aggregateVersion,
-                AggregateMessage::AGGREGATE_OBJECT => $aggregateClass,
+                AggregateMessage::CALLED_AGGREGATE_OBJECT => $aggregateClass,
+                AggregateMessage::RESULT_AGGREGATE_OBJECT => $aggregateClass,
```
dgafka, line 164: Why do we add result aggregate here? Shouldn't this header be only added, when there is we produce one-aggregate from another? And in this case we add events for a single aggregrate?

Blind guess as have no checked that, but the test that fails is making use of withEventsFor, so maybe this is somehow related?

packages/Ecotone/src/Modelling/CallStateBasedAggregateService.php
```diff
@@ -0,0 +1,96 @@
...
+            $result = $methodInvokerChainProcessor->proceed();
+        } else {
+            $result = $methodInvoker->executeEndpoint($message);
+        }
+
+        $resultType = TypeDescriptor::createFromVariable($result);
+        if ($resultType->isIterable() && $this->interfaceToCall->getReturnType()?->isCollection()) {
+            $interfaceReturnType = $this->interfaceToCall->getReturnType();
+            if ($interfaceReturnType?->isUnionType()) {
+                $resultType = TypeDescriptor::createCollection(TypeDescriptor::OBJECT);
```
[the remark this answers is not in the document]
unixslayer: @dgafka turned out that this was redundant from beginning because `$interfaceReturnType` is already correct for union type being returned.

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
unixslayer: @dgafka I wanted to split each operation that happens within a flow into separate steps chained by `ModelingHandlerModule` and make use of the internal message being sent from one step to another. Flow differs depending on what was called, how, and what is returned. For example, the direct repository gateway `CallAggregateService` was never used and after changes made in `SaveAggregateService` which no longer depends directly on a result, I introduced `ResolveAggregateEventsService`. 

This way I achieved a configurable flow that does not alternate results being returned from the business interface.

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
unixslayer: check previous answer to the same question

packages/Ecotone/src/Lite/Test/FlowTestSupport.php
unixslayer: No, this is actually correct because it is related to how the aggregate state is being initialized in tests. `SaveAggregateService` that is being called gets configured that it works on resulted aggregate, not the one being called. I think that `AggregateMessage::CALLED_AGGREGATE_OBJECT` is redundant here.

packages/Ecotone/src/Messaging/Handler/InterfaceToCall.php
```diff
@@ -431,4 +432,20 @@ private function parameterAmount(): int
...
+
+            return $returnTypeInterface->hasClassAnnotation(TypeDescriptor::create(Aggregate::class));
+        }
+
+        return false;
+    }
+
+    public function isFactoryMethod(): bool
+    {
+        return $this->isStaticallyCalled || $this->methodName === '__construct';
```
dgafka, line 449: Ecotone does not support `__construct` based aggregate creation. I guess this can lead to bugs if someone would put attribute on the constructor. 

Can we make define factory method just as a static method?

packages/Ecotone/src/Modelling/ResolveAggregateServiceBuilder.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ecotone\Modelling;
```
dgafka, line 5: We do have a of classes in namespace, can we use package design to split?

So keeping the interfaces on `Ecotone\Modelling` level as API, and moving related implementations to higher namespaces e.g. `Ecotone\Modelling\SaveAggregate`, `Ecotone\Modelling\ResolveAggregate`

packages/PdoEventSourcing/tests/Integration/CreateAggregateByEventSourcingAggregateTest.php
```diff
@@ -0,0 +1,96 @@
...
+            )
+            ->sendCommand(new ScheduleMeeting($calendarId, $meetingId))
+        ;
+
+        self::assertEquals([
+            new MeetingCreated($meetingId, $calendarId),
+        ], $ecotone->getRecordedEvents());
+
+        $meeting = new Meeting($meetingId, $calendarId);
+        $meeting->getRecordedEvents();
```
dgafka, line 54: Is this some debug left over? :)

packages/PdoEventSourcing/tests/Integration/CreateAggregateByEventSourcingAggregateTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+        self::assertEquals($meeting, $ecotone->getAggregate(Meeting::class, $meetingId));
+
+        // todo pure ES aggregate should be able to record events along side with returning another aggregate
```
dgafka, line 58: Well not really related with this scenario :D 
We can remove this TODO, I am sure we will be aware that this feature is missing atm :)

packages/Ecotone/src/Modelling/PublishAggregateEventsService.php
```diff
@@ -0,0 +1,81 @@
...
+            if ($eventDefinition->hasClassAnnotation($namedEvent)) {
+                /** @var NamedEvent $namedEvent */
+                $namedEvent = $eventDefinition->getSingleClassAnnotation($namedEvent);
+
+                $this->eventBus->publishWithRouting($namedEvent->getName(), $event->getPayload(), MediaType::APPLICATION_X_PHP, $event->getMetadata());
+            }
+        }
+    }
+
+    // fixme duplicated
```
dgafka, line 54: Still valid?

packages/PdoEventSourcing/tests/Integration/CreateAggregateByEventSourcingAggregateTest.php
unixslayer: Actually not. `Meeting` records an event when constructor is called so for assertion I had to call `getRecordedEvents` so internal recorder gets empty.

packages/Ecotone/src/Modelling/PublishAggregateEventsService.php
unixslayer: sort of. I guess I've changed a bit this method regarding how it looks like in `ResolveAggregateEventsService` but still most of them are the same. I was planning to do cleanup refactoring with next PR.

packages/Ecotone/src/Modelling/ResolveAggregateServiceBuilder.php
unixslayer: I was thinking about it during implementation to organize the code a bit. However, I was hoping to finish and refactor later.

packages/Ecotone/tests/Modelling/Unit/PublishAggregateEventsServiceBuilderTest.php
```diff
@@ -0,0 +1,64 @@
...
+use Ecotone\Modelling\PublishAggregateEventsServiceBuilder;
+use Ecotone\Modelling\StorageEventBus;
+use Ecotone\Test\ComponentTestBuilder;
+use PHPUnit\Framework\TestCase;
+use Test\Ecotone\Modelling\Fixture\Ticket\Ticket;
+use Test\Ecotone\Modelling\Fixture\Ticket\TicketWasStartedEvent;
+
+final class PublishAggregateEventsServiceBuilderTest extends TestCase
+{
+    public function test_publishing_events(): void
```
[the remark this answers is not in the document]
unixslayer: Yes, it is covered already by high-level tests. I introduced a bunch of unit tests to cover other flow services. I'll check coverage and remove them.

packages/Ecotone/src/Messaging/Handler/InterfaceToCall.php
unixslayer: Currently, Ecotone registers `LoadAggregateService` and `SaveAggregateService` for the aggregate repository channels and repository gateways by using `__construct` as a method name. 

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php:701
packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php:723

I don't really see any harm in this because `__construct` is only used to register a gateway for aggregate if the factory method is not defined in the aggregate. 

I also think that it can be pointed out to not use `#[CommandHandler]` attribute on `__construct`.

unixslayer: @dgafka I've made additional validation in `ModelingHandlerModule`. If the constructor is annotated as `Command`, `Query` or `Event` handler, exception will be thrown on bootstrap.

unixslayer: I decided to check any magic method

dgafka: Ok, cool

PR #206, opened 2023-09-23
Refactor around method invocation

jlabedo:
It will higher a bit the stack trace depth but seems cleaner, notably in the CallAggregate service.


dgafka:
Looks good, should be more readable. Thanks :)

PR #207, opened 2023-09-23
allowing AMQP transport configured via Symfony

m2mtech:
Exploring the usage of an AMQP transport configured via Symfony, I encountered a type mismatch error:

```plaintext
In SymfonyMessengerMessageChannel.php line 38:
count(): Argument #1 ($value) must be of type Countable|array, Generator given
```

It seems the [`AmqpReceiver`](https://github.com/symfony/symfony/blob/bf2d7fac81de2c6ef06b20aa210187ae2a7789cd/src/Symfony/Component/Messenger/Bridge/Amqp/Transport/AmqpReceiver.php#L39-L42) is providing a generator, while the other bridge implementations return simple arrays, e.g. the [`DoctrineReceiver`](https://github.com/symfony/symfony/blob/bf2d7fac81de2c6ef06b20aa210187ae2a7789cd/src/Symfony/Component/Messenger/Bridge/Doctrine/Transport/DoctrineReceiver.php#L65).

I have adjusted the `SymfonyMessengerMessageChannel.php` to handle both `Generator` and `array` types to resolve this type mismatch.

The first two unit test worked already for the original codes. The additional tests needed the changes to pass.

Cudos for this library!

### DIFF
--- a/packages/Symfony/SymfonyBundle/Messenger/SymfonyMessengerMessageChannel.php
+++ b/packages/Symfony/SymfonyBundle/Messenger/SymfonyMessengerMessageChannel.php
@@ -7,7 +7,9 @@ namespace Ecotone\SymfonyBundle\Messenger;
 use Ecotone\Messaging\Message;
 use Ecotone\Messaging\PollableChannel;
 use Ecotone\Messaging\Support\Assert;
+use Generator;
 use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Transport\Receiver\MessageCountAwareInterface;
 use Symfony\Component\Messenger\Transport\TransportInterface;
 
 final class SymfonyMessengerMessageChannel implements PollableChannel
@@ -35,7 +37,12 @@ final class SymfonyMessengerMessageChannel implements PollableChannel
             return null;
         }
 
-        Assert::isTrue(count($symfonyEnvelope) === 1, 'Symfony messenger transport should be configured to return only one message at a time');
+        if (!is_array($symfonyEnvelope) && $symfonyEnvelope instanceof Generator && $this->symfonyTransport instanceof MessageCountAwareInterface) {
+            Assert::isTrue($this->symfonyTransport->getMessageCount() === 1, 'Symfony messenger transport should be configured to return only one message at a time');
+            $symfonyEnvelope = iterator_to_array($symfonyEnvelope);
+        } else {
+            Assert::isTrue(count($symfonyEnvelope) === 1, 'Symfony messenger transport should be configured to return only one message at a time');
+        }
 
         return $this->symfonyMessageConverter->convertFromSymfonyMessage(
             $symfonyEnvelope[0],
--- a/packages/Symfony/SymfonyBundle/Messenger/SymfonyMessengerMessageChannel.php
+++ b/packages/Symfony/SymfonyBundle/Messenger/SymfonyMessengerMessageChannel.php
@@ -9,7 +9,6 @@ use Ecotone\Messaging\PollableChannel;
 use Ecotone\Messaging\Support\Assert;
 use Generator;
 use Symfony\Component\Messenger\Envelope;
-use Symfony\Component\Messenger\Transport\Receiver\MessageCountAwareInterface;
 use Symfony\Component\Messenger\Transport\TransportInterface;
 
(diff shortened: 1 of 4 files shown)


m2mtech:
... while the changes work for the test, they do not for an actual project ... submitted to early ;)

m2mtech:
I found no way to configure the amqp transport via symfony messenger to provide only a single message :(

packages/Symfony/SymfonyBundle/Messenger/SymfonyMessengerMessageChannel.php
```diff
@@ -35,7 +36,14 @@ public function receive(): ?Message
...
         }
 
-        Assert::isTrue(count($symfonyEnvelope) === 1, 'Symfony messenger transport should be configured to return only one message at a time');
+        if ($symfonyEnvelope instanceof Generator) {
+            $symfonyEnvelope = iterator_to_array($symfonyEnvelope);
+            if ($symfonyEnvelope === []) {
+                return null;
+            }
+        } else {
+            Assert::isTrue(count($symfonyEnvelope) === 1, 'Symfony messenger transport should be configured to return only one message at a time');
```
dgafka, line 45: > I found no way to configure the amqp transport via symfony messenger to provide only a single message :(

What, if we fetch and return from the iterator the first Message.
Will calling `$this->symfonyTransport->get();` for the second time return next message correctly?

dgafka: And is this a valid scenario to have multiple messages in the generator? It looks like due to internal implementation that use of  `get` instead of `consume`. https://github.com/symfony/symfony/discussions/47160 it should prefetch always only one?

packages/Symfony/tests/phpunit/SymfonyMessengerMessageChannelUnitTest.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private function singleMessageGenerator(): iterable
+    {
+        yield $this->envelope;
+    }
+
+    public function testSymfonyEnvelopeReturnsGenerator(): void
+    {
+        $transport = $this->createMock(AmqpTransport::class);
```
dgafka, line 65: Mock kind of tests can provide false positives, as there is a lot of things happening under the hood that may actually fail. 

Are you able to provide real integration test for this, where we send message and receive the message?

m2mtech:
I added integration tests. Sending multiple commands, `$this->messaging->run()` fetches only a single command from the queue. Subsequent calls fetch the remaining commands.

packages/Symfony/SymfonyBundle/Messenger/SymfonyMessengerMessageChannel.php
dgafka: > I added integration tests. Sending multiple commands, $this->messaging->run() fetches only a single command from the queue. Subsequent calls fetch the remaining commands.

@m2mtech ok cool :) 

Can we move this assert outside of `else` and remove `testSymfonyEnvelopeReturnsGeneratorFailsBecauseOfTooManyMessages` test case as it will never happen.
After that I am happy to merge :)

m2mtech: I don't think so. The amqp transport seems to provide always all available messages, where the other symfony transports return an array with a single element. It is your code which can cope with it ;)

dgafka: Ok, I see. 
So let's remove `testSymfonyEnvelopeReturnsGeneratorFailsBecauseOfTooManyMessages` test case. We are handling that and confirming it works with integration test already.

m2mtech: Done, I opened a [Symfony discussion](https://github.com/symfony/symfony/discussions/51738#discussion-5668125) for clarification. It seems the [`ReceiverInterface`](https://github.com/symfony/symfony/blob/fbf6f56ca7321e28d9a4368e18b9da683c296046/src/Symfony/Component/Messenger/Transport/Receiver/ReceiverInterface.php#L46) intended to return multiple messages.

dgafka:
Thanks for the contribution :)

PR #210, opened 2023-09-25
add `void` return type to `CacheCleaner::clear()` and `EcotoneCompilerPass::process()`

someniatko:
this fixes deprecation message from Symfony and ensures supporting Symfony 7.x in future


dgafka:
Great, thank you :)

PR #211, opened 2023-09-25
allow aggregates to have different snapshot configurations

unixslayer:


packages/Ecotone/src/Modelling/BaseEventSourcingConfiguration.php
```diff
@@ -2,37 +2,59 @@
...
-        return $this->snapshotsAggregateClasses;
+        $this->snapshotsAggregateClasses[$aggregateClassToSnapshot] = [
+            'thresholdTrigger' => $thresholdTrigger,
+            'documentStore' => $documentStore,
+        ];
+
+        return $this;
+    }
+
+    public function buildDocumentStoreReferences(ReferenceSearchService $referenceSearchService): array
```
dgafka, line 36: This will become public API for end users, even so it's meaningful only during the build phase. 
Let's keep this logic internal inside `EventSourcingRepositoryBuilder`.

dgafka (approved):

packages/Ecotone/src/Modelling/BaseEventSourcingConfiguration.php
```diff
@@ -2,37 +2,59 @@
...
 class BaseEventSourcingConfiguration
 {
     public const DEFAULT_SNAPSHOT_TRIGGER_THRESHOLD = 100;
 
     private array $snapshotsAggregateClasses = [];
-    private int $snapshotTriggerThreshold = self::DEFAULT_SNAPSHOT_TRIGGER_THRESHOLD;
-    private string $documentStoreReference = DocumentStore::class;
 
+    /**
+     * @deprecated use BaseEventSourcingConfiguration::withSnapshotsFor instead
```
dgafka, line 15: I annotate B/C with `@TODO Ecotone 2.0 drop it` comment. So we can easily find it when the time will come ;)

PR #212, opened 2023-09-27
refactor: resolve interceptors parameter converters during build

jlabedo:


packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/AroundInterceptorReference.php
```diff
@@ -163,13 +173,28 @@ public function buildAroundInterceptor(ReferenceSearchService $referenceSearchSe
...
+                $builtConverters[] = ValueConverter::createWith($parameter->getName(), null);
+                continue;
+            }
+            if (!$hasPayloadConverter) {
+                $builtConverters[] = PayloadBuilder::create($parameter->getName())->build($referenceSearchService);
+                $hasPayloadConverter = true;
+                continue;
+            } elseif ($parameter->getTypeDescriptor()->isNonCollectionArray()) {
+                $builtConverters[] = AllHeadersBuilder::createWith($parameter->getName())->build($referenceSearchService);
+                continue;
```
dgafka, line 187: This will simplify the code, yet the performance will stay the same, as build method is executed on run time.
Whatever has `ReferenceSearchService $referenceSearchService`, means it has to work booted container.

dgafka: or maybe it will help, if given class is executed more than once. As we will then reuse constructed parameters

jlabedo: Yes I know :) the plan is to do all this work before something is serialized for caching, but I am not sure at that time of how to do it. My goal is to try something wich would work with incremental refactorings to demonstrate the performance gain.
And whatever you cache, the core messaging components should have low construction cost to show the performance boost.

PR #213, opened 2023-09-27
Refactor: parameter converter interface

jlabedo:


dgafka:
@jlabedo can you rebase?

jlabedo:
Quick question @dgafka while you are here, I can inject the intercepted aggregate in an around interceptor, but not in a before or after interceptor. Is it expected ? Or maybe it is due to a previous regression ?

dgafka:
> Quick question @dgafka while you are here, I can inject the intercepted aggregate in an around interceptor, but not in a before or after interceptor. Is it expected ? Or maybe it is due to a previous regression ?

It's correct behaviour. Before and After are separate chained Message Handlers (endpoints), therefore they do not have access to executed object. 
Where `Around Interceptor` works closely to executed object, as we get access to execute it's method using `MethodInvocation`.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInvoker.php
```diff
@@ -57,8 +57,8 @@ public static function createWith(InterfaceToCall $interfaceToCall, $objectToInv
     {
         $methodParametersConverterBuilders = MethodArgumentsFactory::createDefaultMethodParameters($interfaceToCall, $methodParametersConverterBuilders, $endpointAnnotations, null, false);
         $methodParameterConverters         = [];
-        foreach ($methodParametersConverterBuilders as $methodParameter) {
-            $methodParameterConverters[] = $methodParameter->build($referenceSearchService);
+        foreach ($methodParametersConverterBuilders as $index => $methodParameter) {
+            $methodParameterConverters[] = $methodParameter->build($referenceSearchService, $interfaceToCall, $interfaceToCall->getParameterAtIndex($index));
```
dgafka, line 61: This will end up initializing all the method parameters during kernel boot.
In case of `Asynchronous Message Consumers` that may not be a problem, yet in case of `HTTP req/res`, where we execute only specific Command or Query, we will need to boot all the parameters for all the endpoints even so they are not used. 
You can take a look on the benchmark time, this increased the time a lot even for small code base.

dgafka: This would be fine, if MethodInvoker would be constructed during [preparation phase](https://docs.ecotone.tech/messaging/contributing-to-ecotone/ecotone-run-time#preparation-phase).

Taking as an example ServiceActivatorBuilder, creation of [MethodInvoker](https://github.com/ecotoneframework/ecotone-dev/blob/main/packages/Ecotone/src/Messaging/Handler/ServiceActivator/ServiceActivatorBuilder.php#L165) would need to be moved to the constructor somehow :)

jlabedo: > You can take a look on the benchmark time, this increased the time a lot even for small code base.

Which benchmark ? I am not aware of having changed something regarding that matter.

packages/Ecotone/src/Messaging/Handler/ParameterConverterBuilder.php
```diff
@@ -26,5 +26,5 @@ public function isHandling(InterfaceParameter $parameter): bool;
      * @param ReferenceSearchService $referenceSearchService
      * @return ParameterConverter
      */
-    public function build(ReferenceSearchService $referenceSearchService): ParameterConverter;
+    public function build(ReferenceSearchService $referenceSearchService /*, ?InterfaceToCall $interfaceToCall = null, ?InterfaceParameter $interfaceParameter = null*/): ParameterConverter;
```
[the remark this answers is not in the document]
jlabedo: Fixed

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInvoker.php
dgafka: [benchmark](https://github.com/ecotoneframework/ecotone-dev/actions/runs/6335722883/job/17207395180), but the times seems to be pretty much similar, hmm.

I think to give this PR, a second thought.

dgafka: As far as I can see it's only refactor, so it should not affect performance anyhow

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodArgumentsFactory.php
```diff
@@ -63,8 +68,24 @@ public static function createDefaultMethodParameters(InterfaceToCall $interfaceT
         return $orderedMethodArguments;
     }
 
+    public static function getAnnotationValueConverter(InterfaceParameter $interfaceParameter, InterfaceToCall $interceptedInterface, array $endpointAnnotations): ?ValueBuilder
+    {
+        $allAnnotations = [...$interceptedInterface->getClassAnnotations(), ...$interceptedInterface->getMethodAnnotations(), ...$endpointAnnotations];
```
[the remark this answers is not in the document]
jlabedo: Changed to `array_merge`. Is there a difference between the two ?

packages/Ecotone/src/Messaging/Handler/ParameterConverterBuilder.php
```diff
@@ -26,5 +26,5 @@ public function isHandling(InterfaceParameter $parameter): bool;
      * @param ReferenceSearchService $referenceSearchService
      * @return ParameterConverter
      */
-    public function build(ReferenceSearchService $referenceSearchService): ParameterConverter;
+    public function build(ReferenceSearchService $referenceSearchService, ?InterfaceToCall $interfaceToCall = null, ?InterfaceParameter $interfaceParameter = null): ParameterConverter;
```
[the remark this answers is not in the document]
jlabedo: Fixed.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodArgumentsFactory.php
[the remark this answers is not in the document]
dgafka: Nothing I am aware of. Just we are using array_merge across the code base

Issue #219, opened 2023-10-03
Absolute path to the cache dir seems to be ... cached, using Symfony bundle


someniatko:
**Ecotone version(s) affected**: 1.106.0

**Description**  
Let's continue discussing cache-related stuff!

**How to reproduce**  
1. Set up a project in directory A with Symfony 6.3 + Ecotone + Ecotone Symfony Bundle. Set `APP_ENV=prod`, `APP_DEBUG=false` in ENV. Assume default Symfony cache configuration.
2. Have some code in a Symfony controller which invokes Ecotone messaging in some way.
3. Remove `var/cache/prod` dir.
4. Execute `php bin/console cache:clear`, `php bin/console cache:warmup`.
5. Set up a web server and make an HTTP request which would invoke your controller. All works fine.
6. Move or copy your project directory somewhere else (alongside with var/cache/, vendor/ etc)
7. Again, set up a web server and try making an HTTP request, which should now land in the new dir.

Now it won't work as in step 5, and you'll get a following exception:
```
ProxyManager\Exception\InvalidProxyDirectoryException
Provided directory "OLD-PROJECT-DIR/var/cache/prod/ecotone" does not exist
```

**Possible Solution** 

I took the dir of a project I'm working on from its CI build (which includes vendor/, cache/ etc) and ran a webserver pointing to it locally on my computer.
I've added the following to `Ecotone\Messaging\Handler\Gateway\ProxyFactory::__construct()`:

```php
    private function __construct(?string $cacheDirectoryPath)
    {
        $this->cacheDirectoryPath = $cacheDirectoryPath;
        echo "CACHE DIRECTORY PATH: {$this->cacheDirectoryPath}\n";
        debug_print_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
        echo "\n\n";
    }
```

Here's the output: https://gist.github.com/someniatko/bbaccca66eea61c48245a09b913db74b

It seems Ecotone actually creates the ProxyFactory twice, once for the cached version of MessagingSystemConfiguration, and the second for some other stuff. This cached MessagingSystemConfiguration I presume contains serialized PHP representation of the class, which also happens to contain the absolute path to the cache dir, which makes the project dir non-portable.
UPD: it's part of `ModuleReferenceSearchService::$moduleReferences['gatewayProxyConfiguration']` which is a `ProxyFactory` having `cacheDirectoryPath`.

**Context**  
[...]

PR #220, opened 2023-10-03
update cache directory path after unserializing MessagingSystemConfiguration

someniatko:
should fix https://github.com/ecotoneframework/ecotone-dev/issues/219

### DIFF
--- a/packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
+++ b/packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
@@ -817,7 +817,15 @@ final class MessagingSystemConfiguration implements Configuration
 
         $messagingSystemCachePath = $cacheDirectoryPath . DIRECTORY_SEPARATOR . 'messaging_system';
         if (file_exists($messagingSystemCachePath)) {
-            return unserialize(file_get_contents($messagingSystemCachePath));
+            $result = unserialize(file_get_contents($messagingSystemCachePath));
+            \assert($result instanceof MessagingSystemConfiguration);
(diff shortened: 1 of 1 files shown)


packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
```diff
@@ -1360,4 +1368,12 @@ private function registerAutoloader(AutoloaderInterface $autoloader)
         spl_autoload_register($autoloader);
         self::$registered_autoloader = $autoloader;
     }
+
+    private function remapCacheDirectoryPath(string $cacheDirectoryPath): void
+    {
+        $this->moduleReferenceSearchService->store(
+            ProxyFactory::REFERENCE_NAME,
+            ProxyFactory::createWithCache($cacheDirectoryPath),
```
dgafka, line 1376: The proxy code needs some love and I would like to do refactor to solve this problem, which potentially can also increase the performance. 

I've started looking at the code to solve that. I will let you know when I will have it done :)

someniatko: Yeah, I suspected this is not the only place where it might come up. At least it works for my use-case, so I'll leave it in my fork temporarily.

Thank you so much!

someniatko: Feel free to close this PR if you are going to make a better implementation!

dgafka:
Closing as will be fixed with #221. Thanks for putting effort into solving this :)

PR #221, opened 2023-10-04
Refactor proxy generation

dgafka:
- Enable way to dynamically switch cache directory for Gateway Proxies and JMS.
- Do not use different proxies inside Messaging Configuration and Symfony/Laravel
- In case proxies are already registered (Symfony/Laravel) reuse them instead of booting new

This also will respect the configuration fully now. If using cache is disabled, it will not prepare cache directory, dump messaging configuration or generate proxy files / JMS metadata files. This means that development will become more smooth as will not require cleaning cache directory in some situations and will not do unnecessary things.
To achieve the above in case of Symfony, new configuration parameter was introduced.


packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
```diff
@@ -134,15 +136,27 @@ public function process(ContainerBuilder $container): void
...
+        $container->setDefinition(ServiceCacheDirectory::REFERENCE_NAME, $definition);
+
         foreach ($messagingConfiguration->getRegisteredGateways() as $gatewayProxyBuilder) {
+            // Proxy warm up
+            ProxyFactory::createFor(
+                $gatewayProxyBuilder->getReferenceName(),
+                $container,
+                $gatewayProxyBuilder->getInterfaceName(),
+                $container->getParameter('kernel.cache_dir') . self::CACHE_DIRECTORY_SUFFIX
+            );
```
dgafka, line 151: In case of Symfony we can do it in here, as this `process` method will be only called once to compile container.
I've dropped warming up for `Laravel` and `Lite` tho. As there is no clear distinction to the process. 
Happy to accept an improvement if there is a way to do so.

dgafka:
Aiming to solve and improve the code base readability and performance #219

dgafka:
@unixslayer @jlabedo can you take a look on the changes?

packages/Ecotone/src/Messaging/Handler/Gateway/GatewayProxyBuilder.php
```diff
@@ -318,19 +320,30 @@ public function getEndpointAnnotations(): array
...
+
+        $adapter = new GatewayProxyAdapter(
+            NonProxyCombinedGateway::createWith(
+                $this->referenceName,
+                $this->interfaceName,
+                [$this->getRelatedMethodName() => $this->buildWithoutProxyObject(
+                    $referenceSearchService, $channelResolver
+                )]
+            )
+        );
```
jlabedo, line 339: If I understand correctly, you are passing a combined gateway to the proxy, but with only one gateway ?
Is it for cleaning the way for a future refactoring with one proxy per interface with all its gateways ?

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
```diff
@@ -69,77 +54,20 @@ public function getConfiguration(): Configuration
...
         $factory = new RemoteObjectFactory($adapter, $this->getConfiguration());
 
         return $factory->createProxy($interfaceName);
     }
 
-    /**
-     * @inheritDoc
-     */
-    public function serialize(): string
+    public static function createFor(string $referenceName, ContainerInterface $container, string $interface, ServiceCacheConfiguration $serviceCacheConfiguration): object
```
jlabedo, line 64: The `createFor` command is creating a new proxyFactory with a new `Configuration` on each call. Is it necessary ? Maybe this method should not be static.

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
```diff
@@ -61,11 +61,11 @@ public static function getMessagingConfiguration(ContainerInterface $container,
         /** @TODO Ecotone 2.0 use ServiceContext to configure Symfony */
         $serviceConfiguration = ServiceConfiguration::createWithDefaults()
             ->withEnvironment($container->getParameter('kernel.environment'))
-            ->withFailFast($container->getParameter('kernel.environment') === 'prod' ? false : $container->getParameter(self::FAIL_FAST_CONFIG))
+            ->withFailFast(in_array($container->getParameter('kernel.environment'), ['prod', 'production']) ? false : $container->getParameter(self::FAIL_FAST_CONFIG))
```
someniatko, line 64: hmm, does it mean that if I want this "fail fast" behavior in production for some reason, I cannot enable it explicitly in the config? I know it was introduced earlier and this PR just extends on existing behavior, but anyway. Perhaps generating a default config for prod environment with this set to false would suffice?

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
```diff
@@ -134,15 +135,31 @@ public function process(ContainerBuilder $container): void
         $definition->addArgument(new Reference('service_container'));
         $container->setDefinition(ReferenceSearchService::class, $definition);
 
+        $useCache = in_array($container->getParameter('kernel.environment'), ['prod', 'production']) ? true : $container->getParameter(self::CACHE_CONFIGURATION);
```
someniatko, line 138: and same here, it seems there is no option of disabling the cache for production even if I want to?

someniatko: BTW, a bit of bikeshedding here, maybe it's better to call the config param `useCache` or `enableCache`? `cacheConfiguration` from my intuition like an object with a lot of params to set up, not as a `true` / `false` value.

dgafka: > useCache or enableCache? cacheConfiguration from my intuition looks like an object with a lot of params to set up, not like a true / false value.

Internally we use `useCache` naming :P.
I agree, but this is something that we can change with major version, as this configuration is already [available for Laravel under this name](https://docs.ecotone.tech/modules/laravel-ddd-cqrs-event-sourcing#cacheconfiguration), so I would like to keep consistency within this version.

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
dgafka: Sure we can make it default only if was not set up. 
Would you be willing to contribute that after this PR is merged?

someniatko: Yeah, I can do this change myself

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
someniatko: Before changing in the major version, we should introduce the new one, and deprecate the old one.

I am not familiar with how it's done in Laravel though, I can contribute only for the Symfony Bundle.

packages/Ecotone/src/Messaging/Handler/Gateway/GatewayProxyBuilder.php
dgafka: It's just to reuse same functionality. `GatewayProxyAdapter` is also used in MessagingSystemConfiguration and it uses multiple methods, so this make use of same code for readability. 

The `build` method in reality now is only used in tests, as it does not bring the value due to constructing single method proxy. Probably `buildWithoutProxyObject`, should replace `build` and GatewayProxyBuilder should not be in knowledge of proxy classes. However that requires a bit of refactor :)

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
dgafka: If you take a look on `packages/Laravel/config/ecotone.php` you will find the config there and it's used in `EcotoneProvider`. 
It's even simpler than Symfony config :P

dgafka: > Before changing in the major version, we should introduce the new one, and deprecate the old one.

Do you want to support both at once? Should we throw exception if both are set up instead of one of them?

someniatko: Yeah, I think supporting both in 1.x is the only smooth decision for renaming it in 2.x.

> Should we throw exception if both are set up instead of one of them?

I didn't think about it :D. Makes sense!

PR #223, opened 2023-10-05
PoC: compile messaging system into a container

jlabedo:
This is a proof of concept of Solution 2 discussed in #222

### Performance
Looking at the benchmark we have a 2,8x perf boost on symfony and 3,1x with ecotone lite app. Regarding laravel, It is not well optimized, I may need some help for improving it. Regarding dev mode benchmarks, the implementation is clearly optimized for prod now. I can already see some potential improvements : interfaces to call are eagerly resolved and dumped to a definition in memory, then rehydrated when needed. But I would like not to do it in this PR ;)

#### 1. Channel adapters and polling consumers.
A refactoring was needed because the build was depending on ExecutionPollingMetadata given at runtime. It was quite hard to follow the used abstractions and here is my understanding of the existing implementation:
the `PollingConsumerBuilder` creates an `InboundChannelAdapter` which is a `ConsumerLifecycle`.
A channel adapter can be used in two ways: As a consumer or as a `TaskExecutor` adapter in a message channel adapter.

So I tried to use the same abstractions for Channel adapter and PollingConsumerBuilder. I created the `MessagePoller` abstraction that has a single method `receiveWithTimeout` (a `PollableChannel` would then implement `MessageChannel` and `MessagePoller`). A `ChannelAdapter` is now implementing the `MessagePoller` interface, so its single responsibility is to return an Ecotone `Message` or `null`.

I also created the `EndpointRunner` abstraction which takes ExecutionPollingMetadata to run something.  The only interesting implementation is `InterceptedConsumerRunner` which takes polling metadata to create an intercepted `ConsumerLifecycle`.

Now enqueue channel adapters are also extending the abstract class `InterceptedChannelAdapterBuilder` that can adapt a `MessagePoller` into an `EndpointRunner` that sends the polled messages into a gateway (`PollToGatewayTaskExecutor`)

#### 2. Event sourcing module
Quite hard to follow the flow of building components here as there are a lot of abstractions. Some components have been transformed to singleton notably `LazyProophEventStore` where two instances exists: one for aggregate events and an other for event stream emitter. Tests are passing but I can't say I know what I am doing here, so it would be appreciated to have a special attention reviewing this part.

### Missing work
- ~refactor: ContainerImplementation abstraction~
- ~refactor: compiler passes code duplication~
- [x] remove hard coded strings
- [x] test: Userland attribute with an object as an argument (failing now)
- [x] test: removing PollingMetadata header added on polling consumer messages
- [x] Ecotone Lite caching
- [x] required: Remove all `WARNING: Argument is not a self defined object` for production (not test) component as it will break caching containers
- [x] Your review ;)


packages/Ecotone/src/Messaging/Config/Container/Implementations/PhpDiContainerImplementation.php
```diff
@@ -0,0 +1,100 @@
...
+use DI\ContainerBuilder;
+use Ecotone\Messaging\Config\Container\ContainerImplementation;
+use Ecotone\Messaging\Config\Container\Definition;
+use Ecotone\Messaging\Config\Container\Reference;
+
+use function is_array;
+
+use ReflectionMethod;
+
+class PhpDiContainerImplementation implements ContainerImplementation
```
jlabedo, line 14: This is for the Php-Di implementation. It should be in EcotoneLiteApplication package.

It is there because one of the other solution that I considered was using php-di (or another container) directly into ecotone core package. Ecotone would have been responsible for dumping its dedicated container for messaging. In this case there would have been two containers (the one of the host framework, and the one of ecotone).

I find the solution demonstrated in this PoC (having Ecotone register components in any host's container) better.

packages/Ecotone/src/Messaging/Config/EcotoneRemoteAdapter.php
```diff
@@ -15,11 +16,8 @@ public function __construct(private ContainerInterface $container, private strin
...
     {
-        /** @var MessagingSystem $messagingSystem */
-        $messagingSystem = $this->container->get(ConfiguredMessagingSystem::class);
-
-        $nonProxyCombinedGateway = $messagingSystem->getNonProxyGatewayByName($this->referenceName);
-
-        return $nonProxyCombinedGateway->executeMethod($method, $params);
+        /** @var Gateway $gateway */
+        $gateway = $this->container->get('gateway.'.$this->referenceName.'::'.$method);
+        return $gateway->execute($params);
```
jlabedo, line 21: This is how you get lazy loading for free :)

packages/Ecotone/src/Messaging/Config/MessagingSystemContainer.php
```diff
@@ -0,0 +1,94 @@
...
+use Ecotone\Messaging\MessageChannel;
+use Ecotone\Messaging\MessagePublisher;
+use Ecotone\Messaging\Support\Assert;
+use Ecotone\Modelling\CommandBus;
+use Ecotone\Modelling\DistributedBus;
+use Ecotone\Modelling\EventBus;
+use Ecotone\Modelling\QueryBus;
+use Psr\Container\ContainerInterface;
+
+class MessagingSystemContainer implements ConfiguredMessagingSystem
```
jlabedo, line 17: The new `ConfiguredMessagingSystem` do nothing more than providing a dedicated access to the container.

packages/Ecotone/src/Lite/LiteContainerImplementation.php
```diff
@@ -0,0 +1,71 @@
...
+class LiteContainerImplementation implements ContainerImplementation
+{
+    public function __construct(private InMemoryPSRContainer $container, private ?ContainerInterface $externalContainer = null)
+    {
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function process(array $definitions, array $externalReferences): void
```
dgafka, line 21: External references are Services from DI, which were provided by `ReferenceSearchService`?

packages/Ecotone/src/Lite/LiteContainerImplementation.php
```diff
@@ -0,0 +1,71 @@
...
+                return $factory(...$arguments);
+            } else {
+                $class = $argument->getClassName();
+                $object = new $class(...$arguments);
+                foreach ($argument->getMethodCalls() as $methodCall) {
+                    $object->{$methodCall->getMethodName()}(...$this->resolveArgument($methodCall->getArguments(), $definitions));
+                }
+                return $object;
+            }
+        } else if ($argument instanceof Reference) {
```
dgafka, line 50: I wonder about mixing References with Definitions under same collection. 
Reference is not something we process, we just retrieve it from the external container, yet we treat it like it would be the case.
Do you think we can improve the abstraction here?

packages/Ecotone/src/Lite/Test/Configuration/InMemoryRepositoryBuilder.php
```diff
@@ -50,11 +54,30 @@ public function isEventSourced(): bool
...
-        return $this->inMemoryRepository;
+        if ($this->isEventSourced) {
+            return new InMemoryEventSourcedRepository([], $this->aggregateClassNames);
+        } else {
+            return new InMemoryStandardRepository([], $this->aggregateClassNames);
+        }
     }
 
-    public function getRepository(): EventSourcedRepository|StandardRepository
+    public function compile(ContainerMessagingBuilder $builder): Reference|Definition|null
```
dgafka, line 64: What is the case for returning nulls?

packages/Ecotone/src/Messaging/Config/Container/Definition.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+namespace Ecotone\Messaging\Config\Container;
+
+use Ecotone\Messaging\Handler\Type;
+use Ecotone\Messaging\Handler\UnionTypeDescriptor;
+
+use function get_class;
+
+class Definition
```
dgafka, line 10: I think that's a good decision to build our own abstraction around compilation process. As this helps to push as much as possible inside the Ecotone and avoid framework specific problems that need to be later debugged and maintained.

packages/Ecotone/src/Messaging/Config/Container/Definition.php
```diff
@@ -0,0 +1,112 @@
...
+    public function getMethodCalls(): array
+    {
+        return $this->methodCalls;
+    }
+}
+
+/**
+ * @internal
+ */
+class MethodCall
```
dgafka, line 97: Even so it's internal, let's stick in production code to psr-4 way of defining classes.

packages/Ecotone/src/Messaging/Config/Container/Definition.php
```diff
@@ -0,0 +1,112 @@
...
+        }
+        return $this->factory;
+    }
+
+    public function hasFactory(): bool
+    {
+        return !empty($this->factory);
+    }
+
+    public function setFactory(string|array $factory): self
```
dgafka, line 63: Let's define the shape for the array, as it's unclear what to expect. 
If this follows Symfony way (class_name, method), then `array{class-string, string}`

packages/Ecotone/src/Messaging/Config/Container/Definition.php
```diff
@@ -0,0 +1,112 @@
...
+
+    public function __construct(protected string $className, protected array $constructorArguments = [], protected string|array $factory = '')
+    {
+    }
+
+    public static function fromType(Type $type): self
+    {
+        $typeDescriptorArgument = $type instanceof UnionTypeDescriptor ? $type->getUnionTypes() : $type->toString();
+
+        return new self(get_class($type), [$typeDescriptorArgument]);
```
dgafka, line 26:
```suggestion
        return new self($type->toString(), [$typeDescriptorArgument]);
```

dgafka: And this won't be a class, if this is union. Are we missing test case?

packages/Ecotone/src/Messaging/Config/Container/Definition.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->isLazy = $isLazy;
+        return $this;
+    }
+
+    public function islLazy(): bool
+    {
+        return $this->isLazy;
+    }
+
+    public function addMethodCall(string $string, array $array)
```
dgafka, line 80: What is this used for and do we need this?

packages/Ecotone/src/Messaging/Config/Annotation/ModuleConfiguration/ErrorHandlerModule.php
```diff
@@ -67,11 +75,17 @@ public function prepare(Configuration $messagingConfiguration, array $extensionO
                 $messagingConfiguration
                     ->registerDefaultChannelFor(SimpleMessageChannelBuilder::createPublishSubscribeChannel($extensionObject->getDeadLetterQueueChannel()));
             }
+
+            $reference = new Reference(uniqid(HeaderRouter::class . '.'.MessageHeaders::POLLED_CHANNEL_NAME));
```
dgafka, line 79: Are you sure this will be unique? I used to generate Uuid concatenated ids

packages/Ecotone/src/Messaging/Handler/Router/RouterBuilder.php
```diff
@@ -177,6 +176,27 @@ public function build(ChannelResolver $channelResolver, ReferenceSearchService $
         );
     }
 
+    public function compile(ContainerMessagingBuilder $builder): Reference|Definition|null
+    {
+        $interfaceToCall = $this->methodNameOrInterface;
+        if (! $interfaceToCall instanceof InterfaceToCall) {
+            throw new InvalidArgumentException('RouterBuilder can only be compiled with InterfaceToCall');
```
dgafka, line 183: Some constructions of this class do not have InterfaceToCall so they do not need to create InterfaceToCall directly. 
So we either need to pass it here or during construction to resolve it immediately.

packages/Ecotone/src/Messaging/Config/Container/DefinedObject.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Ecotone\Messaging\Config\Container;
+
+interface DefinedObject
```
dgafka, line 5: Why do some abstractions contains `compile` method which returns `Definition` and some `DefinedObject`?

packages/Ecotone/src/Messaging/Config/Container/CompilableBuilder.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Ecotone\Messaging\Config\Container;
+
+interface CompilableBuilder
+{
+    public function compile(ContainerMessagingBuilder $builder): Reference|Definition|null;
```
dgafka, line 7: I wonder if this should be added directly to `MessageHandlerBuilder`. Is this really optional to add?

packages/Ecotone/src/Messaging/Config/Container/CompilableParameterConverterBuilder.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Ecotone\Messaging\Config\Container;
+
+use Ecotone\Messaging\Handler\InterfaceParameter;
+use Ecotone\Messaging\Handler\InterfaceToCall;
+
+interface CompilableParameterConverterBuilder
+{
+    public function compile(ContainerMessagingBuilder $builder, InterfaceToCall $interfaceToCall, InterfaceParameter $interfaceParameter): Reference|Definition|null;
```
dgafka, line 10: I wonder if this should be added directly to `ParameterConverterBuilder`. Is this really optional to add?

packages/Ecotone/src/Messaging/Config/Container/Definition.php
```diff
@@ -0,0 +1,112 @@
...
+        return !empty($this->factory);
+    }
+
+    public function setFactory(string|array $factory): self
+    {
+        $this->factory = $factory;
+        return $this;
+    }
+
+    public function lazy(bool $isLazy = true): self
```
dgafka, line 69: What is the use case for lazy and what does it means in this context?

packages/Symfony/DepedencyInjection/EcotoneExtension.php
```diff
@@ -13,16 +18,66 @@ public function load(array $configs, ContainerBuilder $container)
...
+        }
+
+        /** @TODO Ecotone 2.0 use ServiceContext to configure Symfony */
+        $serviceConfiguration = ServiceConfiguration::createWithDefaults()
+            ->withEnvironment($container->getParameter('kernel.environment'))
+            ->withFailFast(in_array($container->getParameter('kernel.environment'), ['prod', 'production']) ? false : $config['failFast'])
+            ->withLoadCatalog($config['loadSrcNamespaces'] ? 'src' : '')
+            ->withNamespaces($config['namespaces'])
+            ->withSkippedModulePackageNames($skippedModules)
+//            ->withCacheDirectoryPath($container->getParameter('kernel.cache_dir'))
```
dgafka, line 33: Why is this commented?

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
```diff
@@ -1,210 +0,0 @@
...
-            if ($container->has($requiredReference)) {
-                $alias = $container->setAlias(PsrContainerReferenceSearchService::getServiceNameWithSuffix($requiredReference), $requiredReference);
-
-                if ($alias) {
-                    $alias->setPublic(true);
-                }
-            }
-        }
-
-        foreach ($messagingConfiguration->getRegisteredConsoleCommands() as $oneTimeCommandConfiguration) {
```
dgafka, line 187: Where do we register CLI commands now?

packages/Symfony/DepedencyInjection/EcotoneExtension.php
```diff
@@ -13,16 +18,66 @@ public function load(array $configs, ContainerBuilder $container)
...
+            new ServiceCacheConfiguration($serviceConfiguration->getCacheDirectoryPath(), true),
+        );
+
+        $container->register(ServiceCacheConfiguration::class)
+            ->setArguments([
+                $serviceConfiguration->getCacheDirectoryPath(),
+                true,
+            ]);
+
+        $messagingConfiguration->buildInContainer(new SymfonyContainerAdapter($container));
```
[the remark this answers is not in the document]
dgafka: > With this solution, there is also no need to make public "required references".

That's cool, as that was solution to handle Symfony specific implementation :)

packages/Ecotone/src/Lite/LiteContainerImplementation.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Ecotone\Lite;
+
+use Ecotone\Messaging\Config\ConfiguredMessagingSystem;
+use Ecotone\Messaging\Config\Container\ContainerImplementation;
+use Ecotone\Messaging\Config\Container\Definition;
+use Ecotone\Messaging\Config\Container\Reference;
+use Ecotone\Messaging\Config\MessagingSystemContainer;
+use Psr\Container\ContainerInterface;
+
+class LiteContainerImplementation implements ContainerImplementation
```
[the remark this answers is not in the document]
jlabedo: For production, isn't it with LiteApplication ?

Monorepo/ExampleApp/Lite/app.php
```diff
@@ -39,7 +36,7 @@
             ->doNotLoadCatalog()
             ->withCacheDirectoryPath(__DIR__ . "/var/cache")
             ->withFailFast(false)
-            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE]))
+            ->withSkippedModulePackageNames(array_merge(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE]), [ModulePackageList::TEST_PACKAGE]))
```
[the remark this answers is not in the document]
jlabedo: The test package is not in the AllPackages list, so it was not excluded
I thought it was a bug and that test module was registered in production, but the test package is correctly removed by the Symfony bundle, so I thought it was on purpose ?

packages/Ecotone/composer.json
```diff
@@ -36,9 +36,10 @@
     },
     "require": {
         "php": "^8.0",
+        "friendsofphp/proxy-manager-lts": "^1.0.12",
+        "php-di/php-di": "^7.0.1",
```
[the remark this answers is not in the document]
jlabedo: See my comment on the PhpDiContainerImplementation in the wrong package for "historical" :) reason.
So yes, it should go to LiteApplication package.

packages/Dbal/src/Recoverability/DbalDeadLetterModule.php
```diff
@@ -48,6 +51,7 @@ public function prepare(Configuration $messagingConfiguration, array $extensionO
             return;
         }
 
+        $messagingConfiguration->registerServiceDefinition(DbalDeadLetterConsoleCommand::class, new Definition(DbalDeadLetterConsoleCommand::class));
```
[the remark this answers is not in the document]
jlabedo: I agree that this method seems weird on the `MessagingConfiguration` interface...

As we cannot register object instances in the container, all these object instances should provide a `Definition`. Moreover, if they are unique across messaging components (stateful services), we have to register this service in some way.
I tried something else with the `DefinedObject` interface which provides a way to transform an instance to a `Definition`, but it works only for not singleton value object without external service dependency.

I don't see a better solution for now.

packages/Ecotone/src/Lite/LiteContainerImplementation.php
jlabedo: external references are References that could not be resolved by the messaging container builder, so expected to be present in the external/host container.
It is not really used for now, but you could use it to check for missing required reference during compilation phase.

packages/Ecotone/src/Lite/LiteContainerImplementation.php
jlabedo: No. `Reference` is a reference to a messaging component instance OR an external service.
A definition can take 4 types as arguments: 
- a primitive (string, int, bool)
- a reference (it could be a reference to a parameter converter or an external reference)
- a definition (a value object or a dedicated service)
- an array of the preceding types

Let me take an example:
```php
class MethodInvoker {
 public function __construct(object $objectInstance, string $methodName,  array $parameterConverters) {}
}
```

This can be defined as : 
```php
new Definition(MethodInvoker::class, [
  new Reference(OrderService::class), // this is an external service reference
  'placeOrder',
  [
     // We are mixing Reference and Definition here
     new Reference(PayloadConverter::class), // reference to a singleton messaging component service
     new Definition(ValueConverter::class, ["a Value"]), // definition of a dedicated converter
  ]);
```

It will be dumped in the container like this:
```php
[
  PayloadConverter::class => function () { return new PayloadConverter(); },
  "aMethodInvokerReference" => function (Container $c) {
    return new MethodInvoker($c->get(OrderService::class), "placeOrder", [$c->get(PayloadConverter::class), new ValueConverter("a Value")];
  },
]
```

packages/Ecotone/src/Lite/Test/Configuration/InMemoryRepositoryBuilder.php
jlabedo: It should not in the end. But if a component cannot be compiled because one of its sub component is not compilable (because not yet implemented), it returns null.
It was useful for the solution of progressive refactoring, with some components compiled and some not. But as I said in another comment, it does not give any performance boost and it is more complex.
So we can drop the `null` type.

Regarding `Reference` and `Definition`, I am also a bit concerned of the responsibility and the understandability of the `compile` method:
- if you return a Reference, the compile method has built definitions AND registered those into the container, returning the registered reference.
- if you return a Definition, the compile method has built a definition but it has not registered it into the container, letting the responsibility to a parent component to register its definition.

Wdyt ?

packages/Ecotone/src/Messaging/Config/Container/Definition.php
jlabedo: I think it's a `callable` ? I will check, I am not an expert of this php type

packages/Ecotone/src/Messaging/Config/Container/Definition.php
jlabedo: The code is not clear :) 
In case of `UnionTypeDescriptor`, it will return:
```return new Definition(UnionTypeDescriptor::class, [$type->getUnionTypes()])```
In case of `TypDescriptor`:
```return new Definition(TypDescriptor::class, [$type->toString()])```

File: packages/Ecotone/src/Messaging/Config/Container/Definition.php (hunk 11 above)
Comment (jlabedo) [reply packages/Ecotone/src/Messaging/Config/Container/Definition.php:80]: I would prefer not to use it, but it is used for subscribing message handlers to channels.
When a channel is instantiated, it will instantiate and subscribe its message handlers. The dumped code would look like this:
```php
'channel.direct123' => function(Container $c) {
  return (new DirectChannel('direct123'))
         ->subscribe($c->get('a Message handler reference')); // This way, all message handlers are lazy loaded
}
```

I think it would be better to provide subscribed message handlers directly in the constructor.

packages/Ecotone/src/Messaging/Handler/Router/RouterBuilder.php
jlabedo: I put those exceptions a bit everywhere in the code to gradually check which component needed a rewrite, they should not stay in the code.

packages/Ecotone/src/Messaging/Config/Container/DefinedObject.php
jlabedo: See a previous comment, the name sucks.
It is an object instance that can return its own `Definition`. I think this should be dropped

packages/Ecotone/src/Messaging/Config/Container/CompilableBuilder.php
jlabedo: In the first iteration with progressive refactoring yes, but if we agree to go to full refactoring at once, it should be added to `MessageHandlerBuilder`.

packages/Ecotone/src/Messaging/Config/Container/Definition.php
jlabedo: php-di and Symfony propose the option to create a proxy object to lazy load a component when it is actually used at runtime. I used it to lazy load gateways in my first iteration.

But object proxying is a heavy weapon and we can drop that.

packages/Symfony/DepedencyInjection/EcotoneExtension.php
jlabedo: No more caching in Ecotone core :)
But maybe we should keep it for proxy generation in core... or let the framework do that. I don't know.

packages/Symfony/DepedencyInjection/Compiler/EcotoneCompilerPass.php
jlabedo: Forgot that one.

Monorepo/Benchmark/ReflectionAttributeBench.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Monorepo\Benchmark;
+
+
+use Monorepo\Benchmark\Fixtures\ClassWithAnnotation;
+use Monorepo\Benchmark\Fixtures\Endpoint;
+
+class ReflectionAttributeBench
```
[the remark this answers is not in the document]
dgafka: > It was not supposed to be commited, but it opens the discussion on attribute instantiation :).
It compares the instantiation of attributes from:

Ok, so which is quicker? :D 
And in that case, we can drop this benchmark?

> We could use a fallback in those case (runtime instantiation with reflection or var::exporter component)

For now there are no cases like that, so we can consider that out of the scope :)

Monorepo/ExampleApp/Lite/app.php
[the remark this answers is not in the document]
dgafka: Hmm yea true. Something to be done differently in future, as it rises the questions :)

packages/Dbal/src/Recoverability/DbalDeadLetterModule.php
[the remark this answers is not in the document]
dgafka: It will be hard to follow for other contributors. It's best to make the contribution as straight forward as it's possible, so developers can implement specific interface and be sure that it will take care of all the surroundings.
 
I am leaving this open, and will take a look on this. If we will not find the solution right away, then we can do it in some follow up PR, not a blocker.

packages/Ecotone/composer.json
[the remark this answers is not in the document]
dgafka: Can you point the comment? :)

packages/Ecotone/src/Lite/LiteContainerImplementation.php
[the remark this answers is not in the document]
dgafka: Nop, EcotoneLite::bootstrap is for production and integration with any other frameworks or without any framework. 
You can see [the docs](https://docs.ecotone.tech/install-php-service-bus#install-ecotone-lite-no-framework).

So if I understand correctly, if someone would provide Zend Container, this would boot all the Services that related to Ecotone, right?
If that's the case, I wonder if we can lazy load, as this will decrease performance

packages/Ecotone/src/Lite/LiteContainerImplementation.php
dgafka: Oh I see now, thanks for explanation :) 

Do we indeed use singleton for `PayloadConverter`? If so then this is just epic.

packages/Ecotone/src/Lite/LiteContainerImplementation.php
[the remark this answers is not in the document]
jlabedo: Sure we can, I thought it would be better to eager load for testing. I will work for an in memory lazy implementation then.

Monorepo/Benchmark/ReflectionAttributeBench.php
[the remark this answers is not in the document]
jlabedo: We can definitely drop it. The quickest is dumping :)
What do you do with userland attributes for interception ? They could be complex, but I aggree that it should not happen in a lot of codebase.

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/PollingConsumerBuilder.php
```diff
@@ -149,60 +151,58 @@ public function isSupporting(MessageHandlerBuilder $messageHandlerBuilder, Messa
...
+            )
+            ->addAroundInterceptor(
+                AroundInterceptorReference::create(
+                    PollingConsumerErrorInterceptor::class,
+                    $builder->getInterfaceToCall(new InterfaceToCallReference(PollingConsumerErrorInterceptor::class, 'handle')),
+                    Precedence::ERROR_CHANNEL_PRECEDENCE,
+                )
+            )
+            ->compile($builder);
+    }
```
jlabedo, line 184: Here we are registering a single gateway for every consumers. The gateway will use dedicated channel, error interceptor and post send interceptor. Those components will read from the stateful service `PollingConsumerContext` that knows which asynchronous endpoint is being run.

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/PollingConsumerBuilder.php
```diff
@@ -149,60 +151,58 @@ public function isSupporting(MessageHandlerBuilder $messageHandlerBuilder, Messa
...
+    {
+        $executor = new Definition(PollerTaskExecutor::class, [
+            $messageHandlerBuilder->getInputMessageChannelName(),
+            new ChannelReference($messageHandlerBuilder->getInputMessageChannelName()),
+            $this->compilePollingConsumerGateway($builder)
+        ]);
+
+        $messageHandlerReference = $messageHandlerBuilder->compile($builder);
+        $builder->register("polling.{$messageHandlerBuilder->getEndpointId()}.executor", $executor);
+        $builder->register("polling.{$messageHandlerBuilder->getEndpointId()}.channel", new Definition(DirectChannel::class, ['polling-connection-channel', $messageHandlerReference]));
```
jlabedo, line 196: Each consumer will register a reference to their `TaskExecutor` and connection `MessageChannel`

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/PollingConsumerChannel.php
```diff
@@ -0,0 +1,18 @@
...
+
+class PollingConsumerChannel implements MessageChannel
+{
+    public function __construct(private PollingConsumerContext $pollingConsumerContext)
+    {
+    }
+
+    public function send(Message $message): void
+    {
+        $this->pollingConsumerContext->getPollingConsumerConnectionChannel()->send($message);
```
jlabedo, line 16: This singleton channel is redirecting to the "connection channel"

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/PollingConsumerContextProvider.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * @inheritDoc
+     */
+    public function runEndpointWithExecutionPollingMetadata(string $endpointId, ?ExecutionPollingMetadata $executionPollingMetadata): void
+    {
+        $this->currentPollingMetadata = $this->getPollingMetadataForEndpoint($endpointId)->applyExecutionPollingMetadata($executionPollingMetadata);
+
+        try {
+            $this->getPollingConsumer()->run();
```
jlabedo, line 59: When running an asynchronous consumer, the polling context is defined here

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/PollingConsumerContextProvider.php
```diff
@@ -0,0 +1,94 @@
...
+        $endpointId = $this->currentPollingMetadata->getEndpointId();
+        return $this->container->get("polling.".$endpointId.'.channel');
+    }
+
+    public function getPollingConsumerInterceptors(): array
+    {
+        if (! $this->consumerInterceptors) {
+            $this->consumerInterceptors = InterceptedConsumer::createInterceptorsForPollingMetadata($this->currentPollingMetadata, $this->logger);
+        }
+        return $this->consumerInterceptors;
```
jlabedo, line 48: Interceptors are resolved at runtime with execution polling metadata

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/PollingConsumerErrorInterceptor.php
```diff
@@ -0,0 +1,33 @@
...
+use Ecotone\Messaging\Support\ErrorMessage;
+use Throwable;
+
+class PollingConsumerErrorInterceptor
+{
+    public function __construct(private ChannelResolver $channelResolver)
+    {
+    }
+
+    public function handle(MethodInvocation $methodInvocation, Message $requestMessage, ?PollingMetadata $pollingMetadata)
```
jlabedo, line 19: A new parameter converter was introduced to inject currently running `PollingMetadata` retrieved from `PollingConsumerContext`

packages/Ecotone/src/Messaging/Config/MessagingSystemContainer.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public function run(string $endpointId, ?ExecutionPollingMetadata $executionPollingMetadata = null): void
+    {
+        if (!isset($this->pollingEndpoints[$endpointId])) {
+            throw InvalidArgumentException::create("Endpoint with id {$endpointId} was not found");
+        }
+        /** @var EndpointRunner $pollingConsumerContext */
+        $pollingConsumerContext = $this->container->get($this->pollingEndpoints[$endpointId]);
+        $pollingConsumerContext->runEndpointWithExecutionPollingMetadata($endpointId, $executionPollingMetadata);
```
jlabedo, line 84: This is where the `PollingConsumerContext` is run with execution polling metadata

jlabedo (commented):
Hello @dgafka , could you tell me what you think of this way of running polling consumers.
It was introduced mainly to create polling interceptors with execution polling metadata, but I think it has some other benefits (maybe drop the `AsynchronousRunningEndpoint` annotation.

packages/Ecotone/tests/Messaging/Unit/Handler/Chain/ChainMessageHandlerBuilderTest.php
```diff
@@ -490,16 +495,13 @@ public function test_chaining_with_router_at_the_end()
...
             ->chain(TransformerBuilder::createWithExpression('1 + 1'))
             ->chain(TransformerBuilder::createWithExpression('payload + 1'))
             ->withOutputMessageHandler(RouterBuilder::createRecipientListRouter([$outputChannelName]))
-            ->build(
-                InMemoryChannelResolver::createFromAssociativeArray([
-                    $outputChannelName => $outputChannel,
-                ]),
-                InMemoryReferenceSearchService::createWith([ExpressionEvaluationService::REFERENCE => SymfonyExpressionEvaluationAdapter::create()])
-            );
+        );
```
jlabedo, line 504: Simple unit testing from builders is not achievable anymore: definitions are returned and not instances, so a compiling phase is necessary.

To help on "Component testing", I introduced this `ComponentTestBuilder` helper which can build components without a complete messaging system configuration.

What do you think before I work on refactoring all these unit tests :) ? (behat tests are passing now with minor modifications needed)

packages/Ecotone/src/Test/ComponentTestBuilder.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public static function create(): self
+    {
+        $container = InMemoryPSRContainer::createEmpty();
+        $containerBuilder = new ContainerBuilder();
+        $containerBuilder->addCompilerPass(new RegisterSingletonMessagingServices());
+        $containerBuilder->addCompilerPass(new RegisterInterfaceToCallReferences());
+        $containerBuilder->addCompilerPass(new InMemoryContainerImplementation($container));
+        return new self($container, $containerBuilder);
+    }
```
jlabedo, line 39: The component test builder will build a dedicated in memory container for the test case.

jlabedo (commented):
@dgafka , could you also tell me what you think of this way of unit testing components ?

Monorepo/Benchmark/ReflectionAttributeBench.php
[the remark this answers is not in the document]
dgafka: You're right with userland attributes. We do allow to define pointcut pointing to userland attributes, so we will have to support this to avoid B/C.

packages/Dbal/src/Recoverability/DbalDeadLetterModule.php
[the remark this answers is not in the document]
jlabedo: I just realized that this is the equivalent of `ModuleReferenceSearchService::store` method. We could use that one instead.

packages/Ecotone/src/Messaging/Handler/Router/RouterBuilder.php
```diff
@@ -33,17 +40,13 @@ class RouterBuilder implements MessageHandlerBuilderWithParameterConverters
     private ?string $defaultResolution = null;
     private bool $applySequence = false;
     private ?string $endpointId = '';
+    private ?DefinedObject $directObjectToInvoke = null;
```
dgafka, line 43: I can see now what DefinedObject is about. 
What happens right now is:

```php
create defined object (builder) -> create Definition for Defined Object (compilation) -> create defined object from Definition (container)
```

Like you can see, the first step is not really needed, as we this point of time we actually want to have Definition not the constructed defined object. Let's register direct object directly as Definitions and throw away DefinedObject interface.

```php
create Definition for Defined Object (builder) -> create defined object from Definition (container)
```

From what I see it's matter of copy pasting the Definition which you created inside the DefinedObject directly to the Builder (instead of creating )

So as follows, we place this:

```php
    public static function createHeaderMappingRouter(string $headerName, array $headerValueToChannelMapping): self
    {
        $routerBuilder = new self('', 'route');
        $routerBuilder->setObjectToInvoke(HeaderMappingRouter::create($headerName, $headerValueToChannelMapping));

        return $routerBuilder;
    }
```

with this:

```php
    public static function createHeaderMappingRouter(string $headerName, array $headerValueToChannelMapping): self
    {
        $routerBuilder = new self(new Definition(HeaderMappingRouter::class, [
            $this->headerName,
            $this->headerValueToChannelMapping
        ], 'create'), 'route');

        return $routerBuilder;
    }
```

and we can get rid of `private function setObjectToInvoke(DefinedObject $objectToInvoke)`, as it will all be constructed the same way :)

packages/Ecotone/src/Test/ComponentTestBuilder.php
dgafka: Have not dig into that much yet, but I think that's fine. 
In general we should strive for EcotoneLite in most of the cases and avoid low level component testing, as it's easy to break that tests during refactor. Yet in places where we do it it's fine, as I guess it will also tests registering Definitions :)

packages/Ecotone/tests/Messaging/Unit/Handler/Chain/ChainMessageHandlerBuilderTest.php
dgafka: Let's go: https://github.com/ecotoneframework/ecotone-dev/pull/223#discussion_r1359459938

dgafka:
> some other benefits (maybe drop the `AsynchronousRunningEndpoint` annotation.
> 
> **Edit:** I thinks this is not a good idea, let me propose something better before taking time for review

Ok, not reviewing :P 
Just to add, `AsynchronousRunningEndpoint` can be aimed by userland code. This is even proposed in the docs, if you want to intercept Message Consumer.

packages/Ecotone/src/Messaging/Handler/Router/RouterBuilder.php
jlabedo: I agree that there are too many ways of compiling objects.

`DefinedObject` is useful for component testing: instead of recreating a new instance from the definition of a `DefinedObject`, the in memory container implementation will use the existing instance, allowing tests to be constructed like this:
```php
$channel = QueueChannel::create("queueChannel");
$componentUnderTest = ComponentTestBuilder()::create()
    ->withChannel("queueChannel", $channel) // This won't be possible anymore without DefinedObject interface as we can accept only "compilable builders" and not messaging components
    ->build(....);

$channel->send($message); // Without DefinedObject, this channel won't be "connected" in the container as a new instance would have been created (see below)
$channelFromContainer = $componentTest->getMessageChannelByName("queueChannel);
self::assertSame($channelFromContainer, $channel); // This will fail
```

If we drop `DefinedObject`, unit tests would need to be refactored a lot. My strategy is to come to all green tests as fast as possible without having to changeT tests too much. This way I am confident of not breaking something.

dgafka: `QueueChannel` is not external facing API, `SimpleMessageChannelBuilder::createQueueChannel` is. 
In that case we can register `MessageChannelBuilder` inside the `ComponentTestBuilder` and let the Channel Builder register the needed definition. 

```php
$componentUnderTest = ComponentTestBuilder()::create()
    ->withChannel(SimpleMessageChannelBuilder::createQueueChannel("queueChannel")) 
    ->build(....);
```

> If we drop DefinedObject, unit tests would need to be refactored a lot. My strategy is to come to all green tests as fast as possible without having to changeT tests too much. This way I am confident of not breaking something.

We can wait with this refactor when all tests pass. And if that helps discuss how to split the work in non-conflicting way between us two. 

I want avoid extra complexity which `DefinedObject` introduces. As keeping it is like we are partially in old way of defining configuration where we work with real instances and from the other side the other configuration works with Definitions.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Converter/PayloadBuilder.php
```diff
@@ -63,4 +67,14 @@ public function getRequiredReferences(): array
     {
         return [];
     }
+
+    public function compile(ContainerMessagingBuilder $builder, InterfaceToCall $interfaceToCall, InterfaceParameter $interfaceParameter): Definition
```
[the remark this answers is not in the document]
jlabedo: wip

packages/Ecotone/src/Messaging/Channel/PollableChannel/Serialization/OutboundSerializationChannelBuilder.php
```diff
@@ -56,4 +62,15 @@ public function build(ReferenceSearchService $referenceSearchService): ChannelIn
...
         );
     }
+
+    public function compile(ContainerMessagingBuilder $builder): Reference|Definition|null
+    {
+        return new Definition(OutboundSerializationChannelInterceptor::class, [
+            new Definition(OutboundMessageConverter::class, [
+                $this->headerMapper,
+                $this->channelConversionMediaType ?: new Reference('config.defaultSerializationMediaType'),
+            ]),
```
dgafka, line 72: I guess the hard coded string only apply to internal services. We may create some InternalReferenceMap class and put all of them there.

dgafka: In case we reference to actual object like we ConversionService below, then we can put the reference name directly on the object

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/AroundInterceptorReference.php
```diff
@@ -191,6 +220,89 @@ public function buildAroundInterceptor(ReferenceSearchService $referenceSearchSe
         );
     }
 
+    /**
+     * @param AttributeDefinition[] $endpointAnnotations
+     */
+    public function compile(ContainerMessagingBuilder $builder, array $endpointAnnotations, InterfaceToCall $interceptedInterface): Definition|null
```
dgafka, line 226: As right now Reference internally means a reference to DI container, we should change the name of this class.
We can change it to `AroundInterceptorBuilder` to follow up on existing naming

packages/Ecotone/src/Messaging/Handler/Router/RouterBuilder.php
jlabedo: > We can wait with this refactor when all tests pass. And if that helps discuss how to split the work in non-conflicting way between us two.
> 
> I want avoid extra complexity which `DefinedObject` introduces. As keeping it is like we are partially in old way of defining configuration where we work with real instances and from the other side the other configuration works with Definitions.

I agree !

packages/Ecotone/src/Lite/InMemoryContainerImplementation.php
```diff
@@ -0,0 +1,109 @@
...
+     */
+    public function process(ContainerBuilder $builder): void
+    {
+        // We don't need to combine containers as required references are resolved
+        // directly from this compiler pass
+        // $containerInstance = $this->externalContainer ? new CombinedContainer($this->container, $this->externalContainer) : $this->container;
+        $this->container->set(ContainerInterface::class, $this->container);
+        foreach ($builder->getDefinitions() as $id => $definition) {
+            if (! $this->container->has($id)) {
+                $object = $this->resolveArgument($definition, $builder);
```
dgafka, line 33: Taking under consideration that EcotoneLite is used for production, what we could do in here to allow to set `Definition` within the `InMemoryPSRContainer`. 
Then when it's fetched it actually resolve only that definition and replaces it internally for future calls. This way we would avoid resolving everything at boot time. 
Wdyt?

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/AroundInterceptorReference.php
```diff
@@ -191,6 +220,89 @@ public function buildAroundInterceptor(ReferenceSearchService $referenceSearchSe
...
+            }
+            if (! $hasPayloadConverter) {
+                $converterDefinitions[] = PayloadBuilder::create($parameter->getName())->compile($builder, $interceptingInterface, $parameter);
+                $hasPayloadConverter = true;
+                continue;
+            } elseif ($parameter->getTypeDescriptor()->isNonCollectionArray()) {
+                $converterDefinitions[] = AllHeadersBuilder::createWith($parameter->getName())->compile($builder, $interceptingInterface, $parameter);
+                continue;
+            } elseif ($parameter->getTypeDescriptor()->isClassOrInterface()) {
+                $converterDefinitions[] = ReferenceBuilder::create($parameter->getName(), $parameter->getTypeHint())->compile($builder, $interceptingInterface, $parameter);
```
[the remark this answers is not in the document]
jlabedo: I think you missed the `->compile($builder, $interceptingInterface, $parameter)` at the end.
Will add a line return :)

packages/Ecotone/src/Lite/InMemoryContainerImplementation.php
jlabedo: Yes, this would be quite easy to make a lazy container with more or less the same implementation. 
The container `get` method would be something like this:
```php
public function get(string $id) {
  if (! $this->resolvedObjects[$id]) {
    if (! isset($this->definitions[$id])) {
       throw 'Reference not found';
    }
    $this->resolvedObjects[$id] = $this->resolveArgument($this->definitions[$id]);
  }
  return $this->resolvedObjects[$id];
}
```
It would keep a reference to all definitions

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/AroundInterceptorReference.php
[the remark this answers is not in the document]
dgafka: ah true :)

packages/Ecotone/src/Messaging/Config/Container/ContainerBuilder.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * @var CompilerPass[] $compilerPasses
+     */
+    private array $compilerPasses = [];
+
+    public function __construct()
+    {
+    }
+
+    public function register(string|Reference $id, object|array|string $definition = []): Reference
```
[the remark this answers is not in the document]
jlabedo: While cleaning the code for that comment, it makes the register function different from the container interface which is accepting strings.
The real interface is:
```php
public function register(string $id, Definition|Reference|array|string $definition = []): void
```
as you can register:
- Definition
- Reference (alias)
- primitive values (so we should add `bool`, `int`,`float` and `null` to the union type of `$definition`)
- `array` of those values.

So I can drop the `Reference` in the first union type.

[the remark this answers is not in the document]
jlabedo: If I understand correctly, you would like a class for primitive types ? like:
```php
class PrimitiveValue {
  public function __construct(private string|int|bool|array|null $value) {
    // if an array, assert that all values in array are primitives
  }
}
```

And add it to the Union type: `Definition|Reference|PrimitiveValue` ?
That would make sense.

I will check in existing container implementations how they are doing that.

packages/Ecotone/src/Messaging/Handler/ServiceActivator/ServiceActivatorBuilder.php
```diff
@@ -148,42 +169,64 @@ public function getParameterConverters(): array
...
+            $builder,
             $interfaceToCall,
-            $objectToInvoke,
+            $reference,
             $this->methodParameterConverterBuilders,
-            $referenceSearchService,
             $this->getEndpointAnnotations()
         );
+
+        Assert::notNull($methodInvokerDefinition, "Can't compile {$this} because some of parameter converters are not compilable");
```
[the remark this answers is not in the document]
jlabedo: This was only during the "work in progress phase", now all converters are compilable

packages/Ecotone/src/Messaging/Config/Container/DefinedObject.php
jlabedo: I can't find my "previous comment". But the `DefinedObject` is necessary to maintain how some unit tests are written. With this PR, there is a clear separation of compilation and runtime and you cannot share object instances between the two phases, unless with `DefinedObject`. Take this example:
```php
$calculatorService = new CalculatorService();
$serviceActivatorBuilder = ServiceActivatorBuilder::createWithDefinition(new Definition(CalculatorService::class), 'sum');
$serviceActivator = ComponentTestBuilder::create()->build($serviceActivatorBuilder);
$serviceActivator->handle($message);
self::assertTrue($calculatorService->hasBeenCalled()) // This will fail, as an other instance of CalculatorService has been called, the one built in the container
```

In the in memory container implementation, if it finds a `DefinedObject`, it will return the existing instance instead of constructing a new one from its `Definition`. So the test can work:
```php
$calculatorService = new CalculatorService();
$serviceActivatorBuilder = ServiceActivatorBuilder::createWithDefinition($calculatorService, 'sum');
$serviceActivator = ComponentTestBuilder::create()->build($serviceActivatorBuilder);
$serviceActivator->handle($message);
self::assertTrue($calculatorService->hasBeenCalled()) // This will pass
```

packages/Ecotone/src/Lite/Test/Configuration/InMemoryRepositoryBuilder.php
jlabedo: Now it returns only `Definition|Reference`.
I think we should consider returning only `Definition` from this method, deciding to use singleton should be the responsibility of upper layers.

packages/Ecotone/src/Messaging/Config/Container/DefinedObject.php
jlabedo: See also: https://github.com/ecotoneframework/ecotone-dev/pull/223/files#r1359833803

dgafka (approved):
The time has come... :)

PR #225, opened 2023-10-05
Explicitly use bash shell in PR stats workflow

### LINKED ISSUE #224: When benchmarks fails, then workflow should fail
Currently when benchmark fails, then workflow is considered green: https://github.com/ecotoneframework/ecotone-dev/actions/runs/6421495754/job/17435982502?pr=221

marmichalski:
Checking if what I am saying is true, presumably closes #224.

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -100,11 +100,12 @@ jobs:
         run: composer update --${{ matrix.stability }} --prefer-dist --no-interaction
 
       - name: Benchmark PR
+        shell: bash
         env:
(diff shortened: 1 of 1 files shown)


dgafka:
Benchmark was already fixed, so we would need to commit some failure to `PlaceOrderBenchmark` to test it out :)

marmichalski:
This has indeed been confirmed to be a fix.

More can be read here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

dgafka:
:1st_place_medal:

PR #226, opened 2023-10-06
Construct gateways only during the first execution and for given method

dgafka:
In relation to #223, with current configuration we are always building gateways, even if they are not really used. 
Besides that we may actually want to trigger single method from Gateway Interface, so in reallity booting only specific Gateway for executed method make sense. 

This will provide much of extra performance in testing scenario or HTTP Requests scenarios, where all we do is to send an Command or Query using single method on the Command/Query Bus.


dgafka:
Screenshot from 2023-10-06 17-51-23

@jlabedo amazing spot where the most of the time is actually invested. After refactoring that times have dropped really much :)

Issue #229, opened 2023-10-09
Acknowledge interceptor registered multiple times if multiple polling consumers are running


jlabedo:
**Ecotone version(s) affected**: 1.108.1

**Description**  
The acknowledge interceptor is registered multiple times if multiple polling consumers are running

**How to reproduce**  
I did not manage to make a failing test, help would be welcome. If you run two polling consumers in the same messaging system, put a breakpoint on the `AcknowledgeConfirmationInterceptor::ack` method, the second polling consumer will be intercepted two times.

**Possible Solution**  
Fix the polling consumer builder.

jlabedo:
Failing test added in #230

dgafka:
Hey,
Solved with #233

PR #230, opened 2023-10-09
Acknowledge interceptor registered multiple times if multiple polling consumers are running

jlabedo:
Failing test and comments for #229


dgafka:
@jlabedo are you on track with the fix, or should I take look on that?

jlabedo:
Please take it @dgafka , I am discovering polling consumers for now, and it's still a bit hard to me.

dgafka:
Solved with #233

PR #234, opened 2023-10-11
Properly register cached proxies autoloader + allow readonly cache dir.

### LINKED ISSUE #232: cached proxies are still regenerated on fly even if dumped
**Ecotone version(s) affected**: 1.108.1

**Description**  
Continuation of https://github.com/ecotoneframework/ecotone-dev/issues/217

Now `cache:warmup` Symfony command properly generates proxy classes in `<cache-dir>/ecotone`, however they are still generated on fly.

**How to reproduce**  
0. Set up a Symfony project with Ecotone.
1. Make `<cache-dir>/ecotone` readonly
2. Comment or remove `Assert::isTrue(is_writable($cacheDirectoryPath), ...)` in `MessagingSystemConfiguration`. By the way, this check prevents me from having a readonly cache dir, which I would like to have after I dumped all proxies via cache:warmup. Dunno what to do about it, I suggest just removing it altogether - if the cache dir is not writable at the moment of writing a proxy class, it will tell you anyway.
3. Launch `php bin/console list` or `php bin/console ecotone:list`.

You'll get an error Cannot rename `Cannot rename "/tmp/ProxyManagerGeneratedProxy__PM__EcotoneMessagingGatewayConsole...`.

**Possible Solution**  
See https://github.com/Ocramius/ProxyManager/blob/2.15.x/docs/tuning-for-production.md

Basically we have to register the proxy autoloader before attempting to use proxies. Otherwise they are regenerated each time as now.

I added `spl_autoload_register($this->getConfiguration()->getProxyAutoloader())` to `ProxyFactory::createProxyClassWithAdapter()` as a proof of concept, and proxies stopped being regenerated each time. And it worked with the readonly cache dir (to have fully immutable project dir).

I remember about PRs I have agreed to do in the earlier discussions BTW :D

someniatko:
should fix #232

### DIFF
--- a/packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
+++ b/packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
@@ -26,7 +26,6 @@ use Ecotone\Messaging\Endpoint\PollingMetadata;
 use Ecotone\Messaging\Handler\Bridge\Bridge;
 use Ecotone\Messaging\Handler\Chain\ChainMessageHandlerBuilder;
 use Ecotone\Messaging\Handler\Gateway\GatewayProxyBuilder;
-use Ecotone\Messaging\Handler\Gateway\ProxyFactory;
 use Ecotone\Messaging\Handler\InMemoryReferenceSearchService;
 use Ecotone\Messaging\Handler\InterceptedEndpoint;
 use Ecotone\Messaging\Handler\InterfaceToCall;
@@ -50,11 +49,7 @@ use Ecotone\Messaging\MessagingException;
 use Ecotone\Messaging\Support\Assert;
 use Ecotone\Modelling\Config\BusModule;
 use Exception;
-use ProxyManager\Autoloader\AutoloaderInterface;
 use Ramsey\Uuid\Uuid;
-
-use function spl_autoload_register;
-use function spl_autoload_unregister;
 
 /**
  * Class Configuration
@@ -63,8 +58,6 @@ use function spl_autoload_unregister;
  */
 final class MessagingSystemConfiguration implements Configuration
 {
-    private static ?AutoloaderInterface $registered_autoloader = null;
-
     /**
      * @var MessageChannelBuilder[]
      */
@@ -1273,9 +1266,6 @@ final class MessagingSystemConfiguration implements Configuration
         $referenceSearchService = $this->prepareReferenceSearchServiceWithInternalReferences($referenceSearchService, $converters, $interfaceToCallRegistry);
         /** @var ServiceCacheConfiguration $serviceCacheConfiguration */
         $serviceCacheConfiguration = $referenceSearchService->get(ServiceCacheConfiguration::class);
-        /** @var ProxyFactory $proxyFactory */
-        $proxyFactory = ProxyFactory::createWithCache($serviceCacheConfiguration);
-        $this->registerAutoloader($proxyFactory->getConfiguration()->getProxyAutoloader());
 
         $channelInterceptorsByImportance = $this->channelInterceptorBuilders;
         $channelInterceptorsByChannelName = [];
@@ -1375,14 +1365,5 @@ final class MessagingSystemConfiguration implements Configuration
                 $this->afterCallMethodInterceptors
             )
         );
-    }
-
-    private function registerAutoloader(AutoloaderInterface $autoloader)
-    {
-        if (self::$registered_autoloader) {
-            spl_autoload_unregister(self::$registered_autoloader);
-        }
-        spl_autoload_register($autoloader);
-        self::$registered_autoloader = $autoloader;
     }
 }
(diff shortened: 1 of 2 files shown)


packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
```diff
@@ -872,6 +868,10 @@ public static function cleanCache(ServiceCacheConfiguration $serviceCacheConfigu
     private static function deleteFiles(string $target, bool $deleteDirectory): void
     {
         if (is_dir($target)) {
+            Assert::isTrue(
```
dgafka, line 871: This method is recursive, so we can check the permission once at the beginning in `cleanCache` method

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
```diff
@@ -26,6 +29,9 @@ class ProxyFactory
 {
     public const REFERENCE_NAME = 'gatewayProxyConfiguration';
 
+    private static ?AutoloaderInterface $registeredAutoloader = null;
```
dgafka, line 32: Hmm I even think that we could remove the static, as ProxyFactory can be a shared instance from DI. 

However as this is from outside, let's keep static to avoid potential situation, where DI will provide new instance on each fetch, which would lead to memory leaks.

dgafka, line 32: Hmm I even think that we could remove the static, as ProxyFactory can be a shared instance from DI. 

However as this is from outside, let's keep static to avoid potential situation, where DI will provide new instance on each fetch, which would lead to memory leaks.

dgafka: @jlabedo any thoughts on this?

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
```diff
@@ -77,4 +80,26 @@ public function createWithCurrentConfiguration(string $referenceName, ContainerI
...
+        if (self::$registeredAutoloader) {
+            // another ProxyFactory instance may have already registered an autoloader.
+            // this should not happen normally, but just in case we will unload
+            // the old autoloader.
+            spl_autoload_unregister(self::$registeredAutoloader);
+        }
+
+        self::$registeredAutoloader = $this->getConfiguration()->getProxyAutoloader();
+        spl_autoload_register(self::$registeredAutoloader);
+        $this->autoloaderRegistered = true;
```
dgafka, line 103: What was reasoning to introduce `autoloaderRegistered` instead of checking if `registeredAutoloader` is already assigned?

dgafka (commented):
Minor comments :)

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
jlabedo: Agree with avoiding static, but I think it is acceptable before à better solution. 
There is a test with 2 booted messaging systems in the same process (microservices) : static property could lead to strange behavior, but it is working... 
A better solution would be to register the autoloader in the boot méthod provided by frameworks, and unregister in the terminate méthod, as it is done in doctrine for instance. But we would have to handle this in test cases.

packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
someniatko: that's exactly why I've decided to check them everytime, the parent dir might be writable, but a child one - not, for some reason (an external script may have modified permissions).

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
someniatko: to avoid re-registering autoloader for the same ProxyFactory instance.

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
someniatko: > can be a shared instance from DI

unless it's a singleton, I believe we can't be 100% sure about that

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
someniatko: if we check the static property only, each time a ProxyFactory client wants to create a proxy, we will unregister and then register the same autoloader, which is suboptimal in my opinion, which I hope you'll share too :)

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
jlabedo: if we are not 100% sure, if 2 autoloaders are registered it is not à big issue. Before, every test case was registering it's own autoloader without unregistering it, leading to hundreds of those and the memory leak.

dgafka: Yep, if we can't be sure if it's singleton, then in Message Consumer scenario, that would lead to creating a lot of them and memory leak.

I agree that's good enough solution for now :)

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
jlabedo: Ok, I understood your point @someniatko. By doing that you cannot have two messaging systems running in the same process right ?

someniatko: Not sure about what do you mean. If two ProxyFactories are used in the same PHP process, they will unregister each other's autoloader and register their own when necessary.

someniatko: ah, I see, it won't work as intended in all cases. I'll update the relevant code.

dgafka:
Cool stuff, thanks :)

PR #235, opened 2023-10-13
respect cache configuration in SymfonyBundle's MessagingSystemFactory

someniatko:
Cache again 😭


jlabedo (approved):
Seems good :) Thanks @someniatko

PR #236, opened 2023-10-17
Cache fix

m-guminski:
Fix proxy generation when cache catalog is not available


packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
```diff
@@ -842,8 +842,6 @@ public static function prepareWithModuleRetrievingService(
     public static function prepareCacheDirectory(ServiceCacheConfiguration $serviceCacheConfiguration): void
     {
         if (! $serviceCacheConfiguration->shouldUseCache()) {
-            /** We need to clean, in case stale cache exists. So enabling cache will generate fresh one */
-            self::cleanCache($serviceCacheConfiguration);
```
dgafka, line 846: Looks nice

PR #239, opened 2023-10-21
Make framework specific tests more real

dgafka:
- Send requests in Symfony and Laravel
- Fix laravel production config
- Add bench for EcotoneLite

This also copy bench testing into phpunit


dgafka:
Composer is down today, merging this with red build to retest it on #223

PR #250, opened 2023-10-25
Event sourcing benchmark

jlabedo:


Monorepo/Benchmark/FullAppBenchmarkCase.php
```diff
@@ -39,9 +39,6 @@ public function bench_symfony_dev()
         $this->executeForSymfony($container, $kernel);
     }
 
-    /**
-     * @BeforeMethods("dumpLaravelCache")
-     */
     public function bench_laravel_prod(): void
```
dgafka, line 42: Why? Laravel does not dump it's cache on it's own, we have to trigger CLI

PR #255, opened 2023-10-28
Stop reconnecting dbal on context creation

jlabedo:
What is this `reconnect` for ?

I saw that on the event sourcing benchmark, there is two connection established because of this statement, but it is not needed in this case.

I won 19% performance on ES benchmark, as I guess postgres connection are known to be costly

### DIFF
--- a/packages/Dbal/src/DbalReconnectableConnectionFactory.php
+++ b/packages/Dbal/src/DbalReconnectableConnectionFactory.php
@@ -25,8 +25,6 @@ class DbalReconnectableConnectionFactory implements ReconnectableConnectionFacto
 
     public function createContext(): Context
     {
-        $this->reconnect();
-
         return $this->connectionFactory->createContext();
     }
 
@@ -44,8 +42,10 @@ class DbalReconnectableConnectionFactory implements ReconnectableConnectionFacto
         if (! $context) {
             return false;
         }
+        $connection = $context->getDbalConnection();
+        $isConnected = $connection->isConnected() && self::ping($connection);
 
-        return ! $context->getDbalConnection()->isConnected()  || (method_exists(Connection::class, 'ping') && ! $context->getDbalConnection()->ping());
+        return ! $isConnected;
     }
 
     public function reconnect(): void
@@ -107,5 +107,16 @@ class DbalReconnectableConnectionFactory implements ReconnectableConnectionFacto
         $registry = $registry->getValue($connectionFactory);
 
         return [$registry, $connectionName];
+    }
+
+    private static function ping(Connection $connection): bool
+    {
+        try {
+            $connection->executeQuery($connection->getDatabasePlatform()->getDummySelectSQL());
+        } catch (\Exception $e) {
+            return false;
+        }
+
+        return true;
     }
 }


dgafka:
> I saw that on the event sourcing benchmark, there is two connection established because of this statement, but it is not needed in this case.

This is not how production does work, so that can be the fault of different time results. 
Benchmarks should use `CachedConnectionFactory` instead of lower level ConnectionFactories directly. Can you change it and rerun the benchmarks?

The lower level Factories are supposed to create new connection instance each time they are called. This way we can when connection got lost, we are able to recover without failing

packages/Dbal/src/DbalReconnectableConnectionFactory.php
```diff
@@ -25,8 +25,6 @@ public function __construct(DbalConnectionFactory|ManagerRegistryConnectionFacto
 
     public function createContext(): Context
     {
-        $this->reconnect();
```
dgafka, line 28: This is by design, so we can depend that reconnectable will always reconnect

jlabedo:
> This is not how production does work, so that can be the fault of different time results. Benchmarks should use `CachedConnectionFactory` instead of lower level ConnectionFactories directly. Can you change it and rerun the benchmarks?

The benchmark is configured as stated in the [docs](https://docs.ecotone.tech/modules/dbal-support#using-database-connection-string). What can I change in the benchmark to use a `CachedConnectionFactory` ?

`CachedConnectionFactory` is created directly in `DbalTransactionInterceptor`, for each transaction. So each transaction will stop the current connection and reconnect ? Is it necessary ?

I am wondering if it is necessary to have `CachedConnectionFactory` and even `DbalReconnectableConnectionFactory` with recent changes from #223 . You can have a single `Doctrine\DBAL\Connection` which you can reconnect if necessary. And by the way there is a deprectation of `Doctrine\DBAL\Connection::ping()` which states: 'Retry and reconnecting lost connections now happens automatically, ping() will be removed in DBAL 3.'

dgafka:
> The benchmark is configured as stated in the [docs](https://docs.ecotone.tech/modules/dbal-support#using-database-connection-string). What can I change in the benchmark to use a CachedConnectionFactory ?

Right, it's handled internally :)

> CachedConnectionFactory is created directly in DbalTransactionInterceptor, for each transaction. So each transaction will stop the current connection and reconnect ? Is it necessary ?

Well that's for sure a bug and should not be like that. `CachedConnectionFactory` should verify if connection is disconnected and only then go with reconnect
Maybe your change to `isDisconnected` is enough?

>  And by the way there is a deprectation of Doctrine\DBAL\Connection::ping() which states: 'Retry and reconnecting lost connections now happens automatically, ping() will be removed in DBAL 3.'

Doing random reconnect maybe break the transaction boundaries and commit half of the data. That's why it happen on the Transaction Interceptor, as this is the place where transaction starts and it's safe to do reconnect. I hope this option will be optional.

dgafka:
Sup @jlabedo ?

jlabedo:
Hey @dgafka, I have to dig into it, I want to use opentelemetry to check if the problem really exists, but for now I am not seeing anything strange in my production app.

> Maybe your change to isDisconnected is enough?

This is only a refactor.

dgafka:
@jlabedo this will be covered within another PR with small tweak :)

PR #256, opened 2023-10-29
Test Opentracing with different frameworks

dgafka:
This is pre-final step before we can release with public testing

### DIFF
--- a/packages/OpenTelemetry/src/TracingChannelInterceptor.php
+++ b/packages/OpenTelemetry/src/TracingChannelInterceptor.php
@@ -9,6 +9,8 @@ use Ecotone\Messaging\Message;
 use Ecotone\Messaging\MessageChannel;
 use Ecotone\Messaging\Support\MessageBuilder;
 
+use OpenTelemetry\API\Trace\TracerProviderInterface;
+use OpenTelemetry\Context\Context;
 use function json_decode;
 use function json_encode;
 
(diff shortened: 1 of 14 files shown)


packages/OpenTelemetry/tests/Integration/TracingTreeTest.php
```diff
@@ -329,7 +367,71 @@ public function test_tracing_with_asynchronous_handler()
...
+                    SimpleMessageChannelBuilder::createQueueChannel('async_channel'),
+                ])
+        )
+            ->sendCommand(new RegisterUser('1'), ["flowId" => "1"])
+            ->sendCommand(new RegisterUser('2'), ["flowId" => "2"]);
+
+        $ecotoneLite->run('async_channel', ExecutionPollingMetadata::createWithTestingSetup(amountOfMessagesToHandle: 2));
+
+        $collectedTree = self::buildTree($exporter);
+        self::compareTreesByDetails(
```
dgafka, line 393: Traces in here are somehow incorrect. Even so the flows are separate, it combines them into one. 
It looks like it's related to `TracingChannelInterceptor`, as for synchronous scenarios it works correctly. 

Possibility context is not properly closed in OpenTelemetry?

packages/OpenTelemetry/src/TracingChannelInterceptor.php
```diff
@@ -23,21 +29,25 @@ final class TracingChannelInterceptor implements ChannelInterceptor
...
 
         $carrier = [];
         TraceContextPropagator::getInstance()->inject($carrier);
 
         return MessageBuilder::fromMessage($message)
                 ->setHeader(self::TRACING_CARRIER_HEADER, json_encode($carrier))
+                /** Find a better way to propagate that */
+                ->setHeader(MessageHeaders::NON_PROPAGATED_CONTEXT, [
+                    $span, $spanScope
+                ])
```
dgafka, line 50: Ugly, but works and I don't know how I can fetch this span and context from https://github.com/ecotoneframework/ecotone-dev/pull/256/files#diff-c950874072126a16c43246285d2689c39cb5c9c7f9af4b8693f770bd5f2da3e8R63

PR #257, opened 2023-11-02
Improve traces with events

dgafka:
This starts collecting logs as form of an event. Thanks we will be able to find out more about what actually happening within given trace.

For example sending asynchronous message via Event Bus, firstly collect those instead of sending right away:
Screenshot from 2023-11-02 08-29-54

### DIFF
--- a/packages/Ecotone/src/Messaging/Config/Container/ContainerConfig.php
+++ b/packages/Ecotone/src/Messaging/Config/Container/ContainerConfig.php
@@ -11,6 +11,8 @@ use Ecotone\Messaging\Config\Container\Compiler\ValidityCheckPass;
 use Ecotone\Messaging\Config\ServiceCacheConfiguration;
 use Ecotone\Messaging\ConfigurationVariableService;
 use Ecotone\Messaging\Handler\Gateway\ProxyFactory;
+use Ecotone\Messaging\Handler\Logger\LoggingGateway;
+use Ecotone\Messaging\Handler\Logger\StubLoggingGateway;
 use Ecotone\Messaging\InMemoryConfigurationVariableService;
 use Psr\Container\ContainerInterface;
 
@@ -30,6 +32,9 @@ class ContainerConfig
         $container = new LazyInMemoryContainer($containerBuilder->getDefinitions(), $externalContainer);
         $container->set(ConfigurationVariableService::REFERENCE_NAME, $configurationVariableService ?? InMemoryConfigurationVariableService::createEmpty());
         $container->set(ProxyFactory::class, $proxyFactory ?? new ProxyFactory(ServiceCacheConfiguration::noCache()));
+        if (!$container->has(LoggingGateway::class)) {
+            $container->set(LoggingGateway::class, StubLoggingGateway::create());
+        }
         return $container->get(ConfiguredMessagingSystem::class);
     }
 
--- a/packages/Ecotone/src/Messaging/Handler/Logger/LoggingGateway.php
+++ b/packages/Ecotone/src/Messaging/Handler/Logger/LoggingGateway.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ecotone\Messaging\Handler\Logger;
+
+use Ecotone\Messaging\Attribute\MessageGateway;
+use Ecotone\Messaging\Attribute\Parameter\Header;
+use Ecotone\Messaging\Attribute\Parameter\Payload;
+use Ecotone\Messaging\Message;
+
+interface LoggingGateway
+{
+    #[MessageGateway(LoggingService::INFO_LOGGING_CHANNEL)]
+    public function info(
+        #[Payload] string $text,
+        #[Header(LoggingService::CONTEXT_MESSAGE_HEADER)] Message $message,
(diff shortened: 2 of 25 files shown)


packages/Ecotone/src/Messaging/Handler/Logger/LoggingGateway.php
```diff
@@ -0,0 +1,27 @@
...
+        #[Header(LoggingService::CONTEXT_MESSAGE_HEADER)] Message $message,
+        #[Header(LoggingService::CONTEXT_EXCEPTION_HEADER)] ?\Exception $exception = null
+    ): void;
+
+    #[MessageGateway(LoggingService::ERROR_LOGGING_CHANNEL)]
+    public function error(
+        #[Payload] string $text,
+        #[Header(LoggingService::CONTEXT_MESSAGE_HEADER)] Message $message,
+        #[Header(LoggingService::CONTEXT_EXCEPTION_HEADER)] ?\Exception $exception = null
+    ): void;
```
dgafka, line 26: Whatever logs will be sent via `LoggingGateway` when tracing is enabled, it will be added to the current active span.

packages/Ecotone/src/Messaging/Handler/Logger/LoggingService.php
```diff
@@ -35,6 +43,44 @@ public function __construct(ConversionService $conversionService, LoggerInterfac
         $this->logger = $logger;
     }
 
+    #[ServiceActivator(self::INFO_LOGGING_CHANNEL)]
+    public function info(
```
dgafka, line 47: @jlabedo after your latest changes to DI, we can know define configuration within Ecotone using Attributes just as it would be done on userland code. This will simplify modifications and maintaining this code, as we don't need to manually wire things together.

- [To do this it's enough to register Service Definition](https://github.com/ecotoneframework/ecotone-dev/pull/257/files#diff-40f2c0a1e9fffa9ad3ea21255f6d7316590fd1908471efc510cba9385697b398R176)
- [And the class to be loaded by Ecotone](https://github.com/ecotoneframework/ecotone-dev/pull/257/files#diff-b83c6b9b8da73699a1c84272b4dd7074fb689fd0d0cd524720241556c1e84a04R85)

For internal Gateways only the second step is needed, as it works on interfaces.

cc: @unixslayer

packages/Ecotone/src/Messaging/Config/Container/ContainerConfig.php
```diff
@@ -30,6 +32,9 @@ public static function buildMessagingSystemInMemoryContainer(
         $container = new LazyInMemoryContainer($containerBuilder->getDefinitions(), $externalContainer);
         $container->set(ConfigurationVariableService::REFERENCE_NAME, $configurationVariableService ?? InMemoryConfigurationVariableService::createEmpty());
         $container->set(ProxyFactory::class, $proxyFactory ?? new ProxyFactory(ServiceCacheConfiguration::noCache()));
+        if (!$container->has(LoggingGateway::class)) {
+            $container->set(LoggingGateway::class, StubLoggingGateway::create());
+        }
```
jlabedo, line 37: I think this is not the good place to register this service.
We should discuss that, but I think we have to differentiate 2 types of services in Ecotone: messaging services and framework provided services. When you test with Ecotone Lite against a symfony container (or other), Ecotone Lite will replace messaging services but will use framework services from the container if they exists. Moreover, MessagingConfiguration will register messaging services, but it is the responsability of the framework (symfony/laravel/lite/liteapp) to register framework services. For now, the only framework services are:
- ConfigurationVariableService
- ServiceCacheConfiguration
- Logger

So if you need to register a messaging service like `LoggingGateway`, `RegisterSingletonMessagingServices::process` would be a better place

packages/Ecotone/src/Messaging/Handler/Logger/LoggingService.php
jlabedo: For the first needed step, we could have a discussion around autowiring so it won't be necessary. But I still have some doubts on adding too much magic.

Anyway we will have to rethink how modules are built:
- no need to have different create and prepare phase
- no need for `ModuleReferenceSearchService`

So maybe there is no need of Module Class, and it is just an Attribute to be able to filter component to be registered ?

dgafka: > For the first needed step, we could have a discussion around autowiring so it won't be necessary. But I still have some doubts on adding too much magic.

Hmmm, yea possibly. 

> no need for ModuleReferenceSearchService

That possibility can be removed without much effort

> no need to have different create and prepare phase

We do have prepare and compile phases now. One if for collecting and second is for dumping. I guess I would need more info what would like to achieve here.

Let's open github discussion for next phase of refactors and keep the conversion there. 
I will want to explore this more after opentelemetry will be ready. As my plan is to include tracing as part of the workshop I will be doing at `15.11` on PHPCon, so not so much time left :)

packages/Ecotone/src/Messaging/Config/Container/ContainerConfig.php
dgafka: I will take a look on that within next PR

PR #269, opened 2023-11-14
Fix no orm manager registry

m-guminski:


dgafka:
@m-guminski tests are red Mate :)

dgafka:
Fixed with #271

PR #295, opened 2024-02-21
Add support for Laravel 11

martio:
Laravel 11 requires a minimum PHP version of 8.2. Tests on 8.0 will not pass.

### DIFF
--- a/packages/Laravel/composer.json
+++ b/packages/Laravel/composer.json
@@ -35,7 +35,7 @@
     },
     "require": {
         "ecotone/ecotone": "~1.212.0",
-        "laravel/framework": "^9.5.2|^10.0"
+        "laravel/framework": "^9.5.2|^10.0|^11.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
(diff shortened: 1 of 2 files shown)


packages/Laravel/composer.json
```diff
@@ -35,7 +35,7 @@
     },
     "require": {
         "ecotone/ecotone": "~1.212.0",
-        "laravel/framework": "^9.5.2|^10.0"
+        "laravel/framework": "^9.5.2|^10.0|^11.0"
```
dgafka, line 38: This also need to be added in root's composer.json for monorepo support

martio: @dgafka I've added it. Tomorrow I will continue testing the entire Ecotone framework on Laravel 11. If I encounter any issues, I'll make corrections. Ecotone is a very useful framework. Great job, congratulations! :clap:

dgafka:
Thank you for contribution ;)

packages/Laravel/composer.json
dgafka: Let's see how it goes, and thank you :)

PR #305, opened 2024-03-23
Add support of env variables in Ecotone configuration for Symfony

lifinsky:
We can use env variables in Symfony bundle config. Example:
```
ecotone:
  serviceName: "%env(SERVICE_NAME)%"
  defaultErrorChannel: "dbal_dead_letter"

parameters:
  env(SERVICE_NAME): backoffice_service
```

### DIFF
--- a/packages/Symfony/DependencyInjection/EcotoneExtension.php
+++ b/packages/Symfony/DependencyInjection/EcotoneExtension.php
@@ -22,6 +22,7 @@ class EcotoneExtension extends Extension
     {
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
+        $config = $container->resolveEnvPlaceholders($config, true);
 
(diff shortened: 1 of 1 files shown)


dgafka:
Cool :) 

Are you able to add some test for this?

lifinsky:
Yes, I will do. Maybe on Thursday.

dgafka:
Sup @lifinsky :)

lifinsky:
Sorry, I remember.

dgafka:
Why closed @lifinsky ?

lifinsky:
It is not possible for int and bool config parameters for example,  because in extension we can not use env variables processors, this is not possible before compile phase https://symfony.com/doc/current/configuration/env_var_processors.html. I have tried in different ways...

lifinsky:
I have tried custom casting before apply to configuration, but also got error on container compile phase, it is not possible remove bundle config parameters after load to bundle configuration, or this needs additional research.

PR #308, opened 2024-03-29
Handle same Dead Message for different Asynchronous Endpoints

dgafka:
There may be a case when two Event Handlers for same Event Message will fail and store Message in Dead Letter.
In this situation second will be blocked from being stored as Message Id will be the same. 

For this situation new Message Id for Dead Letter will be generated, yet Message will preserve it's origin Message In in Message Headers. Therefore replaying dead message will replay it with original Message Id.


dgafka:
cc: @lifinsky

dgafka:
First idea was to change Primary Key, yet this will result in hard to make Message Replays and Deletions, as it would require us to us pass much more data in order to do this action. 
That would create unpleasant experience to work with Dead Letter. 

Therefore the solution provided within this PR, handles the edge case yet gracefully.

Issue #309, opened 2024-04-02
Converter services are not applicable for headers


lifinsky:
**Ecotone version(s) affected**: 1.216.1

**Description**  
For example for test purpose I create value object TestUuid
```
final class TestUuid
{
    private function __construct(public readonly string $data) {
    }

    public static function fromString(string $data): self
    {
        return new self($data);
    }

    public function __toString(): string
    {
        return $this->data;
    }
}
```

And converter service:
```
class UuidConverter
{
    #[Converter]
    public function convert(string $source): TestUuid
    {
        return TestUuid::fromString($source);
    }
}
```

And send command:
```
$this->commandBus->sendWithRouting(
    'create.user',
    new CreateUser($name),
    metadata: ['userId' => '58009c9e-822d-4adb-906f-14c4bf9117e4']
);
```

And get an exception:
`Can't convert from application/json:string to application/x-php:App\Infrastructure\TestUuid Could not decode JSON, syntax error - malformed JSON.`

Converter class is not used because HeaderConverter thinks the headers are json media type.

Command handler:
```
#[CommandHandler('create.user')]
public static function register(CreateUser $command, #[Header('userId')] TestUuid $uuid): static
{
    return new static((string) $uuid, $command->user);
}
```

But if I create a media type converter, then everything works:
```
#[MediaTypeConverter]
class UuidConverter implements Converter
{
    public function convert($source, TypeDescriptor $sourceType, MediaType $sourceMediaType, TypeDescriptor $targetType, MediaType $targetMediaType): Test
    {
        return TestUuid::fromString($source);
    }

    /**
     * @inheritDoc
     */
    public function matches(TypeDescriptor $sourceType, MediaType $sourceMediaType, TypeDescriptor $targetType, MediaType $targetMediaType): bool
    {
        return $sourceType->isString() && $targetType->isClassOfType(TestUuid::class);
    }
}

```
Converter service works only for payload.

**How to reproduce**  
Use any value objects in headers except Ramsey\Uuid\UuidInterface (because UuidInterface in headers is supported by StringToUuidConverter)

**Possible Solution**  
HeaderConverter::getArgumentFrom method has a problem:
```
        if (! $sourceValueType->isCompatibleWith($targetType)) {
            if ($sourceValueType->isScalar() && $this->canConvertTo($headerValue, DefaultHeaderMapper::DEFAULT_HEADER_CONVERSION_MEDIA_TYPE, $targetType)) {
[...]

PR #310, opened 2024-04-05
Default php conversion for headers

dgafka:
This changes the default serialization for Headers to use PHP to PHP with Converters, only if this is not possible conversion will fallback to JSON conversion.

This refers to #309


dgafka:
Updated Docs:

https://docs.ecotone.tech/messaging/conversion/conversion/payload-conversion
https://docs.ecotone.tech/messaging/conversion/conversion/headers-conversion

PR #331, opened 2024-07-04
use collection name when deleting document

unixslayer:


packages/Dbal/src/DocumentStore/DbalDocumentStore.php
```diff
@@ -108,6 +108,7 @@ public function deleteDocument(string $collectionName, string $documentId): void
         $this->getConnection()->delete(
             $this->getTableName(),
             [
+                'collection' => $collectionName,
```
dgafka, line 111: Can you add test, with adding two documents with same id under different collections and deleting one? :)

dgafka (approved):

dgafka (approved):

Issue #325, opened 2024-06-20
Synchronous event handlers order or priority


lifinsky:
**Description**  
Currently, event handler priority works only for asynchronous event handlers. However, to ensure that, for example, one of the projections is processed first, it is necessary to have the ability to prioritize the event handler.

lifinsky:
There is an event-sourcing transaction aggregate and an event-sourcing customer balance aggregate. To create a projection of fund movements on the balance, we need up-to-date transaction data. Ideally, to avoid using event handler prioritization, it would be great to guarantee that the synchronous projection is executed immediately after the event is saved to its aggregate in the event store, and only after that, the event is listened to by another aggregate.

dgafka:
Hey @lifinsky,

Yep, priority does work only for async, but I don't see reason why wouldn't it work for sync scenarios too. 
I will see what we can do with it when I will have some free time :)

dgafka:
This feature is now added https://github.com/ecotoneframework/ecotone-dev/releases/tag/1.223.0

PR #333, opened 2024-07-06
Priotization of Synchronous Message Handlers

dgafka:
This allows to priortize synchronous Message Handlers, as before it was only possible to do it for asynchronous ones. 
It allows provides default pritization when handlers have the same priority number. In that case we do the order like:
- Projection Handlers
- Aggregate Handlers
- Any other Handlers

This solves #325


lifinsky:
@dgafka ❤️❤️❤️

PR #337, opened 2024-07-17
fix attribute resolver on promoted properties

### LINKED ISSUE #336: Attribute used in an CPP property is disallowed in attribute resolving callstack of command handlers
**Ecotone version(s) affected**: 1.224.0

**Description**  

The following issue occurs on warmup, when using an Attribute which should be allowed in this place in CPP

`Uncaught Error: Attribute "Symfony\Component\DependencyInjection\Attribute\AutowireLocator" cannot target property (allowed targets: parameter)`

This attribute(Symfony\Component\DependencyInjection\Attribute\AutowireLocator) for example is allowed by symfony to be used for properties in a CPP definition

```php
 public function __construct(
        #[AutowireLocator(FooLocatedService::class)]
        private ContainerInterface $foos,
    ) {
    }
```

So i think this should be allowed by ecotone too

**How to reproduce**  
I set an example symfony application with ecotone integration where this can reproduced at cache warmup
https://github.com/tandev/ecotone-attribute-resolving/tree/main

jlabedo:
## Description

Allow attributes targeting parameters to be used on php>8 promoted properties

Fixes #336 

## Type of change

### DIFF
--- a/packages/Ecotone/src/AnnotationFinder/AnnotationResolver/AttributeResolver.php
+++ b/packages/Ecotone/src/AnnotationFinder/AnnotationResolver/AttributeResolver.php
@@ -6,6 +6,7 @@ use Ecotone\AnnotationFinder\AnnotationResolver;
 use Ecotone\AnnotationFinder\ConfigurationException;
 use Ecotone\AnnotationFinder\TypeResolver;
 use Ecotone\Messaging\Attribute\IsAbstract;
+use Error;
 use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionException;
@@ -81,7 +82,11 @@ class AttributeResolver implements AnnotationResolver
                         return $carry;
                     }
 
-                    $carry[] = $attribute->newInstance();
+                    try {
+                        $carry[] = $attribute->newInstance();
+                    } catch (Error) {
+                        // Do nothing: it is an attribute targeting a parameter promoted to a property
+                    }
 
                     return $carry;
                 }, []);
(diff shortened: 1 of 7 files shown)


packages/Ecotone/src/AnnotationFinder/AnnotationResolver/AttributeResolver.php
```diff
@@ -81,7 +82,11 @@ public function getAnnotationsForProperty(string $className, string $propertyNam
                         return $carry;
                     }
 
-                    $carry[] = $attribute->newInstance();
+                    try {
+                        $carry[] = $attribute->newInstance();
+                    } catch (Error) {
+                        // Do nothing: it is an attribute targeting a parameter promoted to a property
```
dgafka, line 88: That's assumption that this is the error :). We may be swallowing others errors too.
If we can't check by exception type, can we do an message type check e.g. string contains `cannot target property`?

jlabedo: Sure

packages/Ecotone/src/AnnotationFinder/AnnotationResolver/AttributeResolver.php
```diff
@@ -84,8 +84,12 @@ public function getAnnotationsForProperty(string $className, string $propertyNam
 
                     try {
                         $carry[] = $attribute->newInstance();
-                    } catch (Error) {
-                        // Do nothing: it is an attribute targeting a parameter promoted to a property
+                    } catch (Error $e) {
+                        if (\preg_match("/Attribute \"(.*)\" cannot target property/", $e->getMessage())) {
+                            // Do nothing: it is an attribute targeting a parameter promoted to a property
+                        } else {
```
dgafka, line 90: Can we also apply this to `TypeResolver#86`? It's other way around there and we ignore all exceptions.

jlabedo: Ha, you found it ! I knew it was done elsewhere.

dgafka (approved):

PR #339, opened 2024-07-21
Improve messaging configuration performance

jlabedo:
## Description

Improve performance of Ecotone configuration phase.
For a medium sized repository using symfony, my cache:clear command went from **22 seconds** (92% of which is ecotone configuration processing)  to **8 seconds**.
This PR delays the endpoint uniqueness check at the compilation phase, so now the check is done only once
Capture d’écran 2024-07-21 à 13 40 35

## Type of change

- Performance improvement


dgafka:
wow, nice one :+1:

dgafka (approved):

PR #340, opened 2024-07-21
remove laminas/laminas-zendframework-bridge dependency

jlabedo:
## Description

Remove unneeded and deprecated dependency `laminas/laminas-zendframework-bridge`

## Type of change

### DIFF
--- a/packages/Symfony/composer.json
+++ b/packages/Symfony/composer.json
@@ -28,7 +28,6 @@
     "description": "Extends Ecotone with Symfony integration",
     "require": {
         "ecotone/ecotone": "~1.225.0",
-        "laminas/laminas-zendframework-bridge": "^1.0.0",
         "symfony/console": "^5.4|^6.0|^7.0",
         "symfony/dependency-injection": ">=v5.4.19 <6.0.0 || >=v6.0.19 <6.1.0 || >=v6.1.11 <6.2.0 || >=v6.2.5 <7.0.0 || >=v7.0.0 <8.0.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -159,6 +158,7 @@
         "moneyphp/money": "^4.1.0"
     },
     "conflict": {
+        "laminas/laminas-code": "<4.5.0",
```
dgafka, line 161: How does it relate?

jlabedo: It does not :) I thought it was related because `friendsofphp/proxy-manager-lts` requires `laminas/laminas-code` and `laminas/laminas-code:<4.5.0` requires `laminas/laminas-zendframework-bridge dependency` but not after 4.5.0

jlabedo:
The problem comes from a conflict with `symfony/proxy-manager-bridge` which is not required anymore after symfony 6.2.

So droping this dependency would also drop compatibility with symfony <6.2

dgafka:
> So droping this dependency would also drop compatibility with symfony <6.2

We could drop this library and add it suggested. So if someone would use lower version, he can always pull it. 
wdyt?

jlabedo:
I am proposing #343 , I did it a long time ago while strugling with proxy generation ([here](https://github.com/ecotoneframework/ecotone-dev/pull/148/files#diff-d4ee1a820c5fc2fb9870c4d397cc6d9cf5676b671c1e047489fd6f8700d95d9a))

If you don't like it, we can make it suggested in composer as you suggest (haha), but I think it will break CI with `prefer-lowest`

dgafka (approved):

PR #342, opened 2024-07-21
Using headers in event sourcing handler

dgafka:
## Description

This add Enterprise features which does allow for using Messages Headers inside Event Sourcing Handlers

image

image

## Type of change

- This change requires a documentation update


packages/Ecotone/src/Messaging/Attribute/Enterprise.php
```diff
@@ -0,0 +1,17 @@
...
+namespace Ecotone\Messaging\Attribute;
+
+use Attribute;
+use Ecotone\Messaging\Support\Assert;
+
+/**
+ * licence Enterprise
+ */
+#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_CLASS)]
+class Enterprise
```
dgafka, line 14: This allows for conditional enabling of Enterprise feature. Can be used for switching Projection mode.

packages/Ecotone/src/Messaging/Config/EnterpriseLicenceDecider.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static function create(self $enterpriseModeDecider, object $openCoreService, object $enterpriseService): object
+    {
+        if ($enterpriseModeDecider->hasEnterpriseLicence()) {
+            return $enterpriseService;
+        }
+
+        return $openCoreService;
+    }
```
jlabedo, line 67: It should work like that :

```suggestion
    public static function prepareDefinition(string $className, Reference|Definition $openCoreService, Reference|Definition $enterpriseService): Definition
    {
        return new Definition(
            $className,
            [
                $openCoreService,
                $enterpriseService
            ],
            factory: [self::class, "decide"]
        );
    }

    public function decide(object $openCoreService, object $enterpriseService): object
    {
        if ($this->hasEnterpriseLicence()) {
            return $enterpriseService;
        }

        return $openCoreService;
    }
```

jlabedo (approved):
Nice work!
I may suggest that we choose between enterprise and core components at compilation time with a configuration flag. Like that we don't need to register both components in the container, and above all we can fail fast in production if we misconfigured the licence, and fail fast for core user at compilation if they use enterprise features. If you have some code using enterprise without the licence but it appears 2 hours after deploying to production because of a low usage feature, it can be very deceptive :)

dgafka:
> Like that we don't need to register both components in the container, and above all we can fail fast in production if we misconfigured the licence

Make sense, to not register both :) 

I will add licence keys, which will be validated at compilation time. But that I will handle this separate PR a bit later.

packages/Ecotone/src/Messaging/Config/EnterpriseLicenceDecider.php
dgafka: Can you explain the difference? I am having hard time spotting that. 

> factory: [self::class, "decide"]

Shouldn't the factory method be static?

jlabedo: The factory can be an object instance, it depends if the method is static or not. It is similar to symfony : https://symfony.com/doc/current/service_container/factories.html#non-static-factories

Maybe I am wrong with my suggestion and you should pass a reference (not just the class name) as first array key:
```php
        return new Definition(
            $className,
            [
                $openCoreService,
                $enterpriseService
            ],
            factory: [Reference::to(self::class), "decide"]
        );
```

dgafka:
@jlabedo @unixslayer 
This PR actually shows how to register conditional Enterprise features. 
With next PR, I will cover functionality where full Module is either registered based on licence or skipped.
After that we can sync up and plan next steps.

packages/Ecotone/src/Messaging/Config/EnterpriseLicenceDecider.php
dgafka: I will cover conditional registering of Services with next Enterprise related PR

dgafka: So [in here](https://github.com/ecotoneframework/ecotone-dev/pull/342/files#diff-52504a6b952169e62f6e218430717248d53a8ed55d1cff3786d4e790438457caR330), being able to differentiate in what mode we currently are.

dgafka:
[Using metadata docs](https://docs.ecotone.tech/modelling/event-sourcing/event-sourcing-introduction/working-with-metadata#enterprise-accessing-metadata-in-event-sourcing-handler)

[Event versioning metadata](https://docs.ecotone.tech/modelling/event-sourcing/event-sourcing-introduction/event-versioning#enterprise-accessing-metadata-in-event-sourcing-handler)

PR #343, opened 2024-07-21
Proxy generation for ecotone

jlabedo:
## Description

Custom gateway proxy generation with `laminas/laminas-code` instead of `friendsofphp/proxy-manager-lts`.
Even with some low level string manipulation, this abstraction seems less complex (to me) than the `RemoteAdadpter` from `friendsofphp/proxy-manager-lts`.
Also, no more custom Autoloader is required, making it less error prone.

A small performance boost is observed in benchmarks (I don't know why, maybe because of Autoloader)

Drop `friendsofphp/proxy-manager-lts`

## Type of change

- Performance improvement


packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
```diff
@@ -32,76 +21,86 @@
...
-        $configuration = new Configuration();
-
-        if ($this->serviceCacheConfiguration->shouldUseCache()) {
-            MessagingSystemConfiguration::prepareCacheDirectory($this->serviceCacheConfiguration);
-            $configuration->setProxiesTargetDir($this->serviceCacheConfiguration->getPath());
-            $fileLocator = new FileLocator($configuration->getProxiesTargetDir());
-            $configuration->setGeneratorStrategy(new FileWriterGeneratorStrategy($fileLocator));
-            $configuration->setClassSignatureGenerator(new ClassSignatureGenerator(new SignatureGenerator()));
+        $proxyClassName = self::getFullClassNameFor($proxyReference->getInterfaceName());
+        if (! class_exists($proxyClassName, false)) {
```
dgafka, line 47: does this actually check, if file was already required (loaded)?

If so, can we extract it into private method e.g. `isAutoloaded`

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
```diff
@@ -32,76 +21,86 @@
...
+            if ($this->serviceCacheConfiguration->shouldUseCache()) {
+                $file = $this->generateCachedProxyFileFor($proxyReference, false);
+                require $file;
+            } else {
+                $code = $this->proxyGenerator->generateProxyFor($proxyClassName, $proxyReference->getInterfaceName());
+                eval(\substr($code, 5));
```
dgafka, line 53: why do we start from 5th character?

dgafka:
Separate autoloader stuff have proven multiple times to be problematic, if there is possibility to drop it, let's go for it :)

I will give one more look later this day :)

dgafka:
At some point we may also introduce support for [inbuilt Lazy Objects](https://wiki.php.net/rfc/lazy-objects), if it will be merged into core :)

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
jlabedo: I clarified it by adding `<?php\n` for file dumping instead of removing it

packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
jlabedo: Yes, I extracted functions for clarity

dgafka (dismissed):
There is a test failing

jlabedo:
> There is a test failing

Because I broke CI on main 😉

dgafka:
Do we still need to [install this](https://github.com/ecotoneframework/ecotone-dev/pull/345), when this will be merged?

jlabedo:
> Do we still need to [install this](https://github.com/ecotoneframework/ecotone-dev/pull/345), when this will be merged?

Yes we have to keep it. I will check again how to remove that deps.

dgafka (approved):

PR #346, opened 2024-07-22
Pointcut compilation

jlabedo:
## Description

Compile pointcuts during compilation phase.
Pointcuts are tested on a lot of different classes during compilation phase, so this PR may speed compilation phase of large projects.
For a medium sized project with Symfony, I obtained merely 10% performance boost on cache:clear.

(also made some cleaning, droping the dependency on `InterfaceToCallRegistry` for pointcuts. This could have a bad impact on performance as no reflection cache is used, but it required adding the registry to each MethodInterceptor::create calls)

## Type of change

- Performance

### DIFF
--- a/packages/Ecotone/src/Messaging/Config/Annotation/ModuleConfiguration/BasicMessagingModule.php
+++ b/packages/Ecotone/src/Messaging/Config/Annotation/ModuleConfiguration/BasicMessagingModule.php
@@ -88,10 +88,6 @@ class BasicMessagingModule extends NoExternalConfigurationModule implements Anno
         $messagingConfiguration->registerConverter(new DeserializingConverterBuilder());
 
         $messagingConfiguration
-            ->registerInternalGateway(TypeDescriptor::create(InboundGatewayEntrypoint::class))
-            ->registerInternalGateway(TypeDescriptor::create(EnrichGateway::class));
-
-        $messagingConfiguration
             ->registerMessageHandler(
                 RouterBuilder::create(
                     new Definition(HeaderRouter::class, [MessagingEntrypoint::ENTRYPOINT]),
--- a/packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Pointcut.php
+++ b/packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Pointcut.php
@@ -9,7 +9,6 @@ use Ecotone\Messaging\Handler\InterfaceToCall;
 use Ecotone\Messaging\Handler\InterfaceToCallRegistry;
 use Ecotone\Messaging\Handler\TypeDescriptor;
 use Ecotone\Messaging\Handler\UnionTypeDescriptor;
-use Ecotone\Messaging\Support\Assert;
 use Ecotone\Messaging\Support\InvalidArgumentException;
 
 /**
@@ -22,17 +21,16 @@ use Ecotone\Messaging\Support\InvalidArgumentException;
  */
 class Pointcut
 {
-    private ?string $expression;
+    private ?PointcutExpression $parsedExpression;
 
     private function __construct(?string $expression)
     {
-        $results = $this->getExpressionsDivinedByBrackets($expression);
-
-        if ($results) {
-            Assert::isTrue(count($results) === count($this->getInBetweenBracketsExpressions($expression)) + 1, "Expression {$expression} is missing `||` or `&&` between brackets or `{}&&{}` bracket is missing on one of the expressions");
+        if (\is_null($expression) || $expression === '') {
+            $this->parsedExpression = null;
+        } else {
+            $parser = new PointcutParser($expression);
+            $this->parsedExpression = $parser->parse();
         }
-
-        $this->expression = $expression;
     }
 
     /**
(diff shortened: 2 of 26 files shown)


packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
```diff
@@ -0,0 +1,169 @@
...
+            if (strpos($token, '::') !== false) {
+                [$class, $method] = explode('::', $token);
+
+                return new PointcutMethodExpression($class, $method);
+            }
+            if (strpos($token, '*') !== false) {
+                return new PointcutRegexExpression($token);
+            }
+            if (TypeDescriptor::isItTypeOfExistingClassOrInterface($token)) {
+                $classDefinition = ClassDefinition::createFor(TypeDescriptor::create($token));
```
jlabedo, line 112: In the actual implementation, the interface registry is used here.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
```diff
@@ -0,0 +1,168 @@
...
+        $filteredParts = [];
+        for ($i = 0; $i < count($parts); $i++) {
+            if ($parts[$i] === self::OPEN_PAREN && $i + 1 < count($parts) && $parts[$i + 1] === self::CLOSE_PAREN) {
+                $i += 2;
+            } else {
+                $filteredParts[] = $parts[$i];
+            }
+        }
+        return $filteredParts;
+    }
```
jlabedo, line 167: This is to handle a non documented behaviour, but there is one test with a method ending with `()`

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
dgafka: What about using `InterfaceToCallRegistry::getClassDefinitionFor`? 

From the PR I understood that intention to not use this is related to code visibility.
But `ClassDefinition::createFor` is not cached, therefore it will be resolved separately per each Pointcut.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Pointcut.php
```diff
@@ -22,17 +21,16 @@
  */
 class Pointcut
```
dgafka, line 22: Can we make use of package design here, and leave `Pointcut` and `PointcutExpression` on this namespace level as a form of an API, and move the rest of Pointcut implementations under `Ecotone\Messaging\Handler\Processor\MethodInvoker\Pointcut`?

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutExpression.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Ecotone\Messaging\Handler\Processor\MethodInvoker;
+
+use Ecotone\Messaging\Handler\InterfaceToCall;
+
+interface PointcutExpression
```
dgafka, line 7: Nice refactor with extracting all this logic :)

packages/Ecotone/src/Messaging/Config/Annotation/ModuleConfiguration/BasicMessagingModule.php
```diff
@@ -87,10 +87,6 @@ public function prepare(Configuration $messagingConfiguration, array $extensionO
         $messagingConfiguration->registerConverter(new SerializingConverterBuilder());
         $messagingConfiguration->registerConverter(new DeserializingConverterBuilder());
 
-        $messagingConfiguration
-            ->registerInternalGateway(TypeDescriptor::create(InboundGatewayEntrypoint::class))
-            ->registerInternalGateway(TypeDescriptor::create(EnrichGateway::class));
```
dgafka, line 92: Those have not been used?

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Pointcut.php
jlabedo: I hesitated and yes, it is better as you suggest

packages/Ecotone/src/Messaging/Config/Annotation/ModuleConfiguration/BasicMessagingModule.php
jlabedo: `MessagingConfiguration::registerInternalGateway` is dead code. I think the proxies are generated differently since ecotone "compilation" PR.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
jlabedo: Because the compilation is done during pointcut creation (`Pointcut::createWith`) and the registry is needed during compilation.
I had 3 possibilities:
- no cahing :)
- injecting InterfaceToCallRegistry on `Pointcut::createWith` but not sure it is always available, and it is called in a lot of places, I did not want to pollute the review of the PR.
- delaying compilation on first call to `Pointcut::doesItCut`, but in that case invalid pointcuts can be created without throwing and I have to change some tests.

What do you think ?

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
jlabedo: Do we keep it ?

dgafka: Yep, it's valid use case. Do you want to update docs?

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
jlabedo: It would be interesting to have the registry at construction time also here:
https://github.com/ecotoneframework/ecotone-dev/blob/9a2b937b166154db6b3b279aa2d39c2badccce89/packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Pointcut.php#L76

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
jlabedo: I don't really know what could be a good use case for this feature for updating the doc. If you tell me I can do it :)

dgafka: That why it was hidden :D 

But in general you could do something like `\Domain\Article || \Domain\User\User::register()`
So for example catch everything in given namespace, but you do have single endpoint from different place that you would like to pointcut

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
dgafka: > injecting InterfaceToCallRegistry on Pointcut::createWith but not sure it is always available, and it is called in a lot of places, I did not want to pollute the review of the PR.

This is how we do it with other components, but I see your point about amount of extra work. 
Anyways not a blocker for this one, as seems that this PR actually bring performance improvement even without caching. 
Let's go without the cache then :)

dgafka (approved):

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/PointcutParser.php
jlabedo: It is ok for the amount of extra work, it was for clarity for the reviewer :) I may add this in another PR.

PR #347, opened 2024-07-23
Register polling metadata only for polled endpoints

jlabedo:
## Description

Every endpoint has some default PollingMetadata registered in the container even if it is not a polled endpoint.
This PR registers PollingMetadata only for polled endpoints.
On a medium sized Symfony repository, I get a 10% speedup on cache:clear.

## Type of change

- Performance

### DIFF
--- a/packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
+++ b/packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
@@ -24,7 +24,7 @@ class MessagingContainerBuilder
     private array $pollingEndpoints = [];
     private ServiceConfiguration $applicationConfiguration;
 
-    public function __construct(private ContainerBuilder $builder, ?InterfaceToCallRegistry $interfaceToCallRegistry = null, ?ServiceConfiguration $serviceConfiguration = null)
+    public function __construct(private ContainerBuilder $builder, ?InterfaceToCallRegistry $interfaceToCallRegistry = null, ?ServiceConfiguration $serviceConfiguration = null, private array $pollingMetadata = [])
     {
         $this->interfaceToCallRegistry = $interfaceToCallRegistry ?? InterfaceToCallRegistry::createEmpty();
         $this->applicationConfiguration = $serviceConfiguration ?? ServiceConfiguration::createWithDefaults();
@@ -45,7 +45,7 @@ class MessagingContainerBuilder
         return $this->applicationConfiguration;
     }
 
-    public function registerPollingEndpoint(string $endpointId, Definition $definition): void
+    public function registerPollingEndpoint(string $endpointId, Definition $definition, bool $withContinuousPolling = false): void
     {
         if (isset($this->pollingEndpoints[$endpointId])) {
             throw new InvalidArgumentException("Endpoint with id {$endpointId} already exists");
@@ -56,6 +56,8 @@ class MessagingContainerBuilder
             throw new InvalidArgumentException("Endpoint runner {$className} must implement " . EndpointRunner::class);
         }
 
+        $pollingMetadata = $this->getPollingConfigurationForPolledEndpoint($endpointId, $withContinuousPolling);
+        $this->registerPollingMetadata($pollingMetadata);
         $this->register($runnerReference, $definition);
         $this->pollingEndpoints[$endpointId] = $endpointId;
     }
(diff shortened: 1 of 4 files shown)


packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
```diff
@@ -109,4 +102,32 @@ public function has(string|Reference $id): bool
...
         return $this->builder->has($id);
     }
+
+    private function getPollingConfigurationForPolledEndpoint(string $endpointId, bool $withContinuousPolling): PollingMetadata
+    {
+        if (array_key_exists($endpointId, $this->pollingMetadata)) {
+            $pollingMetadata = $this->pollingMetadata[$endpointId];
+        } else {
+            $pollingMetadata = PollingMetadata::create($endpointId);
+        }
```
lifinsky, line 112: $this->pollingMetadata[$endpointId] = $pollingMetadata;

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
```diff
@@ -109,4 +102,32 @@ public function has(string|Reference $id): bool
...
+
+        if ($this->applicationConfiguration->getDefaultErrorChannel() && $pollingMetadata->isErrorChannelEnabled() && ! $pollingMetadata->getErrorChannelName()) {
+            $pollingMetadata = $pollingMetadata
+                ->setErrorChannelName($this->applicationConfiguration->getDefaultErrorChannel());
```
lifinsky, line 116: Does set create new object?

jlabedo: yes :)

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
jlabedo: What do you mean ?

jlabedo: What do you mean ?

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
lifinsky: Why you assign set method result to variable? This is the same object.

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
lifinsky: You create object but do not update pollingMetadata array.

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
jlabedo: I know the API of PollingMetadata is not clear for that, but the set method is returning a clone

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
jlabedo: the pollingMetadata array is the custom endpoint polling configuration, we are enriching it with default configuration and returning the result. There is no need to update the pollingMetadata array, the only useful thing is the polling metadata returned

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
lifinsky: :(

packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
```diff
@@ -45,7 +45,7 @@ public function getServiceConfiguration(): ServiceConfiguration
         return $this->applicationConfiguration;
     }
 
-    public function registerPollingEndpoint(string $endpointId, Definition $definition): void
+    public function registerPollingEndpoint(string $endpointId, Definition $definition, bool $withContinuousPolling = false): void
```
dgafka, line 48: `InboundChannelAdapterBuilder` can be set up for cron execution, and continues polling is meant for [fixed rate](https://github.com/ecotoneframework/ecotone-dev/pull/347/files#diff-7044d70f2a96404a172741375a62e9217f6ef83f905e1af16c1864ef5f753f9fR128).

However I believe we could simplify this code a bit here, as it's called only for Async Endpoints. 
Therefore if we would add `fixed rate` as a default if PollingMetadata was not passed, then potentially we could remove `withContinuousPolling` config completely. 
Can you check that?

jlabedo: Do you mean something like this ?
```php
    // packages/Ecotone/src/Messaging/Config/Container/MessagingContainerBuilder.php
    private function getPollingConfigurationForPolledEndpoint(string $endpointId): PollingMetadata
    {
        if (array_key_exists($endpointId, $this->pollingMetadata)) {
            $pollingMetadata = $this->pollingMetadata[$endpointId];
        } else {
            $pollingMetadata = PollingMetadata::create($endpointId)
                ->setFixedRateInMilliseconds(1);
        }
```

Tests are failing in this case (for e.g. `test_serializing_command_and_event_before_sending_to_asynchronous_channel`)

dgafka (approved):
I've not checked if [that's possible](https://github.com/ecotoneframework/ecotone-dev/pull/347#discussion_r1691424049), so :green_apple:. 
If we can do it then cool, if not we just continue on the logic we had :)

PR #348, opened 2024-07-23
Drop unnecessary InterfaceToCall references in container

jlabedo:
## Description

InterfaceToCall are actually registered in the container, and there is a lot of registrations (InterfaceToCall, InterfaceParameter, Attributes).
This PR avoids injecting unnecessary InterfaceToCall references for service activators.
On a medium sized symfony project, I got 1000 objects dropped from the container after this PR, and got a 10% speed boost on cache:clear. 

## Type of change

- Performance


packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/AroundMethodInvocation.php
```diff
@@ -79,7 +79,7 @@ public function getMethodName(): string
 
     public function getInterfaceToCall(): InterfaceToCall
     {
-        return $this->interceptedMessageProcessor->getInterfaceToCall();
+        return InterfaceToCall::create($this->getObjectToInvokeOn(), $this->getMethodName());
     }
```
jlabedo, line 83: I supposed `MethodInvocation` was public API, so I kept `getInterfaceToCall()` in the interface and used this fallback.

dgafka: This will now need to resolve the Interface using reflection on run time. Let's drop this method then, so we won't accidentally use it.

dgafka: Feel free to merge and do it as follow up

dgafka (approved):

dgafka:
I can see in benchmark build, better performance on run time. I wonder what's the reasoning behind it. 
Like initializating new objects is the case here?

jlabedo:
> I can see in benchmark having better performance on run time with symfony, which dumps the container. I wonder what's the reasoning behind it. Like initializating new objects is the case here?

Maybe, this is the only rationale I have. But yes, InterfaceToCall may be quite expensive to load as there are a lot of objects to get from the container (InterfaceParameters and attributes)

PR #349, opened 2024-07-24
Drop unnecessary interface to call references in aggregate handlers

jlabedo:
## Description

Continue #348 with Aggregate handlers

## Type of change

- Performance

### DIFF
--- a/packages/Ecotone/src/Modelling/AggregateFlow/CallAggregate/AggregateMethodInvoker.php
+++ b/packages/Ecotone/src/Modelling/AggregateFlow/CallAggregate/AggregateMethodInvoker.php
@@ -2,12 +2,16 @@
 
 namespace Ecotone\Modelling\AggregateFlow\CallAggregate;
 
+use Ecotone\Messaging\Conversion\MediaType;
 use Ecotone\Messaging\Handler\ParameterConverter;
 use Ecotone\Messaging\Handler\Processor\MethodInvoker\AroundMethodInterceptor;
 use Ecotone\Messaging\Handler\Processor\MethodInvoker\AroundMethodInvocation;
 use Ecotone\Messaging\Handler\Processor\MethodInvoker\MethodInvoker;
+use Ecotone\Messaging\Handler\TypeDescriptor;
+use Ecotone\Messaging\Handler\Type;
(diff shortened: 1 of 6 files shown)


packages/Ecotone/src/Modelling/AggregateFlow/CallAggregate/AggregateMethodInvoker.php
```diff
@@ -0,0 +1,90 @@
...
+        private array $methodParameterNames,
+        private bool $isCommandHandler,
+        private array $methodParameterConverters,
+        private array $aroundMethodInterceptors,
+    ) {
+        Assert::allInstanceOfType($methodParameterConverters, ParameterConverter::class);
+        Assert::allInstanceOfType($aroundMethodInterceptors, AroundMethodInterceptor::class);
+    }
+
+    public function execute(Message $message): ?MessageBuilder
```
dgafka, line 38: Good logic extraction. 
I did had one problem with current Aggregrate related design, that even so we made it explicit whatever the class relates to `StateStored` or `EventSourcing`, there was a lot of logic that at end was actually an copy/paste.

It's really hard to test all the cases separately, especially that we also have also other cases like storing an Aggregate when it's returned from called Aggregate. This creates a combo of test cases to be sure that things are working. That's why I've introduced `SaveAggregateServiceTemplate` (which could actually be inject like `AggregateMethodInvoker` but had no enough time for that).
On other hand if the logic is shared, even if only single scenario is tested, we do have much higher confidence that it will work for other cases too.

jlabedo: I agree, i focused only on state aggregate as you pushed some changes to event sourced aggregates and i did not want to conflict with an eventual work in progress on your side.
I can propose a refactoring of code duplication in another pr if you are not working on it

dgafka: I am not working on that of the code atm, so feel free

dgafka:
conflicts in the PR :)

dgafka (approved):

PR #351, opened 2024-07-24
Lower stack size of around interceptors

jlabedo:
## Description

Lower call stack for around interceptors.
Checked on my project, the stack size for calling a command handler in an aggregate went from 72 to 64.

## Type of change

- Optimization


dgafka (approved):
:+1:

PR #354, opened 2024-07-26
In process channel

jlabedo:
## Description

This is a proposal to lower the stack trace of message handling.
It replaces direct and public subscribe channels with an "InProcessChannel" that will manage an internal stack of [handler, message] , and process them one by one.
It is particularly effective for before and after interceptors :
on main the stack is like this: `before -> before -> before -> intercepted handler -> after -> after -> after`
with this PR there will be multiple stack from the in process channel:
```
channel -> before
channel -> before
channel -> before
channel -> intercepted handler around
channel -> after
channel -> after
channel -> after
```

It has no impact on performance but I think it helps developer experience and debugging, at the cost of a more complex implementation.
For a simple aggregate command handler, the stack size is 51 on main and 41 after this PR (From gateway proxy to service)
I expected more, but I think the difference could be higher in more complex scenarios.
WDYT ?

For the mention, while working on this PR, I noticed that 'after' interceptors are called inside 'around' interceptors:
`before -> around ( intercepted -> after )`
while I was expecting: `before -> around( intercepted ) -> after`
Is it the way it should be ?

## Type of change

- DX Proposal

### DIFF
--- a/packages/Ecotone/src/Messaging/Channel/InProcessChannel.php
+++ b/packages/Ecotone/src/Messaging/Channel/InProcessChannel.php
@@ -24,13 +24,22 @@ class InProcessChannel implements SubscribableChannel, DefinedObject
 
     private bool $processing = false;
 
-    public function __construct(private string $messageChannelName)
+    public function __construct(private string $messageChannelName, private bool $isDirectChannel = false)
     {
     }
 
-    public static function create(string $messageChannelName = ''): self
+    public static function createDirectChannel(string $messageChannelName = '', ?MessageHandler $messageHandler = null): self
     {
-        return new self($messageChannelName);
+        $channel = new self($messageChannelName, true);
+        if ($messageHandler) {
+            $channel->subscribe($messageHandler);
+        }
+        return $channel;
+    }
+
+    public static function createPublishSubscribeChannel(string $messageChannelName = ''): self
+    {
+        return new self($messageChannelName, false);
     }
 
     /**
@@ -38,6 +47,10 @@ class InProcessChannel implements SubscribableChannel, DefinedObject
      */
     public function send(Message $message): void
     {
+        if ($this->isDirectChannel && \count($this->messageHandlers) !== 1) {
+            throw MessageDispatchingException::create("There is no message handler registered for dispatching Message {$message}");
+        }
+
         if ($message->getHeaders()->containsKey(MessageHeaders::IN_PROCESS_EXECUTOR_INTERCEPTING)) {
             Assert::isTrue(empty($this->messageStack), "Cannot send message to in process channel when it is already intercepted");
             foreach ($this->messageHandlers as $mainMessageHandler) {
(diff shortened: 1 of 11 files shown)


packages/Ecotone/src/Messaging/Channel/InProcessChannel.php
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return true;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function subscribe(MessageHandler $messageHandler): void
+    {
+        if ($this->isDirectChannel && \count($this->messageHandlers) === 1) {
```
lifinsky, line 113: !empty instead of count

packages/Ecotone/src/Messaging/Channel/InProcessChannel.php
```diff
@@ -0,0 +1,146 @@
...
+            throw WrongHandlerAmountException::create("Direct channel {$this->messageChannelName} have registered more than one handler. {$messageHandler} can't be registered as second handler for unicasting dispatcher. The first is {$this->messageHandler}");
```
lifinsky, line 114: {$this->messageHandlers[0]}

packages/Ecotone/src/Messaging/Channel/InProcessChannel.php
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * @inheritDoc
+     */
+    public function send(Message $message): void
+    {
+        if ($this->isDirectChannel && \count($this->messageHandlers) !== 1) {
+            throw MessageDispatchingException::create("There is no message handler registered for dispatching Message {$message}");
+        }
+
+        if ($message->getHeaders()->containsKey(MessageHeaders::IN_PROCESS_EXECUTOR_INTERCEPTING)) {
```
dgafka, line 54: Using headers to decide how Messages are handled is always tricky, as it introduces state. 
This is because the state based logic is much harder to test, as it means that there is one more path to actually do given things.  

With Aggregates we do state based flows on Headers, however we do actually have a state that there which we load from database there. On lower level abstraction everything can be stateless actually, as it's just Message, Pipe and Endpoint. 
Not to mention that for Aggregates it took time to make the abstraction solid and battle tested (e.g. headers been left when sending to other channels, and accidentally loading wrong Aggregate).

So I do agree it would be useful to simplify the call stack to lower the entrypoint, but let's not do it on the cost of introducing state :)

jlabedo: Right, I was not really convinced of that one.

Regarding this

> For the mention, while working on this PR, I noticed that 'after' interceptors are called inside 'around' interceptors:
before -> around ( intercepted -> after )
while I was expecting: before -> around( intercepted ) -> after

Is it the way it should be ?

dgafka: To be honest I don't remember if it was by design or not :) 
Could be that it was in order to decrease the stack size.

I guess from user perspective it make no difference, right?

lifinsky: I think it would be correct to process "after" pointcuts after "around"

jlabedo: I don't use After interceptors in my code, so it is a bit tricky to find a valid use case. I tried here: https://github.com/jlabedo/ecotone-dev/pull/3

On run, this code will output:
```
The result is not a QueryResult !
array (
  'status' => 'success',
  'executionTime' => NULL,
  'result' => 'Milk',
)% 
```

PR #358, opened 2024-07-30
Proposal: apply `Before`, `Around` and `After` intreceptors on message processors instead of message handlers

jlabedo:
## Description

This is a WIP proposal to refactor interceptors and message processors.

`Before`, `Around` and `After` intreceptors are all about in process method invocation and not about messaging, while `BeforeSend` is about messaging, so I am not talking about this last one in this proposal.

The proposal is to apply `Before`, `Around` and `After` to Message Processors instead of Message Handlers. A message processor can be a user space method invocation or an internal Message Processor like all aggregate flow services.

This PR introduces a new interface named `RealMessageProcessor` which will replace the actual `MessageProcessor` interface. This interface has only one method: `process(Message $message): ?Message`

This PR also introduces `MessageProcessorActivator` which would replace `RequestReplyProducer` but will only handle messages instead of mixed return types.

This PR includes only a refactoring of the `ModellingHandlerModule` with the new components. If we agree on the design, we could easily change all modules.

The goal of this proposal are:
- better design : there is actually a lot of piping around interceptors that is hard to follow. I think because they are not about messaging but method invocation
- performance: the execution is more straight forward, I think it could impact performance. It will also positively impact compilation time as there will be less components to compile.
- call stack: the call stack will be a lot simpler

In the end, for example for user space command handler, there would be a single Message Handler registered with a `ChainMessageProcessor` that will handle all `Before`, `Around`, `After` interceptors and the actual method invocation before returning to the Message handler.

WDYT ?


packages/Ecotone/src/Messaging/Handler/Processor/MethodInvocationProcessor.php
```diff
@@ -0,0 +1,96 @@
...
+            ? $objectToInvokeOn::{$this->methodInvoker->getMethodName()}(...$params)
+            : $objectToInvokeOn->{$this->methodInvoker->getMethodName()}(...$params);
+
+
+        if (is_null($result)) {
+            return null;
+        }
+
+        if ($this->shouldChangeMessageHeaders) {
+            Assert::isFalse($result instanceof Message, 'Message should not be returned when changing headers in ' . $this->interfaceToCallName);
```
lifinsky, line 40: It will be better to have additional methods in Assert class: isArray, isNotInstanceOf, isInstanceOf, etc

packages/Ecotone/src/Messaging/Handler/RealMessageProcessor.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Ecotone\Messaging\Handler;
+
+use Ecotone\Messaging\Message;
+
+interface RealMessageProcessor
```
lifinsky, line 7: What is **real** message?

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
```diff
@@ -732,16 +721,18 @@ private function registerLoadAggregate(ClassDefinition $aggregateClassDefinition
...
                 )
                 ->chain(
-                    ServiceActivatorBuilder::createWithDefinition(new Definition(FetchAggregate::class), 'fetch')
-                        ->withMethodParameterConverters([
-                            HeaderBuilder::createOptional('aggregate', AggregateMessage::CALLED_AGGREGATE_OBJECT),
-                            HeaderBuilder::createOptional('aggregate', AggregateMessage::RESULT_AGGREGATE_OBJECT),
-                        ])
+                    new class implements CompilableBuilder {
+
+                        public function compile(MessagingContainerBuilder $builder): Definition|\Ecotone\Messaging\Config\Container\Reference
```
lifinsky, line 726: Add Reference to import

packages/Ecotone/src/Modelling/LoadAggregateService.php
```diff
@@ -15,7 +16,6 @@
 /**
  * licence Apache-2.0
  */
-interface LoadAggregateService
+interface LoadAggregateService extends RealMessageProcessor
```
lifinsky, line 19: Interface without own methods is not ok

lifinsky:
Maybe PR should only have one reason for change: fixing an issue with the after interceptor?

lifinsky:
Interceptors should apply to method calls, not just message handlers (as AOP functionality)

dgafka:
@jlabedo I can't review now, but I can leave a comment to explain reasoning behind different Interceptors. 


So what interceptors are about is related to what type of interceptor we want to use. 
https://docs.ecotone.tech/modelling/extending-messaging-middlewares/interceptors

### `Before and After Interceptors` - 
image

Are meant to hook in into Message flow. For example `before` does not have access to the invocated method at all, because it does happen before we even get into the invoked object. 
This allows for example for filtering out messages, or stopping invocation before even Aggregate is loaded. 
Besides that, in this situation Message can be enriched with details (for example we may add userId to the payload, as Command can be an simple array at this point - is not deserialized to Command Class yet)
In general sense we can view them as features optimized for guarding, filtering and enriching.

### `Presend Interceptors`
image

This is special type of Before Interceptor. The best option to view the use case for that, is having `Asynchronous Command Handler`. If we would use an Before Interceptor with Async Endpoint, Message would land first in the Message Channel and then after it would be consumed it would trigger Before Interceptor. 
This is far from ideal for scenarios where we want to validate if this Message even make sense. So for things like structure validation, or even authentication we want to validate those before they will land in Async Message Channel. 
This way we can ensure we don't push Messages to the Queue which make no sense, and only the correct Messages that we know we want to process will go there.

- `Around Interceptors` - Those are about Method Invocation. This is the moment just before we execute intercepted object, and this is the moment when we actually deserialized the payload to given Command/Query/Event class (Before that Ecotone just carry whatever was given in the payload, it can be json/array for example). 
The previous interceptors are not able to do action before method execution and after method exection (as they are not even aware of the executed object). 
So if we would do intercept Aggregate, we would get access to the Aggregate itself and Command. It make it really easy to build stuff like, call `getUserId` and compare it with `current executor id`, if not in match then this Aggregate does not belong to the executor, so he can't modify it. 
This is also perfect fit for intercepting Gateways like Command Bus, as we can for example start transaction before Gateway execution and finish it after, or just rollback on error.

----
[Example from Lesson 5](https://docs.ecotone.tech/tutorial-php-ddd-cqrs-event-sourcing/php-interceptors-middlewares)
image

It may be not obvious to reason about all those stuff, because what we are used to, is just Middleware pattern available in other Framework, and Interceptors are much more than that. 
Therefore, I hope above helps to understand that Interceptors are not only about Method Execution. In fact only one Interceptor is about method invocation, the rest of them are actually about the Message flow. 

@jlabedo let me know then if this change make sense, and we should continue with the PR. If so, then happy to review.

jlabedo:
Thanks @dgafka for your explanations.

We agree on `Presend` interceptors (they relate to messaging) and `Around` interceptors (they relate to method invocation).

When I think of messaging in my proposal, it is really about channels and the possibility made by ecotone to switch a channel implementation from direct to asynchronous, and that's what makes the power of Ecotone 🚀 :) 

Regarding `Before` and `After` interceptors, the question is: would you change the channel between a `Before` interceptor and its endpoint to something else than a `DirectChannel` ? It is not implemented now, but would the feature be desirable ?
Well, I think It could, but this is already implemented differently with Internal handlers and output channels : you can see them as `Before` or `After` interceptors and they relate to messaging as you can easily switch from a synchronous to an asynchronous channel.

But if `Before` and `After` interceptors are always meant to be run in the same process than the endpoint, we can make it clear in the design and drop the messaging channel abstraction for `Before` and `After`, which would be useless, and I think ecotone code would be easier to understand (and more performant).

jlabedo:
> Maybe PR should only have one reason for change: fixing an issue with the after interceptor?

@lifinsky , this PR is not meant to fix the issue with after interceptors precedence, just to change the design of interceptors with the exact same specifications for the user. But if we decide to get after interceptors out of around interceptors, I think this new design would make the work a lot easier.

dgafka:
> But if Before and After interceptors are always meant to be run in the same process than the endpoint, we can make it clear in the design and drop the messaging channel abstraction for Before and After, which would be useless, and I think ecotone code would be easier to understand (and more performant).

Yes, that's correct they will be only used in synchronous scenarios, before and after execution as this is design decision.
So ye I see your point and it make sense :)

So if we can refactor to simplify it (possible shorten the stack too), yet keep the behaviour they provide, then we can go for that

dgafka:
> This PR introduces a new interface named RealMessageProcessor which will replace the actual MessageProcessor interface. This interface has only one method: process(Message $message): ?Message

Why would we replace MessageProcessor? 

> This PR also introduces MessageProcessorActivator which would replace RequestReplyProducer but will only handle messages instead of mixed return types.

`RequestReplyProducer` is used as foundation for communication between Endpoints, and if I understood correcty the aim of the PR is to handle synchronous Before/After/Around Interceptors an the execution to not use Messaging. 
However other parts will still use Messaging therefore will be in need for functionality which `RequestReplyProducer` provides.
Can you elaborate why would we replace that and how does it related to the PR? 

> better design : there is actually a lot of piping around interceptors that is hard to follow. I think because they are not about messaging but method invocation
Agreed, we can think of it like this, due it's synchronous nature. 

> performance: the execution is more straight forward, I think it could impact performance. It will also positively impact compilation time as there will be less components to compile.
Ok make sense.

> call stack: the call stack will be a lot simpler
Sure, maybe for sync scenarios, it will be easier for Users to understand what's happening, if we will simplify that.

> In the end, for example for user space command handler, there would be a single Message Handler registered with a ChainMessageProcessor that will handle all Before, Around, After interceptors and the actual method invocation before returning to the Message handler.

I see, make sense. Thank you sharing your thoughts on that, it gave me higher perspective on how we can look on this. 
To sum it up in context of interceptors:

- For places where async/sync communication is possible we continue using Message Channels. This way we can easily continue switching the code from async to sync on design level.
- For places where sync communication is the the only possible road, we still continue to use Messages, however we won't be using Message Channels. We will have dedicated Message Processor that will handle intercepting and actual (intercepted) object invocation. 

Let me me know if above statements are true or is there something more to add :)

packages/Ecotone/src/Messaging/Handler/RealMessageProcessor.php
dgafka: This interface name does not tell me much either. Isn't the other Processors real too? :P

jlabedo:
An existing piece of code that shows interceptors do not fit with message channels 😄 :
```php
    private function checkIfInterceptorIsCorrect(MethodInterceptor $methodInterceptor): void
    {
        if ($methodInterceptor->getMessageHandler()->getEndpointId()) {
            throw ConfigurationException::create("Interceptor {$methodInterceptor} should not contain EndpointId");
        }
        if ($methodInterceptor->getMessageHandler()->getInputMessageChannelName()) {
            throw ConfigurationException::create("Interceptor {$methodInterceptor} should not contain input channel. Interceptor is wired by endpoint id");
        }
        if ($methodInterceptor->getMessageHandler()->getOutputMessageChannelName()) {
            throw ConfigurationException::create("Interceptor {$methodInterceptor} should not contain output channel. Interceptor is wired by endpoint id");
        }
    }
```

jlabedo:
Hello guys, I am having strange behavior in the actual implementation that is hard to maintain with the new implementation. This is the code under test :

```php
// packages/Ecotone/tests/Messaging/Fixture/InterceptedBridge/BridgeExample.php

final class BridgeExample
{
    #[ServiceActivator('bridgeExample', outputChannelName: 'bridgeSum')]
    public function result(int $result): int
    {
        return $result;
    }

    #[ServiceActivator('bridgeSum')]
    public function sum(int $amount): int
    {
        return $amount + 1;
    }

    #[Around(precedence: 0, pointcut: BridgeExample::class . '::result')]
    public function multiply(MethodInvocation $methodInvocation): int
    {
        return $methodInvocation->proceed()->getPayload() * 3;
    }
}
```

The actual test expects to return 6 when sending 1 to `bridgeExample`. The flow is:

```
- Around (multiply)
  | ::result(1) = 1
  | ::sum(1) = 2
  -> proceed(2) * 3 = 6
```

Two things to note on this code:
- the output channel is executed inside the around interceptor. Why not, but then around interceptors are related to messaging in a way. This is not obvious to me.
- Looking at the around interceptor code, why would you expect to receive a `Message` as the return value of the `proceed` invocation (`$methodInvocation->proceed()->getPayload()`) ?

Finally, the chain is different if you add a dumb `After` interceptor:
```php
final class BridgeExample
{
    #[ServiceActivator('bridgeExample', outputChannelName: 'bridgeSum')]
    public function result(int $result): int
    {
        return $result;
    }

    #[ServiceActivator('bridgeSum')]
    public function sum(int $amount): int
    {
        return $amount + 1;
    }

    #[Around(precedence: 0, pointcut: BridgeExample::class . '::result')]
    public function multiply(MethodInvocation $methodInvocation): int
    {
        return $methodInvocation->proceed()->getPayload() * 3;
    }

    #[After(pointcut: BridgeExample::class . '::result')]
    public function after(int $amount): void
    {
        $test = $amount;
    }
}
```

Then the result becomes 4 instead of 6. Here is the chain with this new `After` interceptor:
```
- Around (multiply):
  | result(1) = 1
  | after(1) = 1
  -> proceed(1) * 3 = 3
- sum(3) + 1 = 4
```

In this case the output channel is executed outside the around interceptor...

So I propose to keep this last behaviour, meaning:
1. First run the ServiceActivator (bridgeExample) with all its interceptors (before, around, after)
2. Then send the result to outputChannel

WDYT @dgafka ?

lifinsky:
- Around (multiply): 1 * 3
- After: 3
- outputChannelName (bridgeSum: 3 + 1

jlabedo:
Hello Guys,

I finally realized that aggregates have another behavior. I added tests in #359 to explain it better.
Aggregates have the "correct" behavior in the sense that after interceptors are called outside of around interceptors.

So I propose to keep this behavior for all invocations:
1. It is the more natural behavior to me
2. It's way simpler to write with the new proposal implementation

This would be a breaking change:
- for code relying on after calls being inside around interception (but why would you do that ?)
- for code relying on the return value of the `proceed` method : in the case of aggregates it is the aggregate itself (we keep the same logic) but for non aggregates it is a `Message`. In the latter case the return value would be changed with the new implementation to the return value of the intercepted method invocation.

It would be nice if we agree with this BC @dgafka : the implementation will be a LOT easier as it is more natural, but I will have to change a lot of unit tests (we may discuss after if we keep them or not, some are testing implementation details).


To recap, the invocation interception would be like this:
- **before interceptors**
- internal message processors (load aggregate, identifier resolution, ...)
- aggregate method invocation (with eventual **around interceptors**)
- **after interceptors**
- internal message processors (save aggregate, publish events, ...)

lifinsky:
@jlabedo Important to load and save aggregate inside transactional around interceptor.

jlabedo:
> @jlabedo Important to load and save aggregate inside transactional around interceptor.

Right, those are handled on gateways interceptors, upper in the chain.

dgafka:
@jlabedo I will have more time to look on this later this week. 
For now, what I can leave feed for thoughts, because I just remind myself of more advanced scenarios.

In case of Gateways around with naturally wrap everything that is happening inside and all the flows that happens after. 
```php
- gateway (start & begin transaction)
      - command handler (start)
          - event handler (start)
          - event handler (end)
      -  command handler (end)
- gateway (end & commit transaction)
```

Now this actually have a valid use case to track sub-flows when we around intercept Endpoints (not gateways) too
We may trigger an [Transaction in Command Handler](https://docs.ecotone.tech/modules/dbal-support#transactions) and expect subflows to be within transaction too

```php
class OrderService
{
    #[DbalTransaction]
    #[CommandHandler('order.register', 'orderService')]
    public function order(string $orderName): void
    {
        throw new InvalidArgumentException('exception');
    }
```

```php
- command handler (start & begin transcation)
    - event handler (start)
    - event handler (end)
- command handler (end & end transaction)
```

The same use case is used for Tracing. Where we hook in for tracing and use context

```php
- command handler (start and begin upper context)
    - event handler (start and begin lower context)
    - event handler (end and end lower context)
- command handler (end & end upper context)
```

This allows us to actually create meaningful traces that shows parent-child relations in OpenTelemetry.
Correct me if I am wrong, but with this proposal we would lose this, as those wouldn't really be children anymore, but same level actions e.g.

```php
- command handler (start and begin upper context)
- command handler (end & end upper context)
- event handler (start and begin upper context)
- event handler (end and end upper context)
```

jlabedo:
@dgafka , I agree and changed my mind during the night because of what @lifinsky said :) 

Just to clarify first, on `main`, the event handlers are NOT handled inside the command handler around interceptors if you are using the `WithEvents` feature (see #361 for test case). If you are using the event bus directly in the command handler, yes, those are dispatched immediately inside the command handler, so also in the around interceptor.
Today, the around interceptor is around the `CallAggregateService`, so you will intercept the call, eventually any event handlers called inside the command handler. Then transaction ends, and **then** the `SaveAggregateService` is called (outside the transaction)

So I came to the same conclusion that having around interceptors just around the method call is not really useful. Around interceptors should intercept the whole message processing chain to be useful. So I propose this chain (it could easily be changed):
```
- Before interceptors
- START around
   - internal handlers (load, etc.)
   - call aggregate
   - (1) After interceptors (so we can add headers before saving for instance)
   - internal handlers (save, publish events)
- END around
- (2) After interceptors (so we can be sure the transaction ended)
```

I hesitate between (1) and (2) for `After` interceptors. But think I would go for (2) as it seems more natural and it is the actual behavior for aggregates. But we could easily change this behavior, so let me know what you think.

Finally, to answer @dgafka question, we will keep the same calling structure, also for tracing:
```
- gateway (start & begin transaction)
      - command handler (start)
          - event handler (start)
          - event handler (end)
      -  command handler (end)
- gateway (end & commit transaction)
```

jlabedo:
Hello guys,
tests are passing with minimal changes (in tests 😄). There still need some cleaning, this is still a proof of concept.

Regarding performance, there is a 15% speed boost in HttpStackBenchmark, I expect more when all components are moved to this new implementation. I expect compilation time to be improved, because less components are registered in the container (but we don't have a benchmark for that).

Regarding final behavior, the only difference is gateways and service after interceptors that are called outside of around interceptors (for aggregates, this was the existing behaviour)

I will clean it a bit more to explain how things are working here

dgafka:
> I will clean it a bit more to explain how things are working here

Ok, will review after you post your final summary :)

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInvocationWithChainedAfterInterceptorsProvider.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Ecotone\Messaging\Handler\Processor\MethodInvoker;
+
+use Ecotone\Messaging\Handler\RealMessageProcessor;
+use Ecotone\Messaging\Message;
+
+/**
+ * This class provides ability to call after interceptors inside method invocation
+ */
+class MethodInvocationWithChainedAfterInterceptorsProvider implements MethodCallProvider
```
dgafka, line 11: Can you provide licence on the top of new files?

```php
/**
 * licence Apache-2.0
 */
```

jlabedo: Do you use any automated tool for that ?

packages/Ecotone/src/Messaging/Handler/RealMessageProcessor.php
jlabedo: This was the name used during the transition to new implementation, now it is `MessageProcessor`.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInvocationWithChainedAfterInterceptorsProvider.php
dgafka: I had one that was adding on the top, but it was to add each file. Therefore it would just double the licence in existing files. 
So not automation at the moment

jlabedo:
So here are the new abstractions introduced:

## `MessageProcessor`
```php
interface MessageProcessor
{
    public function process(Message $message): ?Message;
}
```
This is the major building block of the messaging system. It can transform a message and/or execute side effects. A lot of existing internal classes did have this signature but now the abstraction is clear.

## `MethodInvocationProvider`
```php
interface MethodInvocationProvider
{
    public function getMethodInvocation(Message $message): MethodInvocation;
}
```
It takes a message and builds a `MethodInvocation`. This provides an abstraction for retrieving the object to invoke on from message (for aggregates)

## `ResultToMessageConverter`
```php
interface ResultToMessageConverter
{
    public function convertToMessage(Message $requestMessage, mixed $result): ?Message;
}
```
This is where we transform the result of a `MethodInvocation` to a `Message`. It can change headers, payload or passthrough request message.

## `RequestReplyProducer`
It handles executing a `MessageProcessor` and resolving the channel to send the result message to. The message building logic is now handled inside message processor and the splitting handler is moved to a dedicated class (`SplitterHandler`). 

## `SplitterHandler`
The splitter handler cannot respond to the reply channel defined in the message anymore, this can only be done in `RequestReplyProducer`. So an output channel is required for the splitter handler. This seems natural, as a splitter should be an intermediary message handler, not the final endpoint.

## `MessageProcessorActivatorBuilder` and `ServiceActivatorBuilder`
It would have been too much refactoring to change all `ServiceActivatorBuilder` instances to the new `MessageProcessorActivatorBuilder`, so `ServiceActivatorBuilder` is kept and use `MessageProcessorActivatorBuilder` internally. The incompatibility comes from the `changeHeaders` and `shouldPassThroughMessage` properties of `ServiceActivatorBuilder` that are now handled by the message processor and not the message handler itself.

## Interception logic
The interception logic is handled in 4 places:
- `GatewayProxyBuilder` for all 3 interceptors types
- `MessageProcessorActivatorBuilder` for after and before interceptors
- `InterceptedMessageProcessorBuilder` interfaces that can hook `Around` interceptors in their method call definitions.
- `MessagingSystemConfiguration` for `Presend` interceptors (not moved)

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
```diff
@@ -0,0 +1,129 @@
...
+        foreach ($interceptorsConfiguration->getBeforeInterceptors() as $beforeInterceptor) {
+            $compiledProcessors[] = $beforeInterceptor->compileForInterceptedInterface($builder, $this->interceptedProcessor->getInterceptedInterface(), $this->getEndpointAnnotations());
+        }
+        foreach ($this->processors as $processor) {
+            if ($processor === $this->interceptedProcessor) {
+                $compiledProcessors[] = $processor->compile($builder, $interceptorsConfiguration);
+                // register after interceptors
+                foreach ($interceptorsConfiguration->getAfterInterceptors() as $afterInterceptor) {
+                    $compiledProcessors[] = $afterInterceptor->compileForInterceptedInterface($builder, $this->interceptedProcessor->getInterceptedInterface(), $this->getEndpointAnnotations());
+                }
```
dgafka, line 116: I think we should move this  after processors.

I know that the result of this will be that Event Handler will be executed before the  `Before Interceptors`. But this can be understood as using Event Bus within Command Handler, as the result would be the same. 

This would be in sync with how before interceptors does work, as they are happening before all Processor, not just before Intercepted Processor.

Wdyt?

packages/Ecotone/src/Messaging/Handler/Gateway/GatewayProxyBuilder.php
```diff
@@ -415,59 +420,58 @@ private function compileGatewayInternalHandler(MessagingContainerBuilder $builde
...
+        if ($aroundInterceptors) {
+            $aroundInterceptors = $this->getSortedInterceptors($aroundInterceptors);
+            $compiledAroundInterceptors = array_map(
+                fn (AroundInterceptorBuilder $aroundInterceptor) => $aroundInterceptor->compileForInterceptedInterface($builder, $interceptedInterfaceReference, $this->endpointAnnotations),
+                $aroundInterceptors
+            );
+            $messageProcessorInvocation = new Definition(MessageProcessorInvocationProvider::class, [
+                $gatewayInternalProcessor,
+            ]);
+            $resultConverter = new Definition(PayloadResultMessageConverter::class, [$interceptedInterface->getReturnType()]);
```
dgafka, line 439: This seems like would be needed in all cases when Gateway can return value. 

Weird that no test has failed. Do I miss something or that's something to fix?

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInvokerBuilder.php
```diff
@@ -2,53 +2,101 @@
...
+            $interfaceToCall,
+            $this->methodParametersConverterBuilders,
+            $interfaceToCall,
+            $this->endpointAnnotations,
+        );
+        if ($interceptorsConfiguration) {
+            $methodCallProvider = $builder->interceptMethodCall($this->interfaceToCallReference, $this->endpointAnnotations, $methodCallProvider);
+        }
+
+        // @todo: this is only used for tests, in production it is always payload converter
```
dgafka, line 83: Can you explain?

jlabedo: The `MethodInvokerBuilder` is only used in `ServiceActivatorBuilder` and `SplitterHandlerBuilder`. All interceptors builders are not using the `MethodInvokerBuilder` anymore.
So my comment is wrong because we still have two cases that uses the `MethodInvokerBuilder` with different result converter:
- `ServiceActivator` attribute with the `changingHeaders` property to true
- `LoggingHandlerInterceptor` that uses the `withPassThroughMessageOnVoidInterface`

There is some sort of code duplication between the `TransformerBuilder` and the `MethodInvokerBuilder` but I did not want to go further in the same refactoring :) 

So for now I would remove this comment and see if there is code duplication (or maybe not) in another PR

packages/Ecotone/src/Messaging/Handler/Gateway/GatewayProxyBuilder.php
jlabedo: The behavior did not change: if the result is null, the converter will return null. Or I misunderstood your comment ?

dgafka:
In general very nice design and idea with this refactor. The code is more readable now and feels like allows to work with higher level of abstraction without bothering with lower ones.
What I mean by that is when I was refactoring the parts with Invocation, I was not really dealing with Messaging. 
Therefore it's like working on given level of abstraction without the need of analysis the other (of course to some degree).

Few thoughts on the PR :) 

- `MethodInvocationProvider` interface felt a bit over abstraction here, and created code duplication along the way. 
I've improved MethodInvoker, so it can be used in those scenarios instead. There is also an LazyMethodInvoker when we do not have yet access to invoked object (aggregates). 
- MethodInvoker will take care also of binding the interceptors, as we've moved it out of Messaging Config, therefore we now need a single place where this automatically hooked in. This will ensure that will happen for all Message Processors using of the Invoker.
- I've not managed to refactor all places, so some code is still using `MethodInvocationProvider`

packages/Ecotone/src/Messaging/Handler/Gateway/GatewayProxyBuilder.php
```diff
@@ -415,59 +420,58 @@ private function compileGatewayInternalHandler(MessagingContainerBuilder $builde
...
+            $messageProcessorInvocation = new Definition(AroundMethodInvocationProvider::class, [
+                $messageProcessorInvocation,
+                $compiledAroundInterceptors,
+            ]);
+            $messageProcessors[] = new Definition(MethodInvocationProcessor::class, [
+                $messageProcessorInvocation,
+                $resultConverter,
+            ]);
```
dgafka, line 447: Refactor to MethodInvoker is missing here

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInterceptorBuilder.php
```diff
@@ -0,0 +1,121 @@
...
+//        @TODO this fails
+//
+//        return MethodInvokerBuilder::createForInterceptor(
+//            $this->interceptorDefinition,
+//            $interceptorInterface,
+//            $this->defaultParameterConverters,
+//            $endpointAnnotations,
+//            $interceptedInterface
+//        )
+//            ->compile($builder);
```
dgafka, line 104: Missing refactor to make use of MethodInvocation. I've tried to do it here, but got failing tests

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
jlabedo: I don't have any good use case in mind for after interceptors. So I tried to make the minimum behavioral changes in this PR, then we can take time to think about it and change it once with the proper documentation, to avoid going back and forth with different user land behaviors.
If you apply `After` interceptors after the "save" and "publish" processors, it could be a really hard BC. I guess one of the relevant use case of the after interceptor is to propagate headers to published events. In that case, headers won't be propagated.

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/MethodInterceptorBuilder.php
jlabedo: For the record, you forgot to call `shouldPasstroughOnVoidInterface`

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
dgafka: OK, let's keep the current behaviour. 
We could add test case for enriching events, as right now it's more like a hidden gem, than actual feature :P

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
```diff
@@ -0,0 +1,129 @@
...
+        $compiledProcessors = [];
+        $interceptorsConfiguration = $this->interceptedProcessor
+            ? $builder->getRelatedInterceptors(
+                $this->interceptedProcessor->getInterceptedInterface(),
+                $this->getEndpointAnnotations(),
+                $this->getRequiredInterceptorNames()
+            )
+            : MethodInterceptorsConfiguration::createEmpty();
+        // register before interceptors
+        foreach ($interceptorsConfiguration->getBeforeInterceptors() as $beforeInterceptor) {
```
dgafka, line 107: Before intercepting was done on MessagingConfiguration level. This ensured that all Handlers are intercepted. 
1. How MessageFilter or Router are before intercepted for example now?
2. If we need to copy paste this config, can we change this so for example MessageFilter does use MessageProcessorActivator? It would be good to just be assured that if we introduce new Handler, it will work out of the box either because of abstraction reuse, or by abstraction handling it automatically.

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/InterceptedPollingConsumerBuilder.php
```diff
@@ -30,49 +28,14 @@
...
-        $this->aroundInterceptorReferences[] = $aroundInterceptorReference;
-
-        return $this;
-    }
-
-    /**
-     * @param MethodInterceptor $methodInterceptor
-     * @return $this
-     */
-    public function addBeforeInterceptor(MethodInterceptor $methodInterceptor): self
```
dgafka, line 54: Where does this happen now?

jlabedo: This is handled at the GatewayProxyBuilder level.

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
jlabedo: - `MessageFilter` became a message processor, handled by `MessageProcessorActivator`. Interception is done there.
- `SplitterHandlerBuilder` has some sort of code duplication with `MessageProcessorActivatorBiuilder` to handle interception
- `Router` is not before or after intercepted. I think I let it go as I am not sure of the use case of intercepting a `Router` (add headers ? stop message flow ?) and it passes tests :)

Options I see for Router:
1. Do not intercept `Router` endpoint
2. Add same small code duplication to `RouterBuilder`
3. Add small abstraction to remove the small code duplication inside `RouterBuilder` and `SplitterBuilder`
4. Change interceptors behaviour in this PR: I think that the code duplication will naturally be removed if we discuss a better behavior for interceptors, but I have the feeling that the discussion will lead to a breaking change. So I would prefer discussing it after this PR is merged :) My personal intuition is that around interceptors should intercept a message processor instead of a method invocation. This could be the `MethodInvokerProcessor`, so the interception is practically the same as the current behavior, but with a common implementation as for gateways. It could also be a "larger" processor to include the "save aggregate processor" for instance to fix `DbalTransaction` attribute (I am pretty sure it is currently broken if applied on an endpoint, I will try to add a failing test on main branch). But the `MethodInvocation` interface would be broken in that case, as there is no point to `getArguments` or to `replaceArguments` on a `MessageProcessor`.

dgafka: > MessageFilter became a message processor, handled by MessageProcessorActivator. Interception is done there.

image

I can only see that MessageFilter was built on top of `MethodInvoker`, yet `MessageProcessorActivatorBuilder` is not used here? Therefore the only interception available will be the around interception. 

----
So the around interceptors are hooked only in `MethodInvoker` and `Gateway`. And `MethodInvoker` is like a base abstraction for all Message Handlers, and Gateway is a separate thing. We won't be adding new `Gateways abstraction`, and new `Message Handler` will follow on using the `MethodInvoker`, therefore this is safe. 

The problematic part is the `before` and `after`, which was so far even if implicitcly (like with Router) able to hook into any Message Handler. With current design this is done on demand. 
So we lose ability to hook into basically anything (I am putting aside whatever `after` interceptor make sense in case of Router :P). 

So the flow is pretty much similar for any Message Handler. Register `before`, `setup Invoker`, `register after`:
image

I would say it's fine that we make it explicit in the Handler Builder that given Handler is hookable, but let's delegate this logic to one place, so we can reuse it across Handlers. So the small abstraction you've mentioned :)

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/InterceptedPollingConsumerBuilder.php
dgafka: Ah, so Gateway will internally fetch the Interceptors, so we don't need to register them on the ConsumerBuilder level?

jlabedo: Yes exactly. WDYT ?

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
jlabedo: The `FilterBuilder` should be chained in a `MessageProcessorActivatorBuilder` that will handle before and after interception.
But you are right, there is a problem with current implementation: around interception is activated in every cases, even if there the message processor was not meant to be intercepted.

jlabedo: @dgafka , Router is not intercepted in `main` as it does not implement interface `MessageHandlerBuilderWithOutputChannel`. Or I am missing something ?

dgafka: @jlabedo ye, it can be intercepted before with current solution thought :)

jlabedo: Well, with the very last refactor, all interceptors are chained in a dedicated builder. This is all or nothing. Do you want a dedicated behavior for `Router` ?

dgafka (approved):

packages/Ecotone/src/Messaging/Handler/ServiceActivator/MessageProcessorActivatorBuilder.php
dgafka: We could actually add intercepting for Router like for any other Handler. 
It will be more understandable from user perspective, as Router is actually [user-land facing component](https://docs.ecotone.tech/messaging/messaging-concepts/message-endpoint/message-routing).

But not a blocker for this PR

PR #360, opened 2024-08-01
Dynamic message channels

### LINKED ISSUE #357: Support consuming multiple queues with a single command
Currently to consume messages from a queue you run a command, for example in Symfony:

```
bin/console ecotone:run async
```

It would be very useful to be able to consume multiple queues using a single worker. For example:

```
bin/console ecotone:run queue1 queue1 queue2
bin/console ecotone:run --all
```

For my use case, I have several really small apps running on Digital Ocean app platform for $5/month, but it uses two different queues (one for local async processing, the second a distributed bus). 

To add these workers it's an additional $5/month which is tripling the cost for each app. If it were under huge load this obviously would not be a problem, but for small low-traffic apps having a single worker for all queues would be really helpful.

Thanks

dgafka:
## Description

This introduces Dynamic Message Channels to Ecotone Enterprise. This may be useful to:
- Simpler deployment strategy
- Optimize system resources 
- Adjust the configuration per Client which is especially useful in Multi-Tenant Environments.

## Part of incoming documentation

image

image

image

image

Fixes #357

## Type of change

- This change requires a documentation update


dgafka:
Documentation: https://docs.ecotone.tech/modelling/asynchronous-handling/dynamic-message-channels

PR #363, opened 2024-08-22
Performance: logging without message gateway

jlabedo:
## Description

Make `LoggingGateway` a registered service instead of a `Gateway` for performance: a logger should not have impact on performance :)

The `TracerInterceptor` from opentelemetry module will not intercept logs anymore:
1. it is strange to trace logs :)
2. converting logs to trace events is typically managed at the logger level for the whole application, including outside ecotone, so all logs can be converted to events (that is what I do in my application). It should not be handled by Ecotone.

## Type of change

- performance

### DIFF
--- a/packages/Ecotone/src/Messaging/Config/ModuleClassList.php
+++ b/packages/Ecotone/src/Messaging/Config/ModuleClassList.php
@@ -92,8 +92,6 @@ class ModuleClassList
         DynamicMessageChannelModule::class,
 
         /** Attribute based configurations */
-        LoggingGateway::class,
-        LoggingService::class,
         MessageHeadersPropagatorInterceptor::class,
         MessageHandlerLogger::class,
     ];
--- a/packages/Ecotone/src/Messaging/Handler/Logger/Config/LoggingModule.php
+++ b/packages/Ecotone/src/Messaging/Handler/Logger/Config/LoggingModule.php
@@ -19,6 +19,7 @@ use Ecotone\Messaging\Handler\InterfaceToCallRegistry;
 use Ecotone\Messaging\Handler\Logger\Annotation\LogAfter;
 use Ecotone\Messaging\Handler\Logger\Annotation\LogBefore;
 use Ecotone\Messaging\Handler\Logger\Annotation\LogError;
+use Ecotone\Messaging\Handler\Logger\LoggingGateway;
 use Ecotone\Messaging\Handler\Logger\LoggingHandlerBuilder;
 use Ecotone\Messaging\Handler\Logger\LoggingInterceptor;
 use Ecotone\Messaging\Handler\Logger\LoggingService;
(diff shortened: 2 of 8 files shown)


dgafka:
> converting logs to trace events is typically managed at the logger level for the whole application, including outside ecotone, so all logs can be converted to events (that is what I do in my application). It should not be handled by Ecotone.

It may be strange, but that's actually power of using Messaging. From low level what we do, is just to intercept of sending an Message (which ends up as a log) :) 
Anyways, let's keep this behaviour because it makes meaningful stamps on the traces which helps in understand the flows, and besides that it also mark given traces as failed and give indicator what part of the flow had an issue.


I can propose alternative solution to avoid using Gateway. 
We can indeed use registered service, however this Service would be just an step. 

So we can register new Service e.g. `"Logger"`, which would add all the Headers, which are now done by `Gateway`.
And `"Logger"` would have an output channel with would be `SubscribableChannel`.
Then we would connect `Tracer` and `LoggingService` to this output channel. 

image

That should do a trick without the need for Gateway :)

jlabedo:
> Anyways, let's keep this behaviour because it makes meaningful stamps on the traces which helps in understand the flows, and besides that it also mark given traces as failed and give indicator what part of the flow had an issue.

If you follow the full [opentelemetry log integration](https://opentelemetry.io/docs/languages/php/instrumentation/#integrations-for-third-party-logging-libraries), you will end up with double events in your spans: one added by ecotone, the other by the userland logger.

I you really want to do it in Ecotone, we could use the opentelemetry Logger api to decorate the logger service.

> It may be strange, but that's actually power of using Messaging. From low level what we do, is just to intercept of sending an Message (which ends up as a log) :)

You know you made me love messaging ;) It empowers connecting low coupled components, with the downside of an harder to follow flow. In case of opentelemetry, high cohesion with the logger seems better than low coupling to me. Opentelemetry specs encourage using the [Events Api](https://opentelemetry.io/docs/specs/otel/logs/event-api/) for shared libraries.

dgafka:
> one added by ecotone, the other by the userland logger.

I believe that won't happen. In general loggers are not used to add traces, as otherwise stamps would be just decorated with huge amount of not needed stuff from different libraries. 
So in case of user land code, it will be rather specific trigger to provide an stamp. 

In case of Ecotone however, those logs are meaningful and we don't publish any possible stuff there. 
If that would become a problem thought, we can just stamp only higher level internal logs.

jlabedo:
> > one added by ecotone, the other by the userland logger.
> 
> I believe that won't happen. In general loggers are not used to add traces, as otherwise stamps would be just decorated with huge amount of not needed stuff from different libraries. So in case of user land code, it will be rather specific trigger to provide an stamp.

You might be right, I am not sure anymore, I thought the logger will add an event to the current span (not a trace).

dgafka:
> You might be right, I am not sure anymore, I thought the logger will add an event to the current span (not a trace).

Ye to be exact that will be span :)

jlabedo:
@dgafka , Just to let you know, i am trying to improve projection reset performance, as it becomes unusable with a quite small db (more than 5 minutes to reset a single projection ending in 4000 rows, maybe 20 to 30 events projected per row).
I have found a pretty good candidate in PR #368. The next one is... `LoggingGateway`! It accounts for 42% of time spent for a small projection reset (no event serialization, in memory projection)

Capture d’écran 2024-08-24 à 19 35 37

jlabedo:
Closing in favor of #369 which keeps adding span events on logs.

PR #365, opened 2024-08-23
initializing aggregate in tests won't publish its events

### LINKED ISSUE #364: Initialize ES aggregates without side-effects
Don't know if this is more of a bug rather than a feature request.

Initializing ES aggregate generates side-effects which may be actual tested functionality. Consider following example:

```php
<?php

use Ecotone\Lite\EcotoneLite;
use Ecotone\Modelling\Attribute\CommandHandler;
use Ecotone\Modelling\Attribute\EventHandler;
use Ecotone\Modelling\Attribute\QueryHandler;
use Ecotone\Modelling\Attribute\EventSourcingAggregate;
use Ecotone\Modelling\Attribute\EventSourcingHandler;
use Ecotone\Modelling\CommandBus;
use Ecotone\Modelling\QueryBus;
use PHPUnit\Framework\TestCase;

#[EventSourcingAggregate]
class Aggregate {

    private array $orders = [];

    #[CommandHandler]
    public function placeOrder(PlaceOrder $command): array
    {
        // some heavy business process

        return [new OrderWasPlaced($command->aggregateId, $command->orderId)];
    }

    #[EventSourcingHandler]
    public function applyOrderWasPlaced(OrderWasPlaced $event): void
    {
        $this->orders[] = $event->order();
    }

    #[QueryHandler('orders')]
    public function orders(): array
    {
        return $this->orders;
    }
}

#[EventSourcingAggregate]
class AnotherAggregate {

    private bool $done = false;

    #[CommandHandler]
    public function placeOrder(DoStuff $command): array
    {
        if ($this->done) {
            return [];
        }

        return [new StuffWasDone($command->orderId)];
    }

    #[EventSourcingHandler]
    public function applyStuffWasDone(StuffWasDone $event): void
    {
        $this->done = true;
    }
}

class Service {

    #[EventHandler]
    public function whenOrderWasPlaced(OrderWasPlaced $event, QueryBus $queryBus, CommandBus $commandBus): void
    {
        $orders = $queryBus->sendWithRouting('orders', metadata: ['aggregate.id' => $event->aggregateId]);
        foreach ($orders as $order) {
            $commandBus->send(new DoStuff($order->orderId));
        }
    }
}

readonly class OrderWasPlaced {
    public function __construct(
        public string $aggregateId,
        public string $orderId
    ) {
    }
}

readonly class DoStuff {
    public function __construct(public string $orderId) {}
}

readonly class StuffWasDone {
    public function __construct(public string $orderId) {}
}

class ServiceTest extends TestCase
{

    public function test_service(): void
    {
        $ecotone = EcotoneLite::bootstrapFlowTestingWithEventStore(
            classesToResolve: [Aggregate::class, Service::class],
            containerOrAvailableServices: [new Service()]
        );

        $aggregateId = 'aggregate-1';
        $orderId = 'order-1';

        $ecotone->withEventsFor($aggregateId, Aggregate::class, [new OrderWasPlaced($aggregateId, $orderId)]);

        // at this point OrderWasPlaced was already published triggering Service::whenOrderWasPlaced, but we don't know that due FlowTestSupport::discardRecordedMessages was also called

        $ecotone->publishEvent(new OrderWasPlaced($aggregateId, $orderId));

        self::assertEquals([new DoStuff($orderId)], $ecotone->getRecordedCommands()); // this actually happened twice
        self::assertEquals([new StuffWasDone($orderId)], $ecotone->getRecordedEvents()); // this will fail
    }
}
```

Currently there is no way to properly check what actually happened because side-effect we would like to test against already happened when `Aggregate` was initialized. The only way to verify business behavior would be to expose value of `AnotherAggregate::$done` which seems a bit odd only for test purpose.

Being able to initialize aggregate without additional side-effect will increase tests readability. It may also introduce testing concurrency in scenario when `Aggregate` is initialized with multiple events, publishing other than the last one may have different outcome based on business rules.

unixslayer:
## Description

In tests, aggregates should be initialized without publishing its events. This helps to initialize complex aggregates without unwanted side-effects which may be subject for testing.

Fixes #364 

## Type of change

- New feature (can introduce breaking change for tests that rely on aggregate initialization side effects)
- This change may require a documentation update which will be propose separately

### DIFF
--- a/packages/Ecotone/src/Lite/Test/FlowTestSupport.php
+++ b/packages/Ecotone/src/Lite/Test/FlowTestSupport.php
@@ -179,10 +179,10 @@ final class FlowTestSupport
                 AggregateMessage::RESULT_AGGREGATE_OBJECT => $aggregateClass,
                 AggregateMessage::RESULT_AGGREGATE_EVENTS => $events,
             ],
-            ModellingHandlerModule::getRegisterAggregateSaveRepositoryInputChannel($aggregateClass)
+            ModellingHandlerModule::getRegisterAggregateSaveRepositoryInputChannel($aggregateClass). '.test_setup_state'
         );
 
-        return $this->discardRecordedMessages();
+        return $this;
     }
 
     public function withStateFor(object $aggregate): self
@@ -192,7 +192,7 @@ final class FlowTestSupport
             [
                 AggregateMessage::RESULT_AGGREGATE_OBJECT => $aggregate,
             ],
-            ModellingHandlerModule::getRegisterAggregateSaveRepositoryInputChannel($aggregate::class)
+            ModellingHandlerModule::getRegisterAggregateSaveRepositoryInputChannel($aggregate::class). '.test_setup_state'
         );
 
         return $this;
--- a/packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
+++ b/packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
@@ -727,19 +727,28 @@ class ModellingHandlerModule implements AnnotationModule
         /** @TODO do not require method name in save service */
         $methodName = $aggregateClassDefinition->getPublicMethodNames() ? $aggregateClassDefinition->getPublicMethodNames()[0] : '__construct';
 
-        $configuration->registerMessageHandler(
-            $chainMessageHandlerBuilder
-                ->chain(ResolveAggregateEventsServiceBuilder::create($aggregateClassDefinition, $methodName, $interfaceToCallRegistry))
-                ->chain(
-                    SaveAggregateServiceBuilder::create(
-                        $aggregateClassDefinition,
-                        $methodName,
-                        $interfaceToCallRegistry,
-                        $baseEventSourcingConfiguration
-                    )
-                        ->withAggregateRepositoryFactories($this->aggregateRepositoryReferenceNames)
+        $saveAggregateBuilder = $chainMessageHandlerBuilder
+            ->chain(ResolveAggregateEventsServiceBuilder::create($aggregateClassDefinition, $methodName, $interfaceToCallRegistry))
+            ->chain(
+                SaveAggregateServiceBuilder::create(
+                    $aggregateClassDefinition,
+                    $methodName,
+                    $interfaceToCallRegistry,
+                    $baseEventSourcingConfiguration
                 )
-                ->chain(PublishAggregateEventsServiceBuilder::create($aggregateClassDefinition, $methodName))
-        );
+                    ->withAggregateRepositoryFactories($this->aggregateRepositoryReferenceNames)
+            )
+        ;
+
+        if ($configuration->isRunningForTest()) {
+            $saveAggregateBuilderWithTestState = clone $saveAggregateBuilder;
+            $configuration->registerMessageHandler(
+                $saveAggregateBuilderWithTestState
+                    ->withInputChannelName($saveAggregateBuilderWithTestState->getInputMessageChannelName() . '.test_setup_state')
+            );
+        }
+
+        $saveAggregateBuilder = $saveAggregateBuilder->chain(PublishAggregateEventsServiceBuilder::create($aggregateClassDefinition, $methodName));
+        $configuration->registerMessageHandler($saveAggregateBuilder);
     }
 }
--- a/packages/Ecotone/src/Messaging/Config/Configuration.php
+++ b/packages/Ecotone/src/Messaging/Config/Configuration.php
@@ -142,4 +142,6 @@ interface Configuration extends CompilerPass
     public function registerServiceAlias(string|Reference $id, Reference $aliasTo): Configuration;
 
     public function isRunningForEnterpriseLicence(): bool;
+
+    public function isRunningForTest(): bool;
 }
--- a/packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
+++ b/packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Ecotone\Messaging\Config;
 
+use Ecotone\Lite\Test\TestConfiguration;
 use function array_map;
 
 use Ecotone\AnnotationFinder\AnnotationFinder;
@@ -156,6 +157,8 @@ final class MessagingSystemConfiguration implements Configuration
 
     private bool $isRunningForEnterpriseLicence;
 
+    private bool $isRunningForTest = false;
+
     /**
      * @param object[] $extensionObjects
      */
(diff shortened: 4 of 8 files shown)


packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
```diff
@@ -748,19 +748,33 @@ private function registerSaveAggregate(ClassDefinition $aggregateClassDefinition
...
+                    ->withAggregateRepositoryFactories($this->aggregateRepositoryReferenceNames)
+            )
+        ;
+
+
+
+        if ($configuration->isRunningForTest()) {
+            $configuration->registerMessageHandler(
+                $chainMessageHandlerBuilder
+                    ->withInputChannelName($inputChannelName . '.quiet')
```
dgafka, line 769: What is the intention here, as we do register it with normal input channel a bit lower?

dgafka:
I think events should not be published during setting up state by default. 
I don't see scenario where we would like those side effects? 
Therefore we can make it default, non-configurable behaviour.

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
unixslayer: the main intention here was to register two message handlers: standard one with publishing and quiet one without publishing events. I didn't want to change default behavior but as we agree that by default setting up aggregate in tests should not cause side effects, this will be changed anyway.

packages/Ecotone/src/Modelling/Config/ModellingHandlerModule.php
```diff
@@ -727,19 +727,23 @@ private function registerSaveAggregate(ClassDefinition $aggregateClassDefinition
...
+                    $baseEventSourcingConfiguration
                 )
-                ->chain(PublishAggregateEventsServiceBuilder::create($aggregateClassDefinition, $methodName))
-        );
+                    ->withAggregateRepositoryFactories($this->aggregateRepositoryReferenceNames)
+            )
+        ;
+
+        if (! $configuration->isRunningForTest()) {
+            $chainMessageHandlerBuilder = $chainMessageHandlerBuilder->chain(PublishAggregateEventsServiceBuilder::create($aggregateClassDefinition, $methodName));
```
dgafka, line 744: Hmm, I think that will apply outside of `withState`? 
What we could do here is to do not publishing default for `withState`, however normal behaviour we would still publish events. This means someone could set up an State, and then send an Command and verify if Event triggered an Event Handler.

What I can propose to depend on existing behaviour, yet to register a copy of existing Handler yet without a publisher e.g.

image

And then `withStateFor` and `withEventsFor` would target the additional one
image

unixslayer: That is correct. I did that previously using the same builder instead of cloning.

packages/Ecotone/src/Lite/Test/FlowTestSupport.php
```diff
@@ -179,7 +179,7 @@ public function withEventsFor(string|object|array $identifiers, string $aggregat
                 AggregateMessage::RESULT_AGGREGATE_OBJECT => $aggregateClass,
                 AggregateMessage::RESULT_AGGREGATE_EVENTS => $events,
             ],
-            ModellingHandlerModule::getRegisterAggregateSaveRepositoryInputChannel($aggregateClass)
+            ModellingHandlerModule::getRegisterAggregateSaveRepositoryInputChannel($aggregateClass). '.test_setup_state'
         );
 
         return $this->discardRecordedMessages();
```
dgafka, line 185: I guess that's not needed anymore, as nothing published?

dgafka (approved):

lifinsky:
@dgafka Can you update documentation? https://docs.ecotone.tech/modelling/testing-support/testing-event-sourcing-applications

dgafka:
@lifinsky what would like to update? 
This behaviour is expected and rather assumed :) 

If you've something in mind, feel free to open [PR to the docs](https://github.com/ecotoneframework/documentation) :)

lifinsky:
Can someone explain how to now initiate the projection state with events so that a separate factory is not created for the projection? Previously, using withEventsFor, we initialized not only the state of the aggregate, but also projections during integration tests involving, for example, commands. Now we need to additionally call publishEvent. This is inconvenient.

lifinsky:
My developers are now forced to add publishEvent to tests, because tests fail after the update. Additional publishEvent calls do not make the test code more concise. Perhaps it makes sense to add an argument to withEventsFor to force publishing of events?

unixslayer:
@lifinsky side effects on setup won't be supported due those should not happen on setup. It seems like in your case projection is still part of a setup. It will be triggered normally by events recorded and published after aggregate handles a message but in setup it should be possible to trigger it directly without publishing an event after calling withEventsFor method. I'll look into it ASAP if this is not working.

lifinsky:
@unixslayer I understand that it is more correct to have a factory to fill projections into the expected state without playing all the events in the event store, but this also makes integration tests a bit more complicated.

dgafka:
> My developers are now forced to add publishEvent to tests, because tests fail after the update. Additional publishEvent calls do not make the test code more concise. Perhaps it makes sense to add an argument to withEventsFor to force publishing of events?

So do you use test your projections in this way:

```php
// pseudo code

$orders = EcotoneLite::bootstrapFlowTestingWithEventStore([OrderProjection::class])
->withEventsFor([
    new OrderPlaced,
    new OrderCompleted
])
->sendQuery('getOrders')

// assert orders
```

because trigger projection would work here

```php
$orders = EcotoneLite::bootstrapFlowTestingWithEventStore([OrderProjection::class])
->withEventsFor([
    new OrderPlaced,
    new OrderCompleted
])
->triggerProjection('order')
->sendQuery('getOrders')
```

If that's the case, then yea I understand that it adds unnecessary step to the testing. 

> It will be triggered normally by events recorded and published after aggregate handles a message but in setup it should be possible to trigger it directly without publishing an event after calling withEventsFor method. I'll look into it ASAP if this is not working.

I am not sure if we are able to conditionally trigger only Projection and not publish Event to other Handlers. At least I would not expect this to be done in easy way.

We can add what you've been proposing at the beginning, so making this logic conditional.
So the default behaviour would state as it's, and then if someone wants to publish those events `withEventsFor(publishEvents: true)`. If set on true, we send it to original channel, otherwise to the one for test (suffixed with `.test_setup_state'`)

unixslayer:
Major point of this change was to make testing integration more predictable. Direct trigger of integration that is a subject of test should not be considered as redundant. IMO with projection being a test subject, additional step of triggering it in test is not unnecessary but makes test more readable. 

TBH if we are tanking about triggering projection on setup, this

```php
EcotoneLite::bootstrapFlowTestingWithEventStore(...)
->withEventsFor(
    events: [
        new OrderPlaced,
        new OrderCompleted
    ],
    publishEvents: true
)
```

is no different from this

```php
EcotoneLite::bootstrapFlowTestingWithEventStore(...)
->withEventsFor(
    events: [
        new OrderPlaced,
        new OrderCompleted
    ]
)
->triggerProjection('projection')
```

but later is more clear to read and understand the context.

dgafka:
Hmm true, either way we need to state `publishEvents: true`, or add `triggerProjection`, so one more step will be needed either way. 

@lifinsky are you fine to follow then triggerProjection which is already built-in function?

unixslayer:
Triggering projection is already there ;)

lifinsky:
But we can use multiple projections and link to their names in the test.
publishEvents: true can be used in a number of cases. Not critical.

dgafka:
> But we can use multiple projections and link to their names in the test.

I guess `triggerProjection` could take array of projection names to avoid chaining those in the test then :)

unixslayer:
@lifinsky would you consider this as an option?

https://github.com/ecotoneframework/ecotone-dev/pull/372

PR #367, opened 2024-08-23
Register SymfonyConfigurationVariableService as a service

### LINKED ISSUE #366: When using the `ConfigurationVariable` feature in a symfony project with ecotone, the cache warmup fails
**Ecotone version(s) affected**: 1.227.0

**Description**  
When using the `ConfigurationVariable` feature
```
        #[ConfigurationVariable("app.foobar")] int $foobarParameter
```
in a symfony project with ecotone, the cache warmup fails with
```
The service "channel-f8d88a3d-52c1-4d99-9c58-b5cf6d2f6953.target" has a dependency on a non  
  -existent service "Ecotone\Messaging\ConfigurationVariableService".     
```

**How to reproduce**  

Run cache clear or warmup in https://github.com/tandev/ecotone-symfony-config-variables 

**Possible Solution**  
Registering the `SymfonyConfigurationVariableService` as a service and switching also the container used to the `Psr\Container\ContainerInterface` fixes it. But i dont know if there is supposed to be another way to register it.

I made a commit which would fix this in the way i know:
https://github.com/ecotoneframework/ecotone-dev/commit/1f653e5ff2ce9e0a5048ea2c08e3b4dc5d056e1d

tandev:
## Description

When using the ConfigurationVariable feature

        #[ConfigurationVariable("app.foobar")] int $foobarParameter

in a symfony project with ecotone, the cache warmup fails with

The service "channel-f8d88a3d-52c1-4d99-9c58-b5cf6d2f6953.target" has a dependency on a non  
  -existent service "Ecotone\Messaging\ConfigurationVariableService".     

Registering the SymfonyConfigurationVariableService as a service and switching also the container used to the Psr\Container\ContainerInterface fixes it. But i dont know if there is supposed to be another way to register it.

Fixes #366

## Type of change


jlabedo:
Hello @tandev , thank you for your contribution, that seems nice!
Would you mind adding a test ? I think simply adding a service with `#[ConfigurationVariable("app.foobar")] int $foobarParameter` in `Symfony/tests/Fixtures` would make it.

tandev:
Could not find a a appropiate place in the fixtures, but the SingleTenant-Test had a matching place to add such behavior i think, added it there

jlabedo:
Tanks @tandev ,
I am torned.
Looking at the documentation, the `#[ConfigurationVariable]` is meant to be used in configuration classes (classes with `#[ServiceContext]` attribute). These classes are used during compilation phase, and not runtime. So in case your parameter depends on runtime (environment variable, ...), it will NOT be interpolated by Ecotone like Symfony does (at runtime). It might be useful to had a warning on this behavior if requesting an "interpolated" parameter with `#[ConfigurationVariable]` attribute (for instance `%env(PARAM)%` or `%kernel.project_dir%`)

If we merge your PR, we would end up with two different behaviors for this attribute (if using a framework that supports interpolation like Symfony does):
1. Compilation phase, like in documentation in `ServiceContext` attribute, it will inject a NON interpolated parameter `%env(PARAM)%`
2. During runtime, thanks to your PR, it will inject the interpolated parameter (`foo` if PARAM=foo)

So first question: Can you use a Symfony attribute for your case ? (`#[Autowire('app.foobar')]`)

I see 2 options:
1. Merge this PR and add documentation (and tests) about the behaviour
2. Reference runtime parameters with the `#[Reference]` attribute

dgafka:
@jlabedo I think that's fine and we can go with that, it's actually useful behaviour :) 
Let's just [update the docs](https://github.com/ecotoneframework/documentation) that during compilation phase (ServiceContext) we won't get interplolated parameters.

jlabedo (approved):

PR #368, opened 2024-08-24
Speedup event sourced projection reset

jlabedo:
## Description

Projection reset is sloooow.
This PR change the default block size before persisting projection state. It is now 1000 (this is the default value of Prooph) so it will send to db the projection state after handling 1000 events.

This change lower by more than 2 the projection reset time in my production tests.

## Type of change

- Performance

### DIFF
--- a/packages/PdoEventSourcing/src/ProjectionRunningConfiguration.php
+++ b/packages/PdoEventSourcing/src/ProjectionRunningConfiguration.php
@@ -24,7 +24,7 @@ class ProjectionRunningConfiguration implements DefinedObject
     public const DEFAULT_WAIT_BEFORE_CALLING_ES_WHEN_NO_EVENTS_FOUND = 0;
 
     public const OPTION_PERSIST_CHANGES_AFTER_AMOUNT_OF_OPERATIONS = 'persist_block_size';
-    public const DEFAULT_PERSIST_CHANGES_AFTER_AMOUNT_OF_OPERATIONS = 1;
+    public const DEFAULT_PERSIST_CHANGES_AFTER_AMOUNT_OF_OPERATIONS = 1000;
 
     public const OPTION_PROJECTION_LOCK_TIMEOUT = 'lock_timeout_ms';
     public const DEFAULT_PROJECTION_LOCK_TIMEOUT = 1000;


unixslayer (approved):

dgafka:
Are you sure this is actually safe? 

This means that projection state may be skipped, awaiting for the persisting 
image

So I would expect what happens now:

1. If projection running async, it will handle 500 events and will fail. In the database the read model will be already built for those 500, however they will be reprojected because state was not finished
2. When synchronous projection happen, it will never reach the 1000 threshold, as there is one or few events in maximum to project. Therefore the state won't be really stored. (Prooph was not build with mindset of sync projections).

jlabedo:
@dgafka , The  Prooph `persist` method is called when projection runner stop, and when the stream is finished, so the state is correctly persisted.
 
> 1. If projection running async, it will handle 500 events and will fail. In the database the read model from client side code will be already in state after 500 events, however those Events will be reprojected because state was not finished for Projection.

If your projection is in the same db, there is no problem: the projection and projector state will be included in the same transaction. If your projection is elsewhere, then 500 first events will be sent to the projection, but the transaction for projector state will rollback. It does not matter if you issued an SQL `UPDATE` for your projection state on each event or not. So I don't think this is a regression.


> 2. When synchronous projection happen, it will never reach the 1000 threshold, as there is one or few events in maximum to project. Therefore the state won't be really stored. (Prooph was not build with mindset of sync projections).

Looking at the test suite, a few test cases cover this possibility.

dgafka:
> The Prooph persist method is called when projection runner stop, and when the stream is finished, so the state is correctly persisted.

Ok if that's the case we are fine :) 
Where does the stream is finished persistence happens in the code?

jlabedo:
after stream completion [here](https://github.com/prooph/pdo-event-store/blob/118e727d9efd90ecf04d712841d67231d84c8f34/src/Projection/PdoEventStoreReadModelProjector.php#L535)

on stop: [here](https://github.com/prooph/pdo-event-store/blob/118e727d9efd90ecf04d712841d67231d84c8f34/src/Projection/PdoEventStoreReadModelProjector.php#L379)

PR #369, opened 2024-08-25
Send logs to dedicated monolog channel in symfony bundle

jlabedo:
## Description

This PR will send all logs to a dedicated `ecotone` channel when `monolog` bundle is present in symfony integration.

This feature required to refactor all the logging handling to have a central service handling internal logs. All internal components are logging through a `LoggingGateway` interface, extending `PSR\Log\LoggerInterface`, while it is still possible to inject another `LoggerInterface` as a parameter for userland code.

The refactoring permits to avoid a lot of special logger handling in messaging configuration.

Also, while still named `LoggingGateway` the service is not a messaging gateway anymore, which will dramatically improve performance.

Fixes #363 

## Type of change

- Performance


jlabedo:
That is funny :) 

Capture d’écran 2024-08-25 à 16 15 30

packages/Ecotone/src/Messaging/Handler/Logger/LoggingService.php
```diff
@@ -25,132 +20,45 @@
...
-            $data = (string)$data;
-        } elseif (! TypeDescriptor::createFromVariable($data)->isScalar()) {
-            if ($this->conversionService->canConvert($sourceTypeDescriptor, $sourceMediaType, TypeDescriptor::createStringType(), MediaType::createApplicationJson())) {
-                $data = $this->conversionService->convert($data, $sourceTypeDescriptor, $sourceMediaType, TypeDescriptor::createStringType(), MediaType::createApplicationJson());
-            } else {
-                $data = $this->conversionService->convert($data, $sourceTypeDescriptor, $sourceMediaType, TypeDescriptor::createStringType(), MediaType::createApplicationXPHPSerialized());
-            }
+        if ($logger && ! in_array($logger, $this->loggers)) {
+            $this->loggers[] = $logger;
         }
```
jlabedo, line 62: I did not use the Message Channel abstraction, but the concept is the same as what you suggested in your [comment](https://github.com/ecotoneframework/ecotone-dev/pull/363#issuecomment-2308184933)

jlabedo:
> @jlabedo can you rebase?

Done @dgafka

packages/OpenTelemetry/src/Configuration/RegisterAddSpanEventLoggerCompilerPass.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    public function process(ContainerBuilder $builder): void
+    {
+        if ($builder->has(LoggingGateway::class)) {
+            $loggingGatewayDefinition = $builder->getDefinition(LoggingGateway::class);
+
+            if (is_a($loggingGatewayDefinition->getClassName(), SubscribableLoggingGateway::class, true)) {
+                $loggingGatewayDefinition->addMethodCall("registerLogger", [new Definition(AddSpanEventLogger::class)]);
+            }
+        }
```
dgafka, line 24: Interesting idea of reusing CompilerPass for that :+1:

dgafka:
> That is funny :)
> 
> Capture d’écran 2024-08-25 à 16 15 30

Looks like Ecotone footprint on the Application will less than minimal, yet still providing all the functionality. 
Having timing lower than `1ms` for handling http call is like running raw Controller with "Hello World" in it. :)

packages/Laravel/src/EcotoneProvider.php
```diff
@@ -201,19 +195,6 @@ public function boot()
...
-                LoggingHandlerBuilder::LOGGER_REFERENCE,
-                function (Application $app) {
-                    if ($app->runningInConsole()) {
-                        return new CombinedLogger($app->get('log'), new EchoLogger());
-                    }
-
-                    return $app->get('log');
-                }
-            );
-        }
```
dgafka, line 216: I guess that could stay, in case Laravel have not registered the Logger?

packages/OpenTelemetry/src/Configuration/RegisterAddSpanEventLoggerCompilerPass.php
jlabedo: This is how it is done in symfony, so not my idea :)
Every symfony extension (ecotone module) registers services in an independant container. If your module needs to hook into other extension services, you register compiler passes that have access to the whole container.

dgafka: I meant using this pattern for solving Logging :)

packages/Laravel/src/CombinedLogger.php
```diff
@@ -1,28 +0,0 @@
...
-    {
-        $this->applicationLogger = $applicationLogger;
-        $this->consoleLogger = $consoleLogger;
-    }
-
-    public function log($level, $message, array $context = []): void
-    {
-        $this->applicationLogger->log($level, $message, $context);
-        $this->consoleLogger->log($level, $message, $context);
-    }
```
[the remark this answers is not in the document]
jlabedo: It seemed me strange that extensions had to choose the logger. It should be defined at the application level I think, but I will add a test for laravel (not before monday... )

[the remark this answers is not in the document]
dgafka: Ok sure, will be helpful if we make changes here :) 
I think Laravel could have it registered under different name, that's why I've added it.

packages/Laravel/src/EcotoneProvider.php
jlabedo: It is not necessary, it is handled at the Ecotone level (if a "logger" service exists, it is added as a logger, if it does not, it injects `null`)

packages/Laravel/src/EcotoneProvider.php
```diff
@@ -143,6 +145,9 @@ function () {
         );
 
         if ($this->app->runningInConsole()) {
+            $this->callAfterResolving(LoggingGateway::class, function (LoggingGateway $loggingGateway) {
+                $loggingGateway->registerLogger(new EchoLogger());
+            });
```
jlabedo, line 150: This is where the logging to console happens. But I am not a big fan of handling that in the provider: there is no way to disable those logs in Laravel now. I don't really how it should be handled in Laravel.

packages/Laravel/src/EcotoneProvider.php
dgafka: If you take a look on what we inject into CombinedLogger, there is an Service called `log`, not `logger`. 
What we look for to inject into Gateway is `logger` thought. 
So I would assume this will cut off logs for Laravel, as there will be no logger registered. 

If `logger` does not exists, we can rebind `log` for that, then it should work.

dgafka: #376 added tests

packages/Laravel/src/EcotoneProvider.php
dgafka: We can drop this, in favour of #376. 
It does send to `stack` logger by default, which can be defined however given App wants

dgafka (approved):

PR #370, opened 2024-08-27
Amqp self heal on acknowledge failures

dgafka:
## Description

- This provides self heal on acknowedge failures
- Recovers subscription based amqp message fetching

### DIFF
--- a/packages/Amqp/src/AmqpInboundChannelAdapter.php
+++ b/packages/Amqp/src/AmqpInboundChannelAdapter.php
@@ -79,7 +79,7 @@ class AmqpInboundChannelAdapter extends EnqueueInboundChannelAdapter
         );
     }
 
-    public function receiveMessage(int $timeout = 0): ?Message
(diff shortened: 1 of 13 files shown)


packages/Amqp/src/AmqpInboundChannelAdapter.php
```diff
@@ -65,10 +69,17 @@ public function enrichMessage(EnqueueMessage $sourceMessage, MessageBuilder $tar
...
+            new AmqpAcknowledgeCallbackWraper(
+                $targetMessage->getHeaderWithName(EnqueueHeader::HEADER_ACKNOWLEDGE),
+                $this->cachedConnectionFactory,
+                $this->loggingGateway,
+            )
+        );
     }
 
-    public function receiveMessage(int $timeout = 0): ?Message
+    public function receiveWithTimeout(int $timeout = 0): ?Message
```
dgafka, line 82: That was not overriding base method.

PR #372, opened 2024-08-31
allow triggering multiple projections with one call

unixslayer:


dgafka (approved):

packages/Ecotone/src/Lite/Test/FlowTestSupport.php
```diff
@@ -198,9 +198,11 @@ public function withStateFor(object $aggregate): self
         return $this;
     }
 
-    public function triggerProjection(string $projectionName): self
+    public function triggerProjection(string ...$projectionName): self
```
dgafka, line 201: Maybe `string|array`? So in case we would need to add some configuration parameter it won't be a B/C

dgafka (approved):

PR #373, opened 2024-08-31
Improved recoverability and logging

dgafka:
## Description

This improves recoverabiility for asynchronous scenarios and logging (to provide more info what's happening with Message when it fails)


packages/Dbal/src/Recoverability/DbalDeadLetterHandler.php
```diff
@@ -191,7 +195,19 @@ public function store(Message $message): void
...
 
-        $this->insertHandledMessage($message->getPayload(), $message->getHeaders()->headers());
+        $retryStrategy = RetryTemplateBuilder::exponentialBackoffWithMaxDelay(10, 3, 1000)
+            ->maxRetryAttempts(3)
+            ->build();
+
+        $retryStrategy->runCallbackWithRetries(function () use ($message) {
+            try {
+                $this->insertHandledMessage($message->getPayload(), $message->getHeaders()->headers());
+            }catch (\Exception $exception) {
```
dgafka, line 205: retrying on failure to store in dead letter. If we will manage to store it, there is no need to nack the message in the channel

packages/Dbal/src/DbalTransaction/DbalTransactionInterceptor.php
```diff
@@ -118,13 +118,11 @@ public function transactional(MethodInvocation $methodInvocation, Message $messa
...
+                    /** Doctrine hold the state, so it needs to be cleaned */
                     $connection->rollBack();
-                } catch (Throwable $rollBackException) {
-                    $logger->info(
-                        'Exception has been thrown, however could not rollback the transaction.',
-                        $message,
-                        $exception
-                    );
+                } catch (Exception) {
+                    $connection->close();
```
dgafka, line 125: If we failed on rollback, it most likely means the connection is already dead, therefore we can close right away.

packages/Ecotone/src/Messaging/Endpoint/AcknowledgeConfirmationInterceptor.php
```diff
@@ -42,68 +43,103 @@ public static function createAroundInterceptorBuilder(InterfaceToCallRegistry $i
...
-                $message
-            );
             throw $exception;
         }
 
-        return $result;
-    }
+        if ($amqpAcknowledgementCallback->isAutoAck()) {
+            $retryStrategy->runCallbackWithRetries(function () use ($message, $methodInvocation, $logger, $messageChannelName, $amqpAcknowledgementCallback) {
+                $amqpAcknowledgementCallback->accept();
```
dgafka, line 136: We can try to ack with retries, under the hood connection will disconnected when acking fails. 
This will help in scenarios like: dbal, redis, sqs channels, however won't help in recovering from AMQP (as same connection is needed for acking). 
In case of AMQP, message will be requeued and rehandled again.

PR #381, opened 2024-10-07
fix gateways fqcn in container definition

jlabedo:
## Description

Gateway proxies are registered without their namespace in containers (`Ecotone_Modelling_CommandBus` instead of `Ecotone\__Proxy__\Ecotone_Modelling_CommandBus`). It actually works because the service is built from the `ProxyFactory`, but it can have weird side effects, for instance false alerts from phpstan using the symfony extension.

## Type of change

### DIFF
--- a/Monorepo/CrossModuleTests/Tests/FullAppTestCase.php
+++ b/Monorepo/CrossModuleTests/Tests/FullAppTestCase.php
@@ -12,9 +12,14 @@ abstract class FullAppTestCase extends FullAppBenchmarkCase
         $this->bench_symfony_prod();
     }
 
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
     public function test_symfony_dev()
     {
         self::clearSymfonyCache();
+        \ini_set("error_log", "/dev/null");
         $this->bench_symfony_dev();
     }
 
--- a/packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
+++ b/packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
@@ -35,7 +35,7 @@ class ProxyFactory
 
     public static function getGatewayProxyDefinitionFor(GatewayProxyReference $proxyReference): Definition
     {
-        return new Definition(self::getClassNameFor($proxyReference), [
+        return new Definition(self::getFullClassNameFor($proxyReference), [
             new Definition(GatewayProxyReference::class, [
                 $proxyReference->getReferenceName(),
                 $proxyReference->getInterfaceName(),


unixslayer (dismissed):

dgafka:
Do you think it could contribute Manuel's issue?
image
image

jlabedo:
Hello @dgafka , I don't think the issues are related. But the error message from Manuel Ortega is quite strange, so it MAY be related :)

Looking at the failing test, this is due to a feature from symfony called `inline_class_loader`, which is activated by default when `debug` is true. This option increases performance (a lot: see https://twitter.com/nicolasgrekas/status/929032213815005184) by bypassing the autoloader and is documented here: https://github.com/symfony/symfony/pull/26276/files

What happen is that symfony will add `include_once`s statement in the dumped container (only in `debug`mode by default) and it takes the file path from reflection. As we are running the `prod` test first, the loaded proxy class is under /var/cache/**prod**/ecotone and this filepath is dumped in the **dev** container, but the proxy file does not exist anymore.

For instance, in the symfony **dev** container, here is the dumped code:
```php
// var/cache/dev/ContainerYCSXjss/getCommandBusService.php

<?php

namespace ContainerYCSXjss;

use Symfony\Component\DependencyInjection\Argument\RewindableGenerator;
use Symfony\Component\DependencyInjection\ContainerInterface;
use Symfony\Component\DependencyInjection\Exception\RuntimeException;

/**
 * @internal This class has been auto-generated by the Symfony Dependency Injection Component.
 */
class getCommandBusService extends Monorepo_ExampleApp_Symfony_KernelDevDebugContainer
{
    /**
     * Gets the public 'Ecotone\Modelling\CommandBus' shared service.
     *
     * @return \Ecotone\__Proxy__\Ecotone_Modelling_CommandBus
     */
    public static function do($container, $lazyLoad = true)
    {
        include_once \dirname(__DIR__, 2).'/prod/ecotone/Ecotone_Modelling_CommandBus.php';

        return $container->services['Ecotone\\Modelling\\CommandBus'] = ($container->services['Ecotone\\Messaging\\Handler\\Gateway\\ProxyFactory'] ?? $container->load('getProxyFactoryService'))->createProxyInstance(new \Ecotone\Messaging\Config\Container\GatewayProxyReference('Ecotone\\Modelling\\CommandBus', 'Ecotone\\Modelling\\CommandBus'), ($container->services['Ecotone\\Messaging\\Config\\ConfiguredMessagingSystem'] ?? $container->load('getConfiguredMessagingSystemService')));
    }
}
```

Possible solutions:
1. customize proxy namespace for each symfony environment to avoid conflicts
2. run tests in separate processes
3. disable `inline_class_loader` for those tests

I would go for **option 2**. WDYT ?

dgafka:
@jlabedo ok thanks for explanation.

Ye it make sense to run them in separate processes then :)

unixslayer (approved):

lifinsky:
@jlabedo 
We got error when use bin/phpunit in Symfony peojects:

`Fatal error: Cannot declare class Ecotone\__Proxy__\Ecotone_Modelling_CommandBus, because the name is already in use in /app/var/cache/test/ecotone/Ecotone_Modelling_CommandBus.php on line 5`

 ```
class getCommandBusService extends Fcards_App_KernelTestDebugContainer
 {
     /**
      * Gets the public 'Ecotone\Modelling\CommandBus' shared service.
      *
      * @return \Ecotone\__Proxy__\Ecotone_Modelling_CommandBus
      */
     public static function do($container, $lazyLoad = true)
     {
         include_once \dirname(__DIR__, 4).'/vendor/ecotone/ecotone/src/Modelling/CommandBus.php';
         include_once $container->targetDir.''.'/ecotone/Ecotone_Modelling_CommandBus.php';
         include_once \dirname(__DIR__, 4).'/vendor/ecotone/ecotone/src/Messaging/Config/Container/Reference.php';
         include_once \dirname(__DIR__, 4).'/vendor/ecotone/ecotone/src/Messaging/Config/Container/GatewayProxyReference.php';
 
         return $container->services['Ecotone\\Modelling\\CommandBus'] = ($container->services['Ecotone\\Messaging\\Handler\\Gateway\\ProxyFactory'] ?? $container->load('getProxyFactoryService'))->createProxyInstance(new \Ecotone\Messaging\Config\Container\GatewayProxyReference('Ecotone\\Modelling\\CommandBus', 'Ecotone\\Modelling\\CommandBus'), ($container->services['Ecotone\\Messaging\\Config\\ConfiguredMessagingSystem'] ?? $container->load('getConfiguredMessagingSystemService')));
     }
 }
```

jlabedo:
@lifinsky , are you using multiple kernels in your tests ?

lifinsky:
@jlabedo Application tests extend Symfony\Bundle\FrameworkBundle\Test\WebTestCase
Integration - Symfony\Bundle\FrameworkBundle\Test\KernelTestCase

lifinsky:
We use getContainer for Ecotone Lite application in application tests. No errors when run vendor/bin/phpunit instead of bin/phpunit

jlabedo:
I think your setup is generating new kernel cache, and proxies conflict with each others. I will prepare a PR for suffixing proxy classes

jlabedo:
In the meantime, you can disable symfony debug for test environment (see the reason [here](https://github.com/ecotoneframework/ecotone-dev/pull/381#issuecomment-2397238684)). I see better performance with my test suite. Or you can disable `inline_class_loader`, but maybe you will have lower speed for your tests

lifinsky:
Thank you, let's try to disable debug

PR #383, opened 2024-10-11
Load classes from autoloader classmap if authoritative

### LINKED ISSUE #382: ClassNotFoundError thrown when excluding files from composer classmap
**Ecotone version(s) affected**: 1.231.1

**Description**  
I have a directory structure where some tests are located in the same namespace as production code, like Symfony does in its monorepo (for instance [this test](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Serializer/Tests/DeserializeNestedArrayOfObjectsTest.php)).
When building without composer dev packages (`--no-dev`), Ecotone will parse the test file and throw a ClassNotFound error because of missing PHPUnit required by the test file.

**How to reproduce**  
Include a test case inside the `src` directory, run `composer install --no-dev` with a **valid** composer.json file:
```json
    "autoload": {
        "psr-4": {
            "App\\": "src/"
        },
        "exclude-from-classmap": [
            "**/Tests/",
            "**/bin/"
        ]
    },
```

**Possible Solution**  
- Ignore unreadable classes when parsing annotations.
- Read "exclude-from-classmap" to exclude it from annotations finder

jlabedo:
## Description

File system annotation finder will get classmap from autoloader if it is authoritative and namespaces to load are provided.
It has no impact if parsing is made with a catalog folder.

Fixes #382 

## Type of change


quickstart-examples/AdvancedNamespaceParser/run_example_authoritative.php
```diff
@@ -0,0 +1,19 @@
...
+require __DIR__ . "/vendor/autoload.php";
+$messagingSystem = EcotoneLiteApplication::bootstrap(
+    pathToRootCatalog: __DIR__,
+    serviceConfiguration: ServiceConfiguration::createWithDefaults()
+        ->withNamespaces(["App\\"]),
+);
+
+$messagingSystem->getEventBus()->publish(new CustomerRegistered(1));
+
+echo "Customer registered\n";
```
dgafka, line 19: Needs to be added to `quickstart-examples/composer.json`  to be part of CI process, with `--no-dev install` to confirm it does skip it.

jlabedo: @dgafka, I added 3 tests:
- `run_example_authoritative.php` will discard `Tests` folder
- `run_example_non_authoritative.php` will try to parse `Tests` folder but throws an exception because of missing PHPUnit (--no-dev)
- `run_example_non_authoritative_with_dev.php` will parse and run `Tests` folder

dgafka (approved):

PR #385, opened 2024-10-13
Allows to enable Ecotone Enterprise

dgafka:
## Description

This allows to enable Ecotone Enterprise in Symfony, Laravel, Ecotone Lite.

## Type of change


lifinsky:
@dgafka it would be great to set the service name and license key using env variables instead of the static yaml config for Symfony. But maybe a better practice would be to use env variables for all Ecotone configuration params and remove the config file.

dgafka:
@lifinsky true, will take a look on this :)

PR #391, opened 2024-10-17
Proxy generation fix

dgafka:
## Description

For some reason evaluating PHP class at run time ends up with errors for Laravel. It acts like it would be looking for a file, when there are none. Therefore it ends up failing.

Therefore instead of relying on in fly evaluation, we can dump the file and depend on this.

### loading Ecotone
Screenshot from 2024-10-17 20-33-56

### adding dump
Screenshot from 2024-10-17 20-35-27

### result
Screenshot from 2024-10-17 20-35-32

## Type of change

### DIFF
--- a/packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
+++ b/packages/Ecotone/src/Messaging/Handler/Gateway/ProxyFactory.php
@@ -64,14 +64,10 @@ class ProxyFactory
     private function loadProxyClass(GatewayProxyReference $proxyReference): string
     {
         if (! self::isLoaded($proxyReference)) {
-            if ($this->serviceCacheConfiguration->shouldUseCache()) {
-                $file = $this->generateCachedProxyFileFor($proxyReference, false);
-                require $file;
-            } else {
-                $code = $this->generateProxyCode($proxyReference);
-                eval($code);
-            }
+            $file = $this->generateCachedProxyFileFor($proxyReference, !$this->serviceCacheConfiguration->shouldUseCache());
+            require $file;
         }
+        
         return self::getFullClassNameFor($proxyReference);
     }
 


dgafka:
@jlabedo I guess it may be a bit slower for dev environments, however we will keep the same behaviour between prod and dev run. 

@lifinsky can you check if it solves [your issue](https://github.com/ecotoneframework/ecotone-dev/pull/381#issuecomment-2417797230) too?

jlabedo:
I am curious @dgafka , how can I reproduce this error ?
I think it may be related to [inline_class_loader](https://github.com/ecotoneframework/ecotone-dev/pull/381#issuecomment-2397238684) of Symfony.
And I think the error is message is wrong, it must be something thrown by the error handler itself, so it is hiding the real error:
I guess the error handler tries to open the file of the throwing class throw to get its code, but the file does not exists (because of `eval()'d code`) and throws this error...

jlabedo:
Opened an issue on Laravel https://github.com/laravel/framework/issues/53203

dgafka:
> I am curious @dgafka , how can I reproduce this error ?

I see you already reproduced that :) 

> I guess the error handler tries to open the file of the throwing class throw to get its code, but the file does not exists (because of eval()'d code) and throws this error...

Ye most likely that's the case.
But I guess we can live without eval and depends on the production flow for dev envs too, right?
Proxy generation was pain for as long as I remember, the less logic there the better for us :)

jlabedo:
> But I guess we can live without eval and depends on the production flow for dev envs too, right?

I don't know. It seems ok to me, but it is not the behavior of most other libs working with proxies (doctrine, proxymanager). The main difference with doctrine that is causing issues, is that proxies are registered in the container by Ecotone (they are services) while it is not the case for doctrine (entities).

Let's keep it like this, but at least we could see the real error when https://github.com/laravel/framework/pull/53204 is merged into laravel

lifinsky:
@dgafka Scheduled endpoints fixed, thanks

PR #392, opened 2024-10-17
Allow to use env variables in Ecotone's Symfony Config

dgafka:
## Description

Allow to use env variables in Ecotone config for Symfony.
It will now work for Symfony, just like it does for Laravel. 
Therefore the environment variables will be resolved during container build and use for preparing Ecotone's configuration.

---
Using environment variables is safe as long as container is build with expected variables. If we will for example build the container, dump it into Docker Image, and the run it in container with different environment variables - only the envs used during build will be used. 

## Type of change

### DIFF
--- a/packages/Laravel/tests/Licence/LicenceTest.php
+++ b/packages/Laravel/tests/Licence/LicenceTest.php
@@ -6,6 +6,7 @@ namespace Test\Ecotone\Laravel\Licence;
 
 use Ecotone\Modelling\CommandBus;
 use Ecotone\Modelling\QueryBus;
+use Ecotone\Test\LicenceTesting;
 use Illuminate\Foundation\Application;
 use Illuminate\Foundation\Http\Kernel;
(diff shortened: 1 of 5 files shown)


dgafka:
cc: @lifinsky

lifinsky:
I once also wanted to make such an PR, but I didn’t like that it wouldn’t be possible to work with bool or int parameters in the config.

dgafka:
@lifinsky this what we can achieve with reasonable amount of time. 
In general that would require refactor to make ServiceConfiguration an Service in DI, then we could inject any config to that at run time. Yet it means compile phase would need to change and not use it at all. Yet we do use it in order to make performance smoother as most of the configuration is basically done and ready to execute (no need for run time reconfigurations)

So at this moment, that's fair enough trade off, which will work for particular cases.

lifinsky:
@dgafka #399

Issue #386, opened 2024-10-15
The distributed consumer in application tests stopped working correctly


lifinsky:
**Ecotone version(s) affected**: 1.229.0-latest

**Description**  
We have tests that send a distributed command and check its processing with a delayed retry. After an update, we noticed that tests pass individually, but when running two or more, only one test passes. The remaining tests do not receive the message; it stays in the queue.

**How to reproduce**  
First test for example:
```
    public function testResilientSending(): void
    {
        $this->logger
            ->expects(self::exactly(3))
            ->method('error')
            ->with(self::isInstanceOf(AggregateNotFoundException::class))
        ;
        $this->distributedBus
            ->expects(self::once())
            ->method('convertAndPublishEvent')
            ->with(
                BlockCardHasFailed::ROUTING_KEY,
                self::isInstanceOf(BlockCardHasFailed::class),
            )
        ;

        $ecotone = $this->bootstrapEcotoneLite();
        $ecotone->getDistributedBus()->convertAndSendCommand(
            destination: self::SERVICE_NAME,
            routingKey: BlockCardCommand::ROUTING_KEY,
            command: new BlockCardCommand(
                id: new UuidV4(),
            )
        );
        $ecotone->run(
            endpointId: self::SERVICE_NAME,
            executionPollingMetadata: ExecutionPollingMetadata::createWithFinishWhenNoMessages(false),
        );
    }
```

Second test (separate test case):
```
    public function testResilientSending(): void
    {
        $this->logger
            ->expects(self::exactly(3))
            ->method('error')
            ->with(self::isInstanceOf(AggregateNotFoundException::class))
        ;
        $this->distributedBus
            ->expects(self::once())
            ->method('convertAndPublishEvent')
            ->with(
                UnblockCardHasFailed::ROUTING_KEY,
                self::isInstanceOf(UnblockCardHasFailed::class),
            )
        ;

        $ecotone = $this->bootstrapEcotoneLite();
        $ecotone->getDistributedBus()->convertAndSendCommand(
            destination: self::SERVICE_NAME,
            routingKey: UnblockCardCommand::ROUTING_KEY,
            command: new UnblockCardCommand(
                id: new UuidV4(),
            )
        );
        $ecotone->run(
            endpointId: self::SERVICE_NAME,
            executionPollingMetadata: ExecutionPollingMetadata::createWithFinishWhenNoMessages(false),
        );
    }
```
```
UnblockCardTest::testResilientSending
[...]

PR #398, opened 2024-10-23
Add test to reproduce reuse of AMQP queue between tests

lifinsky:
## Description

Test to reproduce #386


lifinsky:
@dgafka basic get fixes this issue

dgafka:
Merging this, as this will be useful when we will try to refactor this to consume to ensure, this scenario works as before :)

PR #401, opened 2024-10-26
Asynchronous extendable Gateways

dgafka:
## Description

- This makes Command/Query/Event Buses extendable using interceptors. 

image

- It also adds ability Enterprise feature to make Gateways asynchronous, which provides easy way to build things like Outbox Event Bus.
image

## Type of change


dgafka:
@lifinsky asynchronous gateways coming in :)

lifinsky:
@dgafka Do you plan to support of  multiple distributed bus interfaces as async command bus?

PR #405, opened 2024-11-02
Update dependencies

lifinsky:
## Description

- Remove PHP 8.0 support
- Prepare for PHPUnit 10 (Using PHPUnit 10 is currently blocked by Behat)
- Add Doctrine DBAL support for Laravel 11 (Due to a conflict with PHPUnit, Laravel in current tests is used with version 10, which has built-in Doctrine support, which was removed in Laravel 11)
- Update some dependencies

The deprecation warning in the DbalReconnectableConnectionFactory::reconnect() method remains unpleasant. I suggest rewriting it using Reflection:

```php
 // Doctrine\DBAL\Connection (v3.9)
    public function connect()
    {
        Deprecation::triggerIfCalledFromOutside(
            'doctrine/dbal',
            'https://github.com/doctrine/dbal/issues/4966',
            'Public access to Connection::connect() is deprecated.',
        );

        if ($this->_conn !== null) {
            return false;
        }

        try {
            $this->_conn = $this->_driver->connect($this->params);
        } catch (Driver\Exception $e) {
            throw $this->convertException($e);
        }

        if ($this->autoCommit === false) {
            $this->beginTransaction();
        }

        if ($this->_eventManager->hasListeners(Events::postConnect)) {
            Deprecation::trigger(
                'doctrine/dbal',
                'https://github.com/doctrine/dbal/issues/5784',
                'Subscribing to %s events is deprecated. Implement a middleware instead.',
                Events::postConnect,
            );

            $eventArgs = new Event\ConnectionEventArgs($this);
            $this->_eventManager->dispatchEvent(Events::postConnect, $eventArgs);
        }

        return true;
    }
```


packages/Laravel/src/Config/LaravelConnectionResolver.php
```diff
@@ -20,6 +22,33 @@ public static function resolveLaravelConnection(LaravelConnectionReference $conn
...
+
+    private static function createDriver($driverName): Driver
+    {
+        $className = match ($driverName) {
+            'pgsql' => 'PostgresDriver',
+            'mysql' => 'MySqlDriver',
+            'sqlite' => 'SqliteDriver',
+            'sqlsrv' => 'SqlServerDriver',
+        };
+        $className = '\Ecotone\Laravel\Config\PDO\\' . $className;
```
dgafka, line 50: Can we instead of reimplementing Driver classes, make use of real ones?
E.g. if dbal packages is not installed by default in Laravel 11, then check for class existance here, and if not installed tell something like "Install doctrine/dbal to make use of this"

.github/workflows/test-monorepo.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         operating-system: [ ubuntu-latest ]
-        php-versions: [ '8.0', '8.3' ]
+        php-versions: [ '8.1', '8.3' ]
```
dgafka, line 18: I know PHP 8.0 is not longer supported, but I know people are still using that. 
Is there a specific reason to drop it now?

packages/Laravel/src/Config/LaravelConnectionResolver.php
lifinsky: @dgafka I dragged the implementation of adapters from Laravel 10. These classes were in the framework itself

.github/workflows/test-monorepo.yml
lifinsky: This does not allow us to avoid deprecating and testing with the latest versions of the Laravel and Symfony frameworks. What I think is more commercially important. Upgrade from 8.0 to 8.1 simple, and current quickstart examples depend on PHP 8.2. For example readonly classes. But I think it is necessary to bump the version gradually.

packages/Laravel/src/Config/PDO/Connection.php
```diff
@@ -0,0 +1,186 @@
...
+use Doctrine\DBAL\Driver\PDO\Statement;
+use Doctrine\DBAL\Driver\Result as ResultInterface;
+use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
+use Doctrine\DBAL\Driver\Statement as StatementInterface;
+use Doctrine\DBAL\ParameterType;
+use PDO;
+use PDOException;
+use PDOStatement;
+
+class Connection implements ServerInfoAwareConnection
```
lifinsky, line 16: @dgafka Main connection wrapper for DBAL based on PDO connection

packages/Laravel/src/Config/PDO/Concerns/ConnectsToDatabase.php
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @param mixed[] $params
+     * @param string|null $username
+     * @param string|null $password
+     * @param mixed[] $driverOptions
+     * @return \Ecotone\Laravel\Config\PDO\Connection
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function connect(array $params, $username = null, $password = null, array $driverOptions = [])
```
lifinsky, line 22: https://github.com/doctrine/dbal/blob/3.9.x/src/Driver/PDO/PgSQL/Driver.php
The current drivers are final classes and they are designed to instantiate a connection, not to use existing pdo instance

@dgafka

lifinsky:
If we want to continue supporting Laravel, we will have to take on the support of the pdo adapter in the current connection class in Laravel. To be honest, since version Ecoton 2, I have abandoned Laravel support. Perhaps this package should be separated from the base Ecotone packages in ecotone-dev.

packages/Laravel/src/Config/LaravelConnectionResolver.php
dgafka: Ok, I see make sense then :)

.github/workflows/test-monorepo.yml
dgafka: True, upgrading from 7.x to 8.x could be pain, 8.0 to 8.1 isn't so hard thought

dgafka (dismissed):

packages/Laravel/src/Config/PDO/Concerns/ConnectsToDatabase.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Ecotone\Laravel\Config\PDO\Concerns;
+
+use Ecotone\Laravel\Config\PDO\Connection;
+use InvalidArgumentException;
+use PDO;
+
+trait ConnectsToDatabase
```
dgafka, line 9:
```php
/**
 * licence Apache-2.0
 */
```

Can you add lincensing to new files?

lifinsky: Yes, I can. I wanted to clarify what to do with the fact that, in fact, this is code from Laravel, and there is no copyright in their files.

dgafka: True, for MIT we are obligated to carry on licence terms for those files. 
So there should be readme in that catalog, that classes from this namespace comes from https://github.com/laravel/laravel and were licenced under [MIT](https://opensource.org/license/MIT)

packages/Laravel/src/Config/PDO/Concerns/ConnectsToDatabase.php
dgafka: Oki, make sense we should be reusing connection, in order to ensure transactional boundaries.

packages/Laravel/src/Config/PDO/Concerns/ConnectsToDatabase.php
lifinsky: @dgafka what about this?:

```php
/**
 * This file is a modified version of a class from the Laravel framework.
 *
 * Laravel is licensed under the MIT License.
 * Original authors: Taylor Otwell and the Laravel contributors.
 *
 * @license MIT (https://opensource.org/licenses/MIT)
 *
 * Modifications were made as part of the Ecotone framework under the Apache 2.0 License.
 * See LICENSE file for the Apache 2.0 License details.
 */
```

dgafka: Yep make sense, we can do it

lifinsky: @dgafka done)

dgafka (approved):

PR #407, opened 2024-11-10
Kafka message channel

dgafka:

This provides Kafka as Message Channel

- [X] I have read and agree to the contribution terms outlined in [CONTRIBUTING](https://github.com/ecotoneframework/ecotone-dev/blob/main/CONTRIBUTING.md)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   app:
-    image: simplycodedsoftware/php:8.3.13
+    image: simplycodedsoftware/php:8.3.10
     volumes:
       - "$PWD:/data/app"
     working_dir: "/data/app"
--- a/packages/Kafka/src/Inbound/KafkaAcknowledgementCallback.php
+++ b/packages/Kafka/src/Inbound/KafkaAcknowledgementCallback.php
@@ -61,7 +61,7 @@ class KafkaAcknowledgementCallback implements AcknowledgementCallback
         try {
             $this->consumer->commit($this->message);
         } catch (Exception $exception) {
-            $this->loggingGateway->info('Failed to acknowledge message. Failure happen due to: ' . $exception->getMessage(), exception: $exception);
+            $this->loggingGateway->info('Failed to acknowledge message. Failure happen due to: ' . $exception->getMessage(), ['exception' => $exception]);
 
             throw $exception;
         }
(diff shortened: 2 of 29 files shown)


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   app:
-    image: simplycodedsoftware/php:8.3.13
+    image: simplycodedsoftware/php:8.3.10
```
lifinsky, line 5: Why?)

dgafka: There is [no 8.3.13](https://hub.docker.com/r/simplycodedsoftware/php/tags?name=8.3), that had to be some bug

packages/Kafka/tests/Integration/KafkaChannelAdapterTest.php
```diff
@@ -11,29 +11,24 @@
...
- */
 /**
  * licence Enterprise
  */
 final class KafkaChannelAdapterTest extends TestCase
 {
     public function test_sending_and_receiving_from_kafka_topic(): void
     {
-        $this->markTestSkipped('CI yet to be setup');
+        $this->markTestSkipped('Kafka not yet supported');
```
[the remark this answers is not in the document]
lifinsky: What was the error during CI?

packages/Kafka/src/Inbound/KafkaAcknowledgementCallback.php
```diff
@@ -90,7 +90,7 @@ public function requeue(): void
             //            what to do here?
             //            $this->consumer->pausePartitions([$this->message->partition]);
```
lifinsky, line 91: https://github.com/php-enqueue/rdkafka/blob/d1dfe0422a15f25ef4dfda5eff5cf29633079949/RdKafkaConsumer.php#L152

@dgafka

packages/Kafka/tests/Integration/KafkaChannelAdapterTest.php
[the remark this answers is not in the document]
dgafka: There is no specific error, it just does not see any messages. 
Either the topic is not available, or the consumer is somehow not able to consume the messages

packages/Kafka/src/Inbound/KafkaAcknowledgementCallback.php
dgafka: This is using Kafka as Queue based mechanism, because in reality this message is still in the topic. 

But, ye I will most likely follow on this. This make sense for Message Channels, where there is only single consumer group. 
It does not make sense however for Inbound Adapter, as in there we may consume from topic that is consumed by multiple consumer groups. So republishing it to the same topic, could affect other parties

Issue #410, opened 2024-11-21
SIGTERM, SIGINT doesn't work with AMQP consumers


lifinsky:
**Ecotone version(s) affected**: latest

**Description**  
The consumer process cannot be terminated correctly with AMQP channel.

**How to reproduce**  
```
php bin/console ecotone:run async &
```
Run `ps` and find process pid
```
kill -s SIGTERM <pid>
```

The process will not stopped. At the same time, everything works correctly with db consumer.

lifinsky:
@dgafka Hi)

lifinsky:
A bit of debugging in InterceptedConsumer
```
while ($this->shouldBeRunning()) {
            \var_dump('--run--');
```

```
private function shouldBeRunning(): bool
    {
        if (! $this->shouldBeRunning) {
            return false;
        }

        foreach ($this->consumerInterceptors as $consumerInterceptor) {
            if ($consumerInterceptor->shouldBeStopped()) {
                \var_dump('--should be stopped--');
```


```
php bin/console ecotone:run db
string(7) "--run--"
string(7) "--run--"
^Cstring(21) "--should be stopped--"
```

```
php bin/console ecotone:run async
string(7) "--run--"
string(7) "--run--"
^Cstring(7) "--run--"
string(7) "--run--"
```

lifinsky:
Interceptors list:

Ecotone\Messaging\Endpoint\Interceptor\LimitMemoryUsageInterceptor
Ecotone\Messaging\Endpoint\Interceptor\SignalInterceptor
Ecotone\Messaging\Endpoint\Interceptor\ConnectionExceptionRetryInterceptor

lifinsky:
@dgafka `$extQueue->consume(function (\AMQPEnvelope $extEnvelope, \AMQPQueue $q)` leads to this problem

lifinsky:
https://github.com/php-enqueue/enqueue-dev/issues/332

lifinsky:
Possible solution: inherit from AmqpSubscriptionConsumer and replace the `$extQueue->consume` call with the code below:

```php
            for (;;) {
                $extEnvelope = $extQueue->get();
                if (!$extEnvelope) {
                    usleep(100000);
                    return;
                }

                $message = $this->context->convertMessage($extEnvelope);
                $message->setConsumerTag($q->getConsumerTag());

                /**
                 * @var AmqpConsumer
                 * @var callable     $callback
                 */
[...]

PR #412, opened 2024-11-22
Rework AMQP subscription consumer

lifinsky:
Refusal from AMQP queue consume method that blocks signals. Details #410 

## Description of Changes
Unfortunately, signals are very important for the correct shutdown of applications, especially in the Kunernetes pods. The AMQPQueue::consume used by AmqpSubscriptionConsumer class from the enqueue/amqp-ext package blocks signals and has been replaced by an implementation using AMQPQueue::get method. keeping the same interface.


dgafka:
Done with different pr

PR #414, opened 2024-11-26
Amqp get instead of consume

### LINKED ISSUE #410: SIGTERM, SIGINT doesn't work with AMQP consumers

dgafka:

pnctl signals are not executed due to blocking consume, therefore we have to turn it off. 
The minus of this is consume is quicker in the consumption process, and it works with Streams. 
So if this will be fixed in the amqp extension, we can revert that back.

fixes #410 


dgafka:
@lifinsky I've not found an way to handle pnctl signals with consume, so changing that to get.

lifinsky:
@dgafka Hi, but where is this change in code?

lifinsky:
I see, in parent class method...

PR #421, opened 2024-12-13
Event store

jlabedo:

Event store with resiliency, scalability and availability

## Description of Changes

TODO


packages/EventSourcingV2/src/EventStore/SQL/PostgresEventStore.php
```diff
@@ -0,0 +1,540 @@
...
+    after_transaction_id IS NULL 
+        OR 
+    after_transaction_id < pg_current_xact_id())
+AND (
+    before_transaction_id IS NULL 
+        OR 
+    before_transaction_id < pg_current_xact_id())
+ORDER BY name
+FOR SHARE
+SQL);
```
lifinsky, line 257: heredoc adjusts indentation based on the position of the enclosing element, so you can use normal indentation.

packages/EventSourcingV2/tests/EventStore/Fixtures/PostgresEventCounterProjector.php
```diff
@@ -0,0 +1,74 @@
...
+        if ($this->streams && !in_array((string) $event->streamEventId->streamId, $this->streams, true)) {
+            return;
+        }
+        $statement = $this->connection->prepare(<<<SQL
+            INSERT INTO {$this->tableName} (event_type, counter)
+            VALUES (?, 1)
+            ON CONFLICT (event_type) DO
+                UPDATE SET counter =  {$this->tableName}.counter + 1
+                WHERE  {$this->tableName}.event_type = ?
+        SQL);
```
lifinsky, line 35: Missing indent for SQL

packages/EventSourcingV2/src/Ecotone/EcotoneAsynchronousProjectionRunner.php
```diff
@@ -0,0 +1,45 @@
...
+        }
+        /** @var EcotoneProjector $projector */
+        $projector = $this->ecotoneProjectors->get($command->subscription);
+
+        while(true) {
+            $eventPage = $this->eventStore->readFromSubscription($command->subscription);
+            foreach ($eventPage->events as $event) {
+                $projector->project($event);
+            }
+            $this->eventStore->ack($eventPage);
```
lifinsky, line 36: Ack does not require a commit and opening a new transaction? And if it is in one transaction, then it will be very large in the number of statements. If the transaction is not used for projection run, it may lost ack changes.

packages/EventSourcingV2/src/EventStore/SQL/PostgresEventStore.php
```diff
@@ -0,0 +1,540 @@
...
+
+            $transaction->commit();
+        } catch (\Throwable $e) {
+            $transaction->rollBack();
+            throw $e;
+        }
+
+
+        // Execute missing events
+        $missingEventsLoop = 0;
```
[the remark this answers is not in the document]
jlabedo: I don't get your statement.

We are projecting events in order by avoiding gaps on line 356. When there are no more events to catchup, we switch to inline, but events discarded by gaps are still not handled (and won't be handled by inline processing). This is the reason of this code: handling eventually missing events.

By the way, for now there is no lock on the events table, so there is a small period of time where events can be projected in wrong order: Events handled on line 398 can be projected after first "inline" events. It may be ok, but if it is not, we can add a brief lock on events table.

packages/EventSourcingV2/src/Ecotone/EcotoneAsynchronousProjectionRunner.php
jlabedo: Yes for now it's wrapped in Ecotone's Dbal transaction. But @lifinsky is right, if we are doing the catchup here (full reset) then it will be very large in the number of statements.
I see two options:
1. We let the asynchronous runner wrapped in Ecotone's transaction but we need to first catchup the projection (projection:init command) before the async projection is triggered.
2. We disable Ecotone's transaction on this endpoint an we manually manage small transactions there

packages/EventSourcingV2/src/Ecotone/EcotoneAsynchronousProjectionRunner.php
```diff
@@ -0,0 +1,45 @@
...
+
+            if ($command->until !== null && $eventPage->endPosition->isBefore($command->until)) {
```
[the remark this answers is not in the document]
jlabedo: This is the idea of this code: project until there are no more events to handle (You are right I am missing a check to continue while `$eventPage->events` is not empty)

But the idea behind this is also to force the runner to wait for new events that are still after an interleaved transaction (a gap).
Imagine we start a very long transaction appending events that this projection is not intersested in. Then a second transaction start, appends events related to this projection. The runner starts but won't see any events: there is a gap because of the first transaction still running.
I we stop the runner without waiting, the projection would be stale until another event triggers the projection.
That is the reason why we are stating that this trigger is not complete if it does not reach AT LEAST the position of the last event that triggered the projection.
Does it makes sense ?

packages/EventSourcingV2/src/Ecotone/EcotoneAsynchronousProjectionRunner.php
lifinsky: Option 2

PR #423, opened 2024-12-22
Storing state stored via business repository

dgafka:

Storing state stored via business repository, as it was not covered.


packages/Ecotone/src/Modelling/Config/AggregrateHandlerModule.php
```diff
@@ -526,6 +527,7 @@ public function registerBusinessRepositories(InterfaceToCallRegistry $interfaceT
                     $gatewayParameterConverters = [
                         GatewayHeaderBuilder::create($interface->getFirstParameter()->getName(), AggregateMessage::CALLED_AGGREGATE_OBJECT),
                         GatewayHeaderBuilder::create($interface->getFirstParameter()->getName(), AggregateMessage::RESULT_AGGREGATE_OBJECT),
+                        GatewayPayloadBuilder::create($interface->getFirstParameter()->getName())
```
dgafka, line 530: That was missing.

PR #427, opened 2025-01-19
Distributed Bus with Service Mapping

dgafka:
This introduces new Enterprise functionality - Distributed Bus with Service Mapping

## Description of Changes

image
image

### Command Distribution
image
image

### Event Distribution
image
image
image

Read more about additional features for Service Mapping in documentation page.


lifinsky:
@dgafka Subscription routing is an interesting feature. Can we customize event distribution with custom strategy (something like Presend interceptor)?

lifinsky:
@dgafka If a service does not have a corresponding distributed event handler, even without this module, the service will not create an extra queue and will not receive messages. Perhaps it is worth describing this better. The module probably fits the problem when two services send the same events, but the third service wants to subscribe to only the event of the first service, and the fourth to the second.

lifinsky:
image

I checked for custom service (one of more than ten services in our application).

dgafka:
>  Am I right that now without this module in a situation with a distributed event and 4 services, 3 services will not constantly receive all events from the fourth service if they do not have the corresponding routing key? Am I wrong?

Yes, that's true.

dgafka:
> Can we customize event distribution with custom strategy (something like Presend interceptor)?

image

> If a service does not have a corresponding distributed event handler, even without this module, the service will not create an extra queue binding and will not receive messages. Perhaps it is worth describing this better. 

image

Issue #296, opened 2024-02-22
Add Source Aggregate Id to Events


dgafka:
**Description**  
Currently after Event is published, we don't have information where the event came from. This can be found as side effect under `_aggregate_id` when Event Sourcing is used. We can make this solution generic and apply for all ES and non ES solution a Message Header `source_aggregate_id`, so the behaviour is the same.

dgafka:
Each event published from within Aggregate contains now metadata about the origination

![Image](https://github.com/user-attachments/assets/ee00544e-fa18-406d-be66-85274108485e)

PR #442, opened 2025-02-08
Enrich events on publish and improve customized event sourced repositories

dgafka:

It also simplifies using building new custom ES repositories, and give more features available out of the box in that situation.

Changes:
- Now events stored and published contains information about originating Aggregate (aggregate id, version, type)
- Named Events, Aggregate Types become generic mechanism, that can be used with custom repositories
- Snapshots are now generic mechanism, that can be used with custom repositories
- Change in Event Sourced Repository to allow loading from given moment in time

## Description of Changes

image
image
image


dgafka:
This solves #296

lifinsky:
@dgafka amazing

lifinsky:
@dgafka Is it possible for an event sourcing repository interface to have both a method for retrieving an event stream by aggregate ID and a standard find method that returns the aggregate itself?

What happens to the stream if a snapshot already exists for the aggregate?

Can this also be documented somehow?

Thank you!

dgafka:
> What happens to the stream if a snapshot already exists for the aggregate? Can this also be documented somehow?

That was not part of yet generic functionality, but as part of this change, I've made snapshots possible to use for custom repositories. Documentation added :)

> @dgafka Is it possible for an event sourcing repository interface to have both a method for retrieving an event stream by aggregate ID and a standard find method that returns the aggregate itself?

What would be the point of this? The repository is responsible for storing and fetch events, having object being returned would disallow of using generic snapshoting mechanism (as we wouldn't be able to load from given point in time). 

If you would like to fetch given Aggregate instance in application level code using custom repositories, you can use [Repository Business Interface](https://docs.ecotone.tech/modelling/command-handling/business-interface/repository#for-event-sourced-aggregate) for that.

lifinsky:
> That was not part of generic functionality, but as part of this change I've made snapshots possible to use for custom repositories. Documentation added :)

That is, will it be able to get all events even if there is a snapshot, while the business interface will return the aggregate using the snapshot (when $fromAggregateVersion = 1)?

packages/Ecotone/src/Modelling/EventStream.php
```diff
@@ -10,13 +10,13 @@
...
     private int $aggregateVersion;
-    /** @var Event[]|SnapshotEvent[]  */
+    /** @var Event[]  */
     private array $events;
 
     private function __construct(int $aggregateVersion, array $events)
     {
         foreach ($events as $event) {
-            Assert::isTrue($event instanceof Event || $event instanceof SnapshotEvent, 'Event is not type of Event or SnapshotEvent' . get_class($event));
+            Assert::isTrue($event instanceof Event, 'Event is not type of Event or SnapshotEvent' . get_class($event));
```
lifinsky, line 19:
```php
Assert::isTrue($event instanceof Event, 'Event is not type of Event' . get_class($event));
```

dgafka:
cc: @jlabedo this change make it more easy to integrate new event sourcing packages

packages/PdoEventSourcing/src/EventSourcingRepository.php
```diff
@@ -48,41 +45,30 @@ public function canHandle(string $aggregateClassName): bool
...
 
-    public function findBy(string $aggregateClassName, array $identifiers): EventStream
+    public function findBy(string $aggregateClassName, array $identifiers, int $fromAggregateVersion = 1): EventStream
     {
         $aggregateId = reset($identifiers);
-        $aggregateVersion = 0;
+        $aggregateVersion = $fromAggregateVersion;
         $streamName = $this->getStreamName($aggregateClassName, $aggregateId);
         $aggregateType = $this->getAggregateType($aggregateClassName);
         $snapshotEvent = [];
```
lifinsky, line 54: Should be removed?

dgafka: true

lifinsky:
> If snapshot is used, it will be always be used no matter of the scenario (business interface / command). So if snapshot is done at 300th event, then fetching will start from 301.

That means it is necessary to independently implement the mechanism for retrieving all events, including those for which a snapshot already exists. Did I understand correctly?

dgafka:
> That means it is necessary to independently implement the mechanism for retrieving all events, including those for which a snapshot already exists. Did I understand correctly?

It's only needed to implement retrieving events, (snapshot event was temporary concept, it's removed completely, as it's not needed anymore). 
After command is executed on the Aggregate, Ecotone will take a look whatever it's time to do the snapshot (take a look on `SaveAggregateService`). If it's so, then Ecotone will use instance of the Aggregate which is in memory, to store it using DocumentStore, and on subsequest will use it to load it.

[Snapshoting is explained in separate section](https://docs.ecotone.tech/modelling/event-sourcing/event-sourcing-introduction/persistence-strategy/snapshoting).

lifinsky:
> It's only needed to implement retrieving events, (snapshot event was temporary concept, it's removed completely, as it's not needed anymore). After command is executed on the Aggregate, Ecotone will take a look whatever it's time to do the snapshot (take a look on `SaveAggregateService`). If it's so, then Ecotone will use instance of the Aggregate which is in memory, to store it using DocumentStore, and on subsequest will load it.

We're talking about slightly different things. I'm referring to the case where I don't need to delete the snapshot, but I still want to analyze the entire event stream, including those for which a snapshot already exists. Is it possible to retrieve all events by specifying version 1 as the starting point?

dgafka:
@lifinsky this repository is only needed for custom repositories. For inbuilt repositories there is no such interface available from application level perspective. 

Therefore if your roll out your own repository, you can use it the way you want from your Application.
However if you ask, if you continue using Ecotone to reconstruct the Aggregate, yet reconstruct it without using snapshot (if it's configured for given aggregate), then there is no such option. 

However you could run Ecotone Lite with different configuration where there is no snapshot configuration, therefore loading the Aggregate instance from the beginning of the stream.

lifinsky:
It's a pity, I was hoping there would be an option to retrieve the full event stream. But implementing the retrieval of all events for an aggregate manually isn't difficult, so it's not a big issue.

dgafka:
@lifinsky I am not sure what's the use, but you could also work with [EventStore directly](https://docs.ecotone.tech/modelling/event-sourcing/event-sourcing-introduction/working-with-event-streams) to get the events

packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadEventSourcingAggregateService.php
```diff
@@ -93,4 +117,22 @@ public function process(Message $message): ?Message
...
+                $aggregateVersionPropertyName = $property->getName();
+                break;
+            }
+        }
+
+        return $propertyReader->getPropertyValue(
+            PropertyPath::createWith($aggregateVersionPropertyName),
+            $aggregate
+        );
+    }
```
jlabedo, line 137: This could be optimized a bit if we do class introspection in the builder

packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadEventSourcingAggregateService.php
```diff
@@ -59,14 +68,29 @@ public function process(Message $message): ?Message
...
+                $documentStore = $this->container->get($config['documentStore']);
+
+                $aggregate = $documentStore->findDocument(SaveAggregateService::getSnapshotCollectionName($this->aggregateClassName), SaveAggregateService::getSnapshotDocumentId($aggregateIdentifiers));
+
+                if (! is_null($aggregate)) {
+                    $aggregateVersion = $this->getAggregateVersion($aggregate);
+                    Assert::isTrue($aggregateVersion > 0, sprintf('Serialization for snapshot of %s is set incorrectly, it does not serialize aggregate version', $aggregate::class));
+                }
+            }
+        }
```
jlabedo, line 85: This could be handled in `packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadAggregateServiceBuilder.php`, then you can inject directly an optional `DocumentStore` in this service ?

jlabedo (commented):
Nice, thanks!

packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadEventSourcingAggregateService.php
dgafka: yep true, I did not focus on refactors here, just copied over the logic :)

PR #447, opened 2025-02-11
Deduplication clearing

dgafka:

Deduplication have run automatically inside the Message Consumer, however deleting record may result in index rebuild which will cause database locks. So when this happened performance of the application was affected, as other places in the system which used deduplication had to wait for lock to finish.

## Description of Changes

Ecotone now moves the cleaning of deduplication to separate console command, which can be triggered periodically in background. It also introduces batch removal to avoid big volume of data being deleted at once, which could lock table for longer period of time.

image

### DIFF
--- a/packages/Dbal/src/Deduplication/DeduplicationInterceptor.php
+++ b/packages/Dbal/src/Deduplication/DeduplicationInterceptor.php
@@ -2,7 +2,9 @@
 
 namespace Ecotone\Dbal\Deduplication;
 
+use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Types;
 use Ecotone\Dbal\DbalReconnectableConnectionFactory;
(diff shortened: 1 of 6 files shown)


dgafka:
cc: @lifinsky

lifinsky:
@dgafka Thank you!

packages/Dbal/src/Deduplication/DeduplicationInterceptor.php
```diff
@@ -162,4 +168,17 @@ private function getConnection(ConnectionFactory $connectionFactory): Connection
 
         return $context->getDbalConnection();
     }
+
+    public function getMessageIdsToRemoval(ConnectionFactory $connectionFactory): array
```
lifinsky, line 172: Lock for share?

lifinsky: + skip locked for concurrently clearing @dgafka

lifinsky: > Can you propose a PR with change that will work across Postgres and Mysql?

Something like this prototype code?
```php
use Doctrine\DBAL\Query\QueryBuilder;
use Doctrine\DBAL\Connection;

class CustomQueryBuilder extends QueryBuilder
{
    private ?string $lockMode = null;

    public function __construct(Connection $connection)
    {
        parent::__construct($connection);
    }

    public function lockForShare(): self
    {
        $platform = $this->getConnection()->getDatabasePlatform()->getName();

        if ($platform === 'mysql') {
            $this->lockMode = 'LOCK IN SHARE MODE';
        } elseif ($platform === 'postgresql') {
            $this->lockMode = 'FOR SHARE';
        }

        return $this;
    }

    public function getSQL(): string
    {
        $sql = parent::getSQL();

        if ($this->lockMode !== null) {
            $sql .= ' ' . $this->lockMode;
        }

        return $sql;
    }
}
```

dgafka: I guess look good. Tests on CI will fail if it will not work, so feel free to come with a PR

PR #457, opened 2025-03-13
Identifier metadata mapping fails when metadata is set by interceptor

unixslayer:
I have read and agree to the contribution terms outlined in [CONTRIBUTING](https://github.com/ecotoneframework/ecotone-dev/blob/main/CONTRIBUTING.md).

### DIFF
--- a/packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadAggregateMessageProcessor.php
+++ b/packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadAggregateMessageProcessor.php
@@ -35,6 +35,11 @@ final class LoadAggregateMessageProcessor implements MessageProcessor
     public function process(Message $message): ?Message
     {
         $resultMessage = MessageBuilder::fromMessage($message);
+        $messageType = TypeDescriptor::createFromVariable($message->getPayload());
+
+        if (!$message->getHeaders()->containsKey(AggregateMessage::AGGREGATE_ID)) {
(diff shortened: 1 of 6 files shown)


packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadAggregateMessageProcessor.php
```diff
@@ -35,14 +35,18 @@ public function __construct(
     public function process(Message $message): ?Message
     {
         $resultMessage = MessageBuilder::fromMessage($message);
+        $messageType = TypeDescriptor::createFromVariable($message->getPayload());
+
+        if (!$message->getHeaders()->containsKey(AggregateMessage::AGGREGATE_ID)) {
```
dgafka, line 40: :+1:

packages/Ecotone/tests/Modelling/Unit/LoadAggregateServiceBuilderTest.php
```diff
@@ -51,11 +52,27 @@ public function test_loading_aggregate_by_metadata()
...
+                BeforeFinishOrder::class
+            ],
+            containerOrAvailableServices: [
+                new BeforeFinishOrder()
+            ],
+        );
+        $ecotone->sendCommandWithRoutingKey('order.start', $oderId = 100);
+        $ecotone->publishEvent(PaymentWasDoneEvent::create($oderId));
+        $this->assertEquals('done', $ecotone->sendQueryWithRouting('order.status', metadata: ['aggregate.id' => $oderId]));
+    }
```
dgafka, line 74: Can you also add two more test cases for asynchronous scenarios, where we set `aggregate.id` using `Before` and `Presend`? 

It seems we lack some coverage here, so as this is not ensured at the test case level, we can easily break that behaviour during refactors.

dgafka (approved):

Issue #453, opened 2025-02-26
ConcurrencyException in event sourcing aggregate does not rollback outer transaction


lifinsky:
**Ecotone version(s) affected**: latest

**Description**  
When a command handler for an Event Sourcing Aggregate is executed inside an outer command handler that has its own transaction and a synchronous retry mechanism for ConcurrencyException, Ecotone incorrectly retries the inner command handler without rolling back the outer transaction.

As a result, PostgreSQL raises the following error:
`Error 25P02: current transaction is aborted, commands ignored until end of transaction`

**How to reproduce**  

1. Define an Event Sourcing Aggregate with a command handler that may trigger a ConcurrencyException.
2. Define an outer command handler that:
- Calls the inner command handler.
- Includes a synchronous retry mechanism for ConcurrencyException in `InstantRetryConfiguration::createWithDefaults()->withCommandBusRetry()`.
3. Send a command that triggers a concurrency conflict.
4. Observe that Ecotone retries only the inner command handler, without rolling back the outer transaction.
5. The next database operation fails with PostgreSQL error 25P02, because the outer transaction is already aborted.

**Possible Solution**  
- When ConcurrencyException (or any allowed for sync retry) occurs in the inner command handler, Ecotone should rollback the entire outer transaction.
- The outer command handler should then restart from the beginning, ensuring that the retry logic operates in a clean transactional state.

**Actual Behavior**
- Ecotone only retries the inner command handler but does not rollback the outer transaction.
- This leads to a broken transaction state, where subsequent commands inside the same transaction are ignored due to PostgreSQL transaction failure.

lifinsky:
@dgafka This is important because it's not always feasible to work with a single aggregate within a command handler.

lifinsky:
One possible solution is to not register the instant retry interceptor inside another (parent) instant retry interceptor either always or when the DbalTransactionInterceptor is used.

lifinsky:
Currently, this is the most critical bug when using event sourcing aggregates within parent command handlers, regardless of what the parent command handler relates to. It effectively leads to the inability to handle a race condition conflict in the event sourcing aggregate related to its version, especially in synchronous handlers, as Prooph relies on the violation of uniqueness during insertion.

dgafka:
Can you explain reasoning to use Command Bus inside Command Bus?
Command Bus is meant to begin messaging flow, yet when we are already in the messaging flow, there is no need to re-trigger that.

[...]

PR #464, opened 2025-03-26
feat: avoid nested retries

dgafka:

In case Command Bus with enabled retries is executed within context that is already retried, it leads to situation where retries are multipled. 
Besides in case nested retried was executed due to SQL exception then it will not resolve itself as it requires transaction rollback which is done by outer retries.


dgafka:
Resolved #453

Issue #467, opened 2025-03-27
Custom event sourced repository interface stops working after upgrade to 1.248.0


lifinsky:
**Ecotone version(s) affected**: from 1.248.0

**Description**  
After upgrading to Ecotone v1.248.0, event sourced aggregates no longer work when using business repository interface. The problem occurs only when we have custom repositories for state-based aggregates.

**How to reproduce**  
1. We have an event-sourced aggregate:
```php
#[EventSourcingAggregate]
class Balance
{
    use WithAggregateVersioning;

    #[Identifier]
    private Uuid $id;
    private string $value;

    #[CommandHandler]
    public static function create(OpenBalanceCommand $command): array
    {
        return [new BalanceOpened($command->id, $command->value)];
    }

    #[EventSourcingHandler]
    public function applyBalanceOpened(BalanceOpened $event): void
    {
        $this->id = $event->id;
        $this->value = $event->value;
    }

    public function getValue(): string
    {
        return $this->value;
    }
}
```

2. We define a custom repository interface:
```php
interface BalanceRepositoryInterface
{
    #[Repository]
    public function getById(Uuid $id): Balance;
}
```

3. We also define custom state-based aggregate repository:
```php
#[Repository]
readonly class UserRepository extends AbstractRepository implements UserRepositoryInterface
{
    public static function getAggregateClass(): string
    {
        return User::class;
    }

    public function getById(UuidV4 $id): User
    {
        // ...
    }
}
```

```php
abstract readonly class AbstractRepository  implements StandardRepository
{
    public function __construct(protected EntityManagerInterface $entityManager, private bool $flushOnSave) {}

    /**
     * @return class-string<object>
     */
    abstract public static function getAggregateClass(): string;

    public function find(Uuid $id): null|object
    {
        return $this->findBy(static::getAggregateClass(), ['id' => (string) $id]);
    }

    public function get(Uuid $id): object
    {
        $entity = $this->find($id);

        if (!$entity) {
            throw EntityNotFoundException::byId(static::getAggregateClass(), $id);
        }

        return $entity;
    }

    public function remove(Uuid $id): void
    {
        $entity = $this->entityManager->find(static::getAggregateClass(), $id);
        if ($entity) {
            $this->entityManager->remove($entity);
            $this->entityManager->flush();
        }
    }

    public function canHandle(string $aggregateClassName): bool
    {
        return $aggregateClassName === static::getAggregateClass();
    }

    /**
[...]

PR #449, opened 2025-02-21
Event sourcing: decouple from core

jlabedo:
## Why is this change proposed?

Decouple event sourcing from core Ecotone aggregate modeling.
The end goal is to push event sourcing related logic into its dedicated package `Ecotone\EventSourcing`

## Description of Changes

Refactoring.
Introduction of `AggregateRepository` interface which is implemented by `EventSourcedRepositoryAdapter` and `StandardRepositoryAdapter`.

## Pull Request Contribution Terms

- [X] I have read and agree to the contribution terms outlined in [CONTRIBUTING](https://github.com/ecotoneframework/ecotone-dev/blob/main/CONTRIBUTING.md).

PR #468, opened 2025-03-28
Fix repository handling

jlabedo:

Probable fix for #467 

## Description of Changes

I did not find a simple way to add a failing test, it should be done with Symfony (for a non emulated ManagerRegistry), Dbal and Event sourcing packages.

#449, by trying to decouple event sourcing from core, removes the majority of "`isEventSourced`" checks. For this reason, a `StandardRepository` implementation should be more precise in its `canHandle` as it can be asked if it can handle event sourced aggregates classes.


lifinsky:
We can check on Monday. Thanks.

dgafka (approved):

PR #473, opened 2025-04-03
passing metadata matcher with projection running configuration

unixslayer:
allow passing `MetadataMatcher` with projection running configuration

`GapDetection` will be passed with `ProjectionRuntimeConfiguration` instead of `ProjectionSetupConfiguration` which was always using the one from DI container. 


unixslayer:
@lifinsky this PR changes `GapDetection` management. Previously `GapDetection` container dependency was overwriting the one passed with `ProjectionSetupConfiguration`. Now `GapDetection` can be passed individually with `ProjectionRunningConfiguration`

packages/PdoEventSourcing/src/ProjectionSetupConfiguration.php
```diff
@@ -33,22 +37,11 @@ public function __construct(
 
     public static function create(string $projectionName, ProjectionLifeCycleConfiguration $projectionLifeCycleConfiguration, string $eventStoreReferenceName, ProjectionStreamSource $projectionStreamSource, ?string $asynchronousChannelName): static
     {
-        return new self($projectionName, $projectionLifeCycleConfiguration, $eventStoreReferenceName, $projectionStreamSource, $asynchronousChannelName, projectionOptions: [PdoEventStoreReadModelProjector::OPTION_GAP_DETECTION => new GapDetection(
-            retryConfig: [0, 5, 50, 500, 800],
```
dgafka, line 37: Can we make this as default config? That was initial intention

packages/PdoEventSourcing/src/ProjectionSetupConfiguration.php
```diff
@@ -2,13 +2,17 @@
...
+use const E_USER_DEPRECATED;
+
 use Ecotone\Messaging\Config\Container\DefinedObject;
 use Ecotone\Messaging\Config\Container\Definition;
 use Ecotone\Messaging\NullableMessageChannel;
 use Ecotone\Messaging\Support\Assert;
 use Prooph\EventStore\Pdo\Projection\GapDetection;
 use Prooph\EventStore\Pdo\Projection\PdoEventStoreReadModelProjector;
 
+use function trigger_error;
```
dgafka, line 14: I guess that can be dropped?

dgafka (commented):
Nice tests :) 
Just please confirm on the default gap detection and let's go

lifinsky:
🥳

lifinsky:
> @lifinsky this PR changes `GapDetection` management. Previously `GapDetection` container dependency was overwriting the one passed with `ProjectionSetupConfiguration`. Now `GapDetection` can be passed individually with `ProjectionRunningConfiguration`

Is it possible to define gap detection configuration for synchronous projections in Ecotone?
I see that the current class supports event-sourced and polling projections, but it would also be useful for synchronous projections. By default, I believe synchronous projections (just like the other types) should have default gap detection settings available.

unixslayer:
@lifinsky @jlabedo 

I've managed to test sync projections in  `packages/PdoEventSourcing/tests/Integration/GapDetectionInSynchronousProjectionTest.php`. I was able to "simulate" projection by removing one of the events "for a moment" from the stream. Depending on how `GapDetection` is configured, it behaves as expected.

unixslayer:
> Thanks! @dgafka @unixslayer Should these capabilities perhaps be documented for all types of projection updates (sync, event-driven, or polling) through an example in the service configuration?
> 
> https://docs.ecotone.tech/modelling/event-sourcing/setting-up-projections/executing-and-managing/running-projections#projection-custom-configuration

Do you mean like by describing different behaviors for different gap_detection configs?

lifinsky:
> Do you mean like by describing different behaviors for different gap_detection configs?
Rather about such possibilities with any type of projection

packages/PdoEventSourcing/src/Prooph/Metadata/MetadataMatcher.php
```diff
@@ -0,0 +1,109 @@
...
+        ?FieldType $fieldType = null
+    ): self {
+        $this->validateValue($operator, $value);
+
+        if (null === $fieldType) {
+            $fieldType = FieldType::METADATA();
+        }
+
+        $self = clone $this;
+        $self->data[] = ['field' => $field, 'operator' => $operator, 'value' => $value, 'fieldType' => $fieldType];
```
dgafka, line 58: Operator, fieldType is object in this, does it work correctly when dumping to defintion

dgafka (approved):
Nice test coverage :heart:

dgafka:
For sure we can document gap detection, as I don't think it's described anywhere

dgafka (changes requested):
I think hanging tests are somehow related to changes. As only event-sourcing package is getting stuck on tests.

packages/PdoEventSourcing/src/Prooph/Metadata/MetadataMatcher.php
unixslayer: objects won't, but I've created MetadataMatcher as DefinedObject using PHP Enums for Operator and FieldType and those works properly.

unixslayer:
> For sure we can document gap detection, as I don't think it's described anywhere

I've already started describing GapDetection behavior in documentation. Should be done tomorrow.

PR #474, opened 2025-04-11
metadata propagation issue

unixslayer:

when message flow goes outside of aggregate which was called as first, its identifier is always propagated

## Test scenario

1. `Item` gets added to `Basket`
2. `Basket` (separated EH) records reservation
3. `ItemInventory` (different aggregate) also records reservation

Problem occurs when loading `ItemInventory` due `Basket` identifier is in use.

### DIFF
--- a/packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadAggregateMessageProcessor.php
+++ b/packages/Ecotone/src/Modelling/AggregateFlow/LoadAggregate/LoadAggregateMessageProcessor.php
@@ -38,7 +38,7 @@ final class LoadAggregateMessageProcessor implements MessageProcessor
         $messageType = TypeDescriptor::createFromVariable($message->getPayload());
 
         if (! $message->getHeaders()->containsKey(AggregateMessage::AGGREGATE_ID)) {
-            throw AggregateNotFoundException::create("Can't call Aggregate {$this->aggregateClassName}:{$this->aggregateMethod} as identifier header is missing. Please check your identifier mapping in {$messageType->toString()}. Have you forgot to add #[TargetIdentifier] in your Command or `aggregate.id` in metadata?");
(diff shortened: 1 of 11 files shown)


lifinsky:
@unixslayer Hi!
Does this example work in a synchronous flow? It seems quite unexpected to me.
What if we explicitly annotate the event with `#[TargetIdentifier("itemId")]`?

unixslayer:
> @unixslayer Hi! Does this example work in a synchronous flow? It seems quite unexpected to me. What if we explicitly annotate the event with `#[TargetIdentifier("itemId")]`?

@lifinsky scenario works fine for sync flow but it should also work for async. The problem is, that `ecotone.modelling.aggregate.id` is propagated to a queue with value of first aggregate and according to `packages/Ecotone/src/Modelling/AggregateIdentifierRetrevingService.php:42` it won't be resolved anymore as it already exists. But last message is handled by another BC so it should be resolved once more.

unixslayer:
also, `#[TargetIdentifier]` doesn't make it work

lifinsky:
> > @unixslayer Hi! Does this example work in a synchronous flow? It seems quite unexpected to me. What if we explicitly annotate the event with `#[TargetIdentifier("itemId")]`?
> 
> @lifinsky scenario works fine for sync flow but it should also work for async. The problem is, that `ecotone.modelling.aggregate.id` is propagated to a queue with value of first aggregate and according to `packages/Ecotone/src/Modelling/AggregateIdentifierRetrevingService.php:42` it won't be resolved anymore as it already exists. But last message is handled by another BC so it should be resolved once more.

Yes, this is a critical bug, since each message must relate to a specific aggregate and should not depend on the parent aggregate ID according to the event handler logic — unlike a command, which is expected to have a single handler.

dgafka (approved):
Great found :)

PR #479, opened 2025-05-22
WIP: Event routing behavior

jlabedo:

Synchronous and asynchronous event handler priorities have the same behavior.
Respect ordering of `NamedEvent`.

### DIFF
--- a/packages/Ecotone/tests/Modelling/Fixture/PriorityEventHandler/SynchronousPriorityHandlerWithInheritance.php
+++ b/packages/Ecotone/tests/Modelling/Fixture/PriorityEventHandler/SynchronousPriorityHandlerWithInheritance.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Test\Ecotone\Modelling\Fixture\PriorityEventHandler;
+
+use Ecotone\Messaging\Attribute\Endpoint\Priority;
+use Ecotone\Modelling\Attribute\EventHandler;
+use Ecotone\Modelling\Attribute\QueryHandler;
+
+/**
+ * licence Apache-2.0
+ */
+final class SynchronousPriorityHandlerWithInheritance
+{
+    public array $triggers = [];
+
+    #[Priority(3)]
+    #[EventHandler(endpointId: 'middlePriorityHandler')]
+    public function middlePriorityHandler(OrderWasPlaced $event): void
+    {
+        $this->triggers[] = 'middlePriorityHandler';
+    }
+
+    #[Priority(1)]
+    #[EventHandler(endpointId: 'lowerPriorityHandlerWithObjectRouting')]
+    public function lowerPriorityHandler(object $event): void
+    {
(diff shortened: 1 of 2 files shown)


packages/Dbal/tests/Fixture/ORM/AutogeneratedIdentifier/Person.php
```diff
@@ -43,20 +44,15 @@ class Person
...
     {
         $this->name = $name;
-
-        $this->recordThat(new PersonRegistered($personId, $name));
     }
 
-    #[CommandHandler]
-    public static function register(RegisterPerson $command): static
+    #[CommandHandler(self::REGISTER_COMMAND)]
+    public static function register(string $name): static
```
jlabedo, line 53: The previous test was not working in `ORMTest`: if there is a personId in the command, the commandBus will return the provided `personId` instead of the autogenerated id in previous behaviour.

If there is no personId in the command, the test passes.
I think it is OK has there is no point of passing an Id in the factory command if the id is autogenerated.

packages/Ecotone/tests/Messaging/Unit/Handler/Processor/InterceptorsOrderingTest.php
```diff
@@ -170,10 +170,6 @@ public function test_aggregate_with_factory_method(): void
 
         self::assertEquals(
             [
-                'beforeChangeHeaders',
-                'before',
-                'afterChangeHeaders',
-                'after',
```
jlabedo, line 176: This PR fixes running interceptors twice in case of factory and action methods on aggregate

jlabedo: This is not the case anymore, as it broke a test where identifier metadata mappings where added from a before interceptor

packages/Ecotone/src/Modelling/Config/MessageHandlerRoutingModule.php
```diff
@@ -459,50 +150,56 @@ public static function getRoutingInputMessageChannelForEventHandler(AnnotatedFin
...
-        } else {
-            $inputChannelName = $annotationForMethod->getInputChannelName();
-        }
-
-        if (! $inputChannelName) {
-            $interfaceToCall = $interfaceToCallRegistry->getFor($registration->getClassName(), $registration->getMethodName());
-            if ($interfaceToCall->hasNoParameters()) {
-                throw ConfigurationException::create("Missing class type hint or routing key for {$registration}.");
+        $routingEventHandlers = ExtensionObjectResolver::resolve(RoutingEventHandler::class, $extensionObjects);
+        $commandBusRoutingConfig = new BusRoutingMapBuilder(true, $routingEventHandlers, $messagingConfiguration);
```
jlabedo, line 159: Passing `$messagingConfiguration` here is to allow routing event handlers to register messaging handlers. This could be avoided by passing `$messagingConfiguration` in the `create` method of module classes. This will be handled in another PR to avoid fat PR.

unixslayer (approved):

PR #481, opened 2025-05-30
use aggregate version from its property for command handlers

### LINKED ISSUE #480: In sync, stateful, event sourced workflow aggregate is being called with invalid target version
**Ecotone version(s) affected**: 1.254.1

**Description**

In sync, stateful, event sourced workflow aggregate is being called with target version `0` which causes its persistence to fail due second attempt tries to save aggregate in version `1` which was already persisted.

**How to reproduce**  

Test for following example, fails with `Ecotone\Messaging\Support\ConcurrencyException: Some of the aggregate IDs or event IDs have already been used in the same stream.`.

```php
#[EventSourcingSaga(withInternalEventRecorder: true)]
class Cycle
{
    use WithEvents;
    use WithAggregateVersioning;

    #[Identifier]
    private string $cycleId;

    #[CommandHandler(
        routingKey: 'cycle.submitAnAudit',
        outputChannelName: 'cycle.validateAnAudit',
        identifierMetadataMapping: ['cycleId' => 'cycleId'],
    )]
    public static function startCycleBySubmittingAnAudit(
        #[Header('cycleId')] string $cycleId
    ): self {
        $cycle = new self();
        $cycle->recordThat(new CycleStarted($cycleId));

        return $cycle;
    }

    #[CommandHandler(
        routingKey: 'cycle.submitAnAudit',
        outputChannelName: 'cycle.validateAnAudit',
        identifierMetadataMapping: ['cycleId' => 'cycleId'],
    )]
    public function submitAnAudit(#[Header('audit')] Audit $audit): Audit
    {
        return $audit;
    }

    #[CommandHandler(
        routingKey: 'cycle.validateAnAudit',
        outputChannelName: 'cycle.conductAnAudit'
    )]
    public function validateAnAudit(#[Header('audit')] Audit $audit): Audit
    {
        // some validations

        return $audit;
    }

    #[CommandHandler(
        routingKey: 'cycle.conductAnAudit',
        outputChannelName: 'cycle.issueACertificate'
    )]
    public function conductAnAudit(#[Payload] Audit $audit): ?Certificate
    {
        $this->recordThat(new AuditConducted($this->cycleId, $audit->auditId));

        return $audit->certificate;
    }

    #[CommandHandler(
        routingKey: 'cycle.issueACertificate',
    )]
    public function issueACertificate(#[Payload] Certificate $certificate): void
    {
        $this->recordThat(new CertificateIssued($this->cycleId, $certificate->certificateId));
    }

    #[EventSourcingHandler]
    public function applyCycleStarted(CycleStarted $event): void
    {
        $this->cycleId = $event->cycleId;
    }
}
```

Proper test case will be provided with possible solution.

unixslayer:
Resolves #480 


lifinsky:
Symfony 7.3: [DependencyInjection] Don't skip classes with private constructor when autodiscovering.

✅ Symfony 7.3 Service Class Requirements

In order for a class to be used as a service in Symfony 7.3:

1. Its constructor must be public,

2. It must have a factory method (either static or external), and you must explicitly configure Symfony to use it:

- via factory: in service configuration
- or using the attribute:

`#[Autoconfigure(constructor: 'staticFactoryMethod')]`

lifinsky:
Symfony 7.3 introduces stricter validation for autowired services. The Fixture\User\User class has a private constructor and is not intended to be a service. Symfony attempted to autoconfigure it, resulting in a RuntimeException.

 Fix

Updated services.yaml to explicitly exclude the User.php file:

```
Fixture\:
    resource: '../tests/Fixture/*'
    exclude:
        - '../tests/Fixture/User/User.php'
```

✅ Result

This prevents Symfony from treating the test fixture as a service, resolving the error while keeping other fixtures discoverable.

packages/Ecotone/src/Modelling/AggregateFlow/CallAggregate/CallAggregateResultToMessageConverter.php
```diff
@@ -39,18 +39,17 @@ public function convertToMessage(Message $requestMessage, mixed $result): ?Messa
 
         if ($this->isCommandHandler) {
             $calledAggregate = $requestMessage->getHeaders()->containsKey(AggregateMessage::CALLED_AGGREGATE_INSTANCE) ? $requestMessage->getHeaders()->get(AggregateMessage::CALLED_AGGREGATE_INSTANCE) : null;
-            $versionBeforeHandling = $requestMessage->getHeaders()->containsKey(AggregateMessage::TARGET_VERSION) ? $requestMessage->getHeaders()->get(AggregateMessage::TARGET_VERSION) : null;
```
jlabedo, line 42: We should keep looking in message headers before anything else: the point is to pass a target version directly in the command. Like this:
```php
$commandBus->sendWithRouting('cycle.submitAnAudit', metadata: [AggregateMessage::TARGET_VERSION => 0]);
```
WDYT ?

unixslayer: @jlabedo in provided example, target version is always `0` between each workflow steps so after first aggregate modification, every other will fail due `ConcurrencyException`

jlabedo: You would send the aggregate version to the frontend, then you post the command from frontend with the expected version before handling. If the version don't match then you have a concurrency exception. This is to avoid doing an action on an aggregate without the last known state from the client.

lifinsky: I think that for the synchronous flow, it makes sense to store target version only in the case of the initial command handlers, and only if it's an update of an existing aggregate.

dgafka:
> I've overwritten TargetVersion for reply message after aggregate is saved so further processing in synchronous flow will have latest version of an aggregate.

But what in situation when output channel will be actually pointing to different aggregate class? If I follow correctly we may then try to load that Aggregate with version of the initial aggregate?

PR #485, opened 2025-06-08
stateful event sourced workflow with multiple aggregates

unixslayer:

### DIFF
--- a/packages/PdoEventSourcing/tests/Fixture/StatefulEventSourcedWorkflowWithMultipleAggregates/AddItemToBasket.php
+++ b/packages/PdoEventSourcing/tests/Fixture/StatefulEventSourcedWorkflowWithMultipleAggregates/AddItemToBasket.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Test\Ecotone\EventSourcing\Fixture\StatefulEventSourcedWorkflowWithMultipleAggregates;
+
+class AddItemToBasket
+{
+    public function __construct(public string $basketId, public string $itemId, public int $quantity)
+    {
+    }
+}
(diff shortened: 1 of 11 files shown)


unixslayer:
@dgafka this PR is a follow-up to #481 

Scenario here is to build synchronous, stateful workflow with multiple aggregates. Apparently there is an issue with mapping aggregate id when workflow jumps from one aggregate to another as identifier of initial aggregate remains in use. The same happens, when `itemInventory.makeReservation` is asynchronous as in second commit here.

However, I wonder if this would really be a use case. As models from example can be modified independently from each other. Therefore, so they will probably be in their own BCs so connecting them in synchronous workflow would lead to unexpected behavior. I'd rather don't connect them in the same workflow as we most likely have to deal with eventual consistency here.

cc: @jlabedo @lifinsky

lifinsky:
Hi! I’d like to suggest a feature that would greatly improve developer experience and workflow flexibility.

Currently, the outputChannelName option for command or event handlers only supports returning a single message. However, there are many real-world cases where a handler naturally results in multiple follow-up commands or events. For example:

- Broadcasting a webhook to multiple applications of the same merchant (each with its own webhook URL)
- Splitting one logical command into multiple async commands

In these scenarios, being able to return an array or iterable from a handler — and having Ecotone automatically route each item to the specified OutputChannel — would simplify the architecture and eliminate boilerplate.

unixslayer:
Although, that would make `outputChannelName` obsolete as its design lies in passing single message withing a workflow. With `#[Result]` approach, Ecotone may handle each element directly.

dgafka (approved):

Issue #484, opened 2025-06-08
Error serializing UuidV7 during event sourcing after upgrading to Ecotone v1.256


m2mtech:
**Ecotone version(s) affected**: 1.256.0

**Description**  
When upgrading from `v1.255` to `v1.256`, commands containing a `UuidV7` parameter cause an error during serialization in event sourcing:

```
Error: Typed static property Symfony\Component\Uid\UuidV7::\$seedParts must not be accessed before initialization
```

https://github.com/m2mtech/ecotoneUuidProblem/blob/ea98ecd5c106f85a1ad32077e3311aa0f49ecd96/error.log#L1

This happens during event sourcing when a change command contains the projected and modified entity.

**How to reproduce**  
1. Clone the test repository: [ecotoneUuidProblem](https://github.com/m2mtech/ecotoneUuidProblem)
2. Run:
   ```bash
   docker compose run php83 composer update
   docker compose run php83 bin/phpunit
   ```

**Context**  
* PHP 8.3 (also tested with 8.4)
* Symfony 6.4 (also tested with 7.3)
* The issue was introduced with Ecotone `v1.256`
* No issue on `v1.255`

lifinsky:
Hi. You have a bug in your event-sourced aggregate — you need to handle the ProductCreated event via an event sourcing handler and assign the product id (aggregate id) and name.

lifinsky:
https://docs.ecotone.tech/modelling/event-sourcing/event-sourcing-introduction/event-sourcing-aggregates/different-ways-to-record-events

lifinsky:
@m2mtech You also need converter for UuidV7 class:
https://docs.ecotone.tech/modules/jms-converter

m2mtech:
I did not need one for 1.255?

dgafka:
@m2mtech I don't know context of your project, but I would say that sending an Entity via Command is odd, yet it should still work. 
The problem that you face now happens because we want to serialize Command from Object to Array to retrieve identifier (AggregateIdentifierRetrievingService). However Aggregate Identifier should be retrievable from the Command itself without serialization. So from a quick look, it looks like there is some behaviour change.

PR #487, opened 2025-06-10
fix: aggregate id resolver

dgafka:

Currently when Command is sent we are trying to serialize it to array, as we lost the context of expected Command Class. 
This use case was not covered by test case, therefore that behaviour was lost with latest refactor.

Solves #484 


dgafka:
@jlabedo was first https://github.com/ecotoneframework/ecotone-dev/pull/486/files

PR #493, opened 2025-06-18
Use a Clock abstraction to provide easier testing in userland

jlabedo:

It is now considered a good practice to abstract the clock to provide a way of mocking time in tests.

## Description of Changes

This PR adds a Clock abstraction that is injected where needed in core ecotone. It extends the PSR clock interface to provide some helper functions for a better Ecotone developer experience, notably a Duration abstraction which provide better clarity (avoid questions like `$timeout` property, is it in seconds or milliseconds ?)


packages/Ecotone/src/Messaging/Handler/Recoverability/RetryTemplate.php
```diff
@@ -52,31 +49,9 @@ public function calculateNextDelay(int $retryNumber): int
         return $this->delayForRetryNumber($retryNumber);
     }
 
-    public function runCallbackWithRetries(Closure $closure, Message $message, string $exceptionClass, LoggingGateway $logger, string $retryMessage): void
```
jlabedo, line 55: This function has been moved to a new `RetryRunner` service that has access to the clock (to avoid a call to global `usleep`)

packages/Ecotone/src/Messaging/Scheduling/DatePoint.php
```diff
@@ -0,0 +1,146 @@
...
+    {
+        return parent::createFromMutable($object);
+    }
+
+    public static function createFromTimestamp(int|float $timestamp): static
+    {
+//        Not passing phpstan check
+//        if (\PHP_VERSION_ID >= 80400) {
+//            return parent::createFromTimestamp($timestamp);
+//        }
```
dgafka, line 51: some left over?

packages/Ecotone/src/Messaging/Scheduling/Clock.php
```diff
@@ -1,21 +1,53 @@
...
+    private static ?EcotoneClockInterface $globalClock = null;
+
+    public function __construct(
+        private readonly ?PsrClockInterface $clock = null,
+    ) {
+    }
+
+    public static function set(PsrClockInterface $clock): void
+    {
+        self::$globalClock = $clock instanceof EcotoneClockInterface ? $clock : new self($clock);
```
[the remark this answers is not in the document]
jlabedo: I am not sure I understand your question but I will try to answer.
If you provide a `Psr\Clock\ClockInterface` in the container you will also have a `Clock` instance available in the container that uses your provided `Psr\Clock\ClockInterface` to get now(). It will just decorate it to return a  `DatePoint` instead of a `DateTimeImmutable`.

But, if you use anywhere the global clock `Clock::get`, you have to manually set it to have the same as in your container. By default, the global clock is `NativeClock` whatever is your `Psr\Clock\ClockInterface` provided in the container.
I think setting the global clock on each container bootstrap is a bit odd.

packages/Ecotone/src/Messaging/Scheduling/DatePoint.php
dgafka: Ye, let's change phpstan. Can be in follow up

packages/Ecotone/src/Messaging/Scheduling/Clock.php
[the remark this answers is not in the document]
dgafka: This does reference to my previous comment about global state. In general I would rather avoid any global things as they do tend to affects things in unpredictable ways, and makes things that are otherwise easy - hard e.g. bootstraping multiple applications within same process, doing parallel testing etc.

So my point here was to not use this method at all, because in ideal scenario something like global clock wouldn't exist in isolated environments. 
But yea, I get your argument about `MessageHeaders::createMessageHeadersWith`, we are using there microtime which is actually working as global clock. So that's not a blocker, it wasn't just visible as so far :)

PR #495, opened 2025-06-25
Fix message conversion of named event

jlabedo:

Bug: If a named event is published with routing to an object catch all handler, it is effectively routed since #479 but the conversion fails

### DIFF
--- a/packages/Ecotone/tests/Modelling/Fixture/RoutingTest/RoutingTestHandler.php
+++ b/packages/Ecotone/tests/Modelling/Fixture/RoutingTest/RoutingTestHandler.php
@@ -6,14 +6,8 @@ declare(strict_types=1);
 
 namespace Test\Ecotone\Modelling\Fixture\RoutingTest;
 
-use Ecotone\Messaging\Attribute\Asynchronous;
-use Ecotone\Modelling\Attribute\EventHandler;
-use Test\Ecotone\Modelling\Fixture\NamedEvent\GuestWasAddedToBook;
-
 class RoutingTestHandler
 {
-    public const ASYNC_CHANNEL = 'async';
-
     protected array $messages = [];
 
     public function getMessages(): array
--- a/packages/Ecotone/tests/Modelling/Unit/RoutingTest.php
+++ b/packages/Ecotone/tests/Modelling/Unit/RoutingTest.php
@@ -79,7 +79,7 @@ class RoutingTest extends TestCase
             classesToResolve: [$handler::class, GuestWasAddedToBook::class, GuestWasAddedToBookConverter::class],
             containerOrAvailableServices: [$handler, new GuestWasAddedToBookConverter()],
             enableAsynchronousProcessing:  $async ? [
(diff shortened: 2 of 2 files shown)


packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Converter/PayloadConverter.php
```diff
@@ -70,6 +73,22 @@ public function getArgumentFrom(Message $message): mixed
                 ) {
                     $convertedData = $this->doConversion($data, $sourceTypeDescriptor, $sourceMediaType, $resolvedTargetParameterType, $parameterMediaType);
                 }
+            } elseif ($this->mapper && $message->getHeaders()->containsKey(MessageBusChannel::EVENT_CHANNEL_NAME_BY_NAME)) {
```
dgafka, line 76: This feels like something was not handled well along the way. 
So far object and interface conversion did worked, because it did fallback to `__TYPE_ID__` header.
What have changed?

jlabedo: Yes, I feel it wrong too. I am not happy of what is done here.

I am adding some tests on 1.255.0 tag (before #479) to understand how it was working before and what changed. For now, I can already say that the async test case was failing on 1.255.0:
```php
class CatchAllAsynchronousHandler
{
    const ASYNC_CHANNEL = 'async';

    #[EventHandler(listenTo: GuestWasAddedToBook::EVENT_NAME, endpointId: 'async_catch_all')]
    #[Asynchronous('async')]
    public function handleAsync(GuestWasAddedToBook $message): void
    {
    }
}

$eventBus->publishEventWithRoutingKey(GuestWasAddedToBook::EVENT_NAME, [
    'bookId' => 'an-id',
    'guestName' => 'John Doe',
]);

$ecotoneLite->run('async');

// Fail with error: "Argument #1 ($message) must be of type GuestWasAddedToBook, array given"
```

jlabedo (commented):
I am quite happy with this implementation. The type alias can be added to __TypeId__ header, making it more robust against class refactorings.

Now named events can be converted in both directions: object->array or array->object, synchronous or asynchronous (the bug existed before routing behavior change)

packages/Ecotone/src/Messaging/Handler/Processor/MethodInvoker/Converter/PayloadConverter.php
```diff
@@ -24,15 +24,10 @@
  */
 class PayloadConverter implements ParameterConverter
 {
-    public function __construct(private ConversionService $conversionService, private string $interfaceName, private string $parameterName, private Type $targetType)
+    public function __construct(private ConversionService $conversionService, private ?EventMapper $mapper, private string $interfaceName, private string $parameterName, private Type $targetType)
```
dgafka, line 27: Can we make it non nullable and always inject? 
As it requires handling nulls and having the knowledge why it's null now.

PR #515, opened 2025-09-17
Allow interfaces in EventSourcingHandler

kovenant:
According to [documentation](https://docs.ecotone.tech/modelling/command-handling/external-command-handlers/event-handling#subscribe-to-interface-or-abstract-class) it's possible to use Interfaces or Abstract Classes in EventHandlers. However, when I try to use an interface in an EventSourcingHandler, I get the error "<...> is Event Sourcing Handler and should have first parameter as Event Class type hint."

## Description of Changes
Add `|| interface_exists` check in InterfaceParameter::isObjectTypeHint

### DIFF
--- a/packages/Ecotone/src/Messaging/Handler/InterfaceParameter.php
+++ b/packages/Ecotone/src/Messaging/Handler/InterfaceParameter.php
@@ -185,7 +185,7 @@ final class InterfaceParameter
     /**
      * @return bool
      */
-    public function isObjectTypeHint(): bool
+    public function isClassOrInterface(): bool
     {
         return class_exists($this->getTypeHint()) || interface_exists($this->getTypeHint());
     }
--- a/packages/Ecotone/src/Messaging/Handler/InterfaceParameter.php
+++ b/packages/Ecotone/src/Messaging/Handler/InterfaceParameter.php
@@ -187,7 +187,7 @@ final class InterfaceParameter
      */
     public function isObjectTypeHint(): bool
     {
-        return class_exists($this->getTypeHint());
+        return class_exists($this->getTypeHint()) || interface_exists($this->getTypeHint());
     }
 
     public function __toString()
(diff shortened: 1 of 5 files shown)


packages/Ecotone/src/Messaging/Handler/InterfaceParameter.php
```diff
@@ -187,7 +187,7 @@ public function isAnnotation(): bool
      */
     public function isObjectTypeHint(): bool
     {
-        return class_exists($this->getTypeHint());
+        return class_exists($this->getTypeHint()) || interface_exists($this->getTypeHint());
```
dgafka, line 190: If we change this, let's also refactor the method name to be more straight forward to `isClassOrInterface`. 
This will follow same pattern as in `TypeDescriptor` class.

kovenant: done :)

kovenant:
> @kovenant sure,
> 
> You can place test scenario in `Test\Ecotone\Modelling\Unit\EventSourcingAggregateTest`

I added the test scenario in `Test\Ecotone\Modelling\Unit\EventSourcingAggregateTest` by simply adding an interface to `AggregateCreated` and listening for it in `EventSourcingAggregateExample`. It's ok for you? I think creating a separate test case, aggregate, command, and event for such a small change seems like overkill :) 

Also my example above was only for reproducing the issue. We expected an exception to be thrown before the fix. Now we expecting normal flow.

dgafka:
@kovenant it's fine to roll out specific test scenario for such cases. Thanks to that in month or two, we will exactly know that this test was rolled out for preserving given behaviour. Tests using Ecotone Lite are really fast, therefore it's not a problem to create them for any cases needed. 

You don't even need to create separate classes for given test case, you can see how @jlabedo did that in [his pr for example](https://github.com/ecotoneframework/ecotone-dev/pull/519/files#diff-0197632ff404196a9eeba7bd9d829fede428a1131c375055ca742e612612eb8eR199)

Anyways not a blocker for this, let's merge and let go with this one. 
Thanks for contribution :)

PR #519, opened 2025-09-22
Bug projection priority

### LINKED ISSUE #517: Channel is already registered with another priority when mixing handlers with and without priority in the same projection
**Ecotone version(s) affected**: 1.256.0

**Description**  
After upgrading to Ecotone 1.256.0, our tests started failing with the following error:
```
RuntimeException: Channel projection_handler_card_transaction is already registered with another priority
/app/vendor/ecotone/ecotone/src/Modelling/Config/Routing/BusRoutingMapBuilder.php:248
/app/vendor/ecotone/ecotone/src/Modelling/Config/Routing/BusRoutingMapBuilder.php:134
/app/vendor/ecotone/ecotone/src/Modelling/Config/Routing/BusRoutingMapBuilder.php:119
/app/vendor/ecotone/ecotone/src/Modelling/Config/Routing/BusRoutingMapBuilder.php:89
/app/vendor/ecotone/ecotone/src/Modelling/Config/MessageHandlerRoutingModule.php:152
/app/vendor/ecotone/ecotone/src/Messaging/Config/Container/Compiler/ModuleConfigurationCompilerPass.php:50
/app/vendor/ecotone/ecotone/src/Messaging/Config/MessagingSystemConfiguration.php:856
```

This happens only when at least one event handler in a projection has an explicit priority, while other handlers in the same projection do not specify any priority.

**How to reproduce**  
```php
#[Projection(name: self::TABLE, fromStreams: CardTransaction::class)]
readonly class CardTransactionProjection extends AbstractReadModel
{
    public const string TABLE = 'card_transaction';

    public static function getTableName(): string
    {
        return self::TABLE;
    }

    #[EventHandler]
    public function onRefundWasSuccessful(RefundSuccessful $event): void
    {
        //...
    }

    #[Priority(50)]
    #[EventHandler]
    public function onRefundWasDeclined(RefundDeclined $event): void
    {
        //...
    }

    //...
}
```

**Possible Solution**  
Ecotone should either:

Treat missing priority as priority = 0 automatically. Allow mixing handlers with and without explicit priority without throwing an exception.

jlabedo:
Fixes #517 

## Description of Changes
Before this PR, channels were prioritize globally for all routing keys of a router.
This PR prioritize channels for each routing key, fixing the triggering of projections that has the same target channel for multiiple routing keys that can now have different priorities

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -94,11 +94,9 @@ jobs:
         with:
           clean: false
 
-      - name: Remove cached files
+      - name: Remove cached files but keep phpbench results
         run: |
-          rm -rf Monorepo/ExampleApp/Lite/var/cache
-          rm -rf Monorepo/ExampleApp/Symfony/var/cache
-          rm -rf Monorepo/ExampleApp/Laravel/framework/cache
+          git clean -fXd -e .phpbench
 
       - name: Install dependencies
         run: composer update --${{ matrix.stability }} --prefer-dist --no-interaction
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -96,7 +96,7 @@ jobs:
 
       - name: Remove cached files but keep phpbench results
         run: |
-          git clean -fXd -e .phpbench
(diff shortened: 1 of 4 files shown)


lifinsky:
👍

Issue #513, opened 2025-09-11
InstantRetry is incompatible with Transactional Interceptor when using event-sourced aggregates (Prooph Event Store)


lifinsky:
**Ecotone version(s) affected**: latest

### Description
When using Ecotone InstantRetry together with a Transactional Interceptor for Doctrine DBAL, retries are executed inside the same broken database transaction, instead of starting a fresh transaction for each retry.

This makes the retry mechanism ineffective, especially when working with Prooph PDO Event Store, where optimistic concurrency is enforced via a unique constraint on
`(_aggregate_type, _aggregate_id, _aggregate_version).`

If a concurrency conflict occurs, PostgreSQL marks the current transaction as aborted, and all subsequent retry attempts immediately fail with:
`ERROR: current transaction is aborted, commands ignored until end of transaction block`

**How to reproduce**  

- Use Prooph PDO Event Store with Ecotone.
- Run two concurrent commands for the same aggregate.
- The second command triggers a concurrency conflict:

 ```
ERROR: duplicate key value violates unique constraint "_56de7a2a6187e9678628e42314087eb0ce67e3a1_expr_expr1_expr2_idx"
DETAIL: Key ((metadata ->> '_aggregate_type'),
             (metadata ->> '_aggregate_id'),
             (metadata ->> '_aggregate_version')) =
       (Domain\Entity\CustomerAccountBalance, 81e1cb83-ec5f-48ec-bb27-3b5b79e2e1b9, 123) already exists.
```

- InstantRetry catches the ConcurrencyException and retries inside the same DBAL transaction, which is already aborted.
- Retry attempt immediately fails with:
`ERROR: current transaction is aborted, commands ignored until end of transaction block`

### Possible Solution
**Default order change:**
Make InstantRetryInterceptor wrap the TransactionalInterceptor by default.

Correct call order:
InstantRetry → TransactionalInterceptor → CommandHandler

Fail-safe rollback:
On ConcurrencyException, force an immediate rollback before retry logic continues.

Configuration option:
Provide a way for developers to explicitly define interceptor priority if custom setups are needed.

### Context
**Current retry setup**
```php
public function instantRetries(): InstantRetryConfiguration
{
    return InstantRetryConfiguration::createWithDefaults()
        ->withCommandBusRetry(
            isEnabled: true,
            retryExceptions: [
                ConcurrencyException::class,
                OptimisticLockException::class,
            ],
        );
}
```

```
[...]

PR #522, opened 2025-09-24
Projecting improvements 1

jlabedo:
## Description of Changes
- fix DbalProjectionStorage when using manager registry in the container
- Init all projections in one console command: `ecotone:projecting:init --all=true`
- better gap timeout processing for event store stream source (default to 60 seconds timeout)
- don't add gaps if the event timestamp is after gap timeout (avoids adding gaps while backfilling)

### DIFF
--- a/packages/Ecotone/src/Projecting/Config/ProjectingConsoleCommands.php
+++ b/packages/Ecotone/src/Projecting/Config/ProjectingConsoleCommands.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 namespace Ecotone\Projecting\Config;
 
 use Ecotone\Messaging\Attribute\ConsoleCommand;
+use Ecotone\Messaging\Attribute\ConsoleParameterOption;
 use Ecotone\Projecting\ProjectionRegistry;
 use InvalidArgumentException;
 
@@ -18,12 +19,21 @@ class ProjectingConsoleCommands
     }
 
     #[ConsoleCommand('ecotone:projection:init')]
-    public function initProjection(string $name): void
+    public function initProjection(?string $name = null, #[ConsoleParameterOption] bool $all = false): void
     {
-        if (! $this->registry->has($name)) {
-            throw new InvalidArgumentException("There is no projection with name {$name}");
+        if ($name === null) {
+            if (! $all) {
+                throw new InvalidArgumentException('You need to provide projection name or use --all option');
+            }
+            foreach ($this->registry->projectionNames() as $projection) {
+                $this->registry->get($projection)->init();
+            }
+        } else {
+            if (! $this->registry->has($name)) {
+                throw new InvalidArgumentException("There is no projection with name {$name}");
+            }
+            $this->registry->get($name)->init();
         }
-        $this->registry->get($name)->init();
     }
 
     #[ConsoleCommand('ecotone:projection:backfill')]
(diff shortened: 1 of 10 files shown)


lifinsky:
@jlabedo Hi! Thank you! Could you please suggest the recommended value for synchronous projections to set in GapDetection, and remind me if it's possible to configure it for specific synchronous projections using something like:

```php
->withOption(
    PdoEventStoreReadModelProjector::OPTION_GAP_DETECTION,
    new GapDetection(
        retryConfig: [0, 5, 10, 50, 100, 100, 200, 500],
        detectionWindow: new GapDetection\DateInterval('PT2M'),
    )
)
```

packages/PdoEventSourcing/src/Projecting/PartitionState/DbalProjectionStateStorage.php
```diff
@@ -28,7 +29,7 @@ class DbalProjectionStateStorage implements ProjectionStateStorage
     private bool $initialized = false;
 
     public function __construct(
-        DbalConnectionFactory $connectionFactory,
+        DbalConnectionFactory|ManagerRegistryConnectionFactory $connectionFactory,
```
lifinsky, line 32: Was a different connection used when we use Doctrine? #513

lifinsky:
@jlabedo @unixslayer It looks like there is still no way to tune gap detection for a specific synchronous projection or for all projections through ServiceContext using ProjectionRunningConfiguration.

#473

packages/PdoEventSourcing/src/Projecting/PartitionState/DbalProjectionStateStorage.php
jlabedo: Hello @lifinsky , the type depends on your configuration. 

This do not relate to #513 . This is a work in progress implementation for a new projecting system, it DOES NOT impact the existing one. You should not be impacted by my following PRs until we stabilize and document this new projecting system.

packages/Ecotone/src/Projecting/Config/ProjectingConsoleCommands.php
```diff
@@ -18,12 +19,21 @@ public function __construct(private ProjectionRegistry $registry)
     }
 
     #[ConsoleCommand('ecotone:projection:init')]
-    public function initProjection(string $name): void
+    public function initProjection(?string $name = null, #[ConsoleParameterOption] bool $all = false): void
```
dgafka, line 22: Can we add a test case for running this? `$ecotoneLite->runConsoleCommand(...)`

packages/PdoEventSourcing/tests/Projecting/GapAwarePositionIntegrationTest.php
```diff
@@ -142,15 +141,21 @@ public function test_gap_timeout_cleaning(): void
...
+        // Verify: Gaps 2, 4 should be removed (old timestamps), gap 6 should remain (recent timestamps)
         $newTracking = GapAwarePosition::fromString($result->lastPosition);
         self::assertSame([6], $newTracking->getGaps());
+
+        // Delay 3 more second to exceed timeout for all gaps (6 seconds since insertion of the last event)
+        self::$clock->sleep(Duration::seconds(4));
+        $result = $streamSource->load($result->lastPosition, 100);
+        $newTracking = GapAwarePosition::fromString($result->lastPosition);
+        self::assertSame([], $newTracking->getGaps());
     }
```
dgafka, line 160: Won't we get flooded by gaps, when we will treat Event Log as table for multiple Aggregate Types? 
This means that we position number will be global across different aggregates, meaning gaps will be be constant (as we won't be filtering out those).

jlabedo: We could think of another tracking strategy in this future (postgres transaction ids, ...).
But if we want to dumbly use this strategy with an event log table, you would query all events in the event log (not filtered), then some would be handled (if EventHandler) other won't.
So yes, you would see gaps from all aggregates but I think this is ok.

PR #524, opened 2025-09-27
Instant retries with dbal transactions

dgafka:

This is to reproduce issue https://github.com/ecotoneframework/ecotone-dev/issues/513

- [] I have read and agree to the contribution terms outlined in [CONTRIBUTING](https://github.com/ecotoneframework/ecotone-dev/blob/main/CONTRIBUTING.md).

### DIFF
--- a/packages/Dbal/tests/Fixture/Transaction/Nested/NestedDbalHandlers.php
+++ b/packages/Dbal/tests/Fixture/Transaction/Nested/NestedDbalHandlers.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Test\Ecotone\Dbal\Fixture\Transaction\Nested;
+
+use Ecotone\Messaging\Attribute\Parameter\Reference;
+use Ecotone\Modelling\Attribute\CommandHandler;
+use Ecotone\Modelling\Attribute\QueryHandler;
+use Ecotone\Modelling\CommandBus;
+use Enqueue\Dbal\DbalConnectionFactory;
+use Interop\Queue\ConnectionFactory;
+
+final class NestedDbalHandlers
+{
+    #[CommandHandler('nested.prepare')]
+    public function prepare(#[Reference(DbalConnectionFactory::class)] ConnectionFactory $connectionFactory): void
+    {
+        $connection = $connectionFactory->createContext()->getDbalConnection();
+        $connection->executeStatement('DROP TABLE IF EXISTS nested_events');
+        $connection->executeStatement('CREATE TABLE nested_events (id INT PRIMARY KEY)');
+    }
+
+    #[CommandHandler('nested.inner')]
+    public function inner(#[Reference(DbalConnectionFactory::class)] ConnectionFactory $connectionFactory): void
+    {
+        $connection = $connectionFactory->createContext()->getDbalConnection();
+        $connection->executeStatement('INSERT INTO nested_events (id) VALUES (1)');
+    }
+
+    #[CommandHandler('nested.outer')]
+    public function outer(#[\Ecotone\Messaging\Attribute\Parameter\Reference] CommandBus $commandBus): void
+    {
+        // Trigger a nested synchronous command bus call
+        $commandBus->sendWithRouting('nested.inner');
+    }
+
+    #[QueryHandler('nested.count')]
+    public function count(#[Reference(DbalConnectionFactory::class)] ConnectionFactory $connectionFactory): int
+    {
+        $connection = $connectionFactory->createContext()->getDbalConnection();
+        return (int) $connection->executeQuery('SELECT COUNT(*) FROM nested_events')->fetchOne();
+    }
+}
+
--- a/packages/Dbal/tests/Fixture/Transaction/Nested/TestCountingLogger.php
+++ b/packages/Dbal/tests/Fixture/Transaction/Nested/TestCountingLogger.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Test\Ecotone\Dbal\Fixture\Transaction\Nested;
+
+use Ecotone\Messaging\Handler\Logger\LoggingGateway;
+use Ecotone\Messaging\Message;
+use Stringable;
+
+final class TestCountingLogger implements LoggingGateway
+{
+    private int $started = 0;
+    private int $committed = 0;
+
+    public function info(Stringable|string $message, array|Message|null $context = [], array $additionalContext = []): void
+    {
+        $msg = (string) $message;
+        if ($msg === 'Database Transaction started') {
+            $this->started++;
+        }
+        if ($msg === 'Database Transaction committed') {
+            $this->committed++;
+        }
+    }
+
+    public function error(Stringable|string $message, array|Message|null $context = [], array $additionalContext = []): void
+    {
+        // no-op for tests
+    }
+
+    public function critical(Stringable|string $message, array|Message|null $context = [], array $additionalContext = []): void
+    {
+        // no-op for tests
+    }
+
+    // PSR-3 interface stubs
+    public function emergency($message, array $context = []): void {}
+    public function alert($message, array $context = []): void {}
+    public function warning($message, array $context = []): void {}
+    public function notice($message, array $context = []): void {}
+    public function debug($message, array $context = []): void {}
+    public function log($level, $message, array $context = []): void {
+        if ($level === \Psr\Log\LogLevel::INFO) {
+            $msg = (string) $message;
+            if ($msg === 'Database Transaction started') {
+                $this->started++;
+            }
+            if ($msg === 'Database Transaction committed') {
+                $this->committed++;
+            }
+        }
+    }
+
+    public function getStartedCount(): int
+    {
+        return $this->started;
+    }
+
+    public function getCommittedCount(): int
+    {
+        return $this->committed;
+    }
+
+    public function reset(): void
+    {
+        $this->started = 0;
+        $this->committed = 0;
+    }
+}
+
--- a/packages/Dbal/tests/Integration/Transaction/NestedDbalTransactionTest.php
+++ b/packages/Dbal/tests/Integration/Transaction/NestedDbalTransactionTest.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Test\Ecotone\Dbal\Integration\Transaction;
+
+use Ecotone\Dbal\Configuration\DbalConfiguration;
+use Ecotone\Lite\EcotoneLite;
+use Ecotone\Messaging\Config\ModulePackageList;
+use Ecotone\Messaging\Config\ServiceConfiguration;
+
+use Enqueue\Dbal\DbalConnectionFactory;
+use PHPUnit\Framework\TestCase;
+use Test\Ecotone\Dbal\DbalMessagingTestCase;
+use Test\Ecotone\Dbal\Fixture\Transaction\Nested\NestedDbalHandlers;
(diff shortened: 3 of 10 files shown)


dgafka:
@lifinsky current implementation does ensure that it does work as expected. I need your input here to reproduce that issue

lifinsky:
> @dgafka , I think the problem could come from the event store not sharing the same connection as the rest of the transaction. Prooph use PDO directly. When you initialize an event store, it will keep this connection until the end, even if it has been closed. If Ecotone retries by starting a new connection and a new transaction, the event store would still see the old PDO. Does it make sense ?

It might be that too, but the problem also comes from the nested transaction created by Doctrine's Unit of Work when using built-in repositories for Doctrine entities as state aggregates. Am I right?

lifinsky:
@jlabedo @dgafka Hi! Today I checked one of the services where a saga state aggregate is used via a Doctrine entity together with an event-sourcing aggregate, and where we haven’t yet added the deprecated setting `use_savepoints: false`.
On a conflict in the event-sourcing aggregate, the synchronous retry doesn’t work, because the savepoint created by Doctrine for persisting the saga aggregate is rolled back, not the entire transaction.

lifinsky:
Perhaps it makes sense to add a wrapper to Ecotone and recommend in the documentation configuring it for Doctrine when using Symfony?

lifinsky:
But then there’s a problem with committing during the flush — it will commit the top-level transaction, meaning the transactionNestingLevel mechanism is needed, but without creating savepoints.

lifinsky:
@dgafka @jlabedo And when transactions are enabled on asynchronous endpoints with a command handler inside, the instant retry should be registered at the entry point of the asynchronous endpoint, before the transaction begins, rather than in the middle of it? Right now, as I understand it, there’s also a problem because the retry occurs in the middle of the transaction opened in the asynchronous endpoint.

dgafka:
@lifinsky @jlabedo indeed Event Store would reuse old connection after reconnecting, plus the instant retry would not re-trigger the db transaction on second run. This PR now fixes that and it works with Postgres.
However some tests fails with Mysql, which I suspect is related to implicit commits (due to DDL changes). 

@lifinsky if you're able help in fixing those tests, then we can push it forward. Otherwise it will have to wait when I will find a bit more time, to finish it up.

lifinsky:
@dgafka Where do you see the need for an emergency call to handleImplicitCommitAfterDDLOperation?
In our case, all schemas are created in advance through migrations, so DDL operations do not occur at runtime.

packages/PdoEventSourcing/src/Prooph/LazyProophEventStore.php
```diff
@@ -302,11 +331,13 @@ public function getEventStoreType(): string
...
 
-    private function getConnection(): \Doctrine\DBAL\Connection
+    public function getConnection(): \Doctrine\DBAL\Connection
     {
         $connectionFactory = new DbalReconnectableConnectionFactory($this->connectionFactory);
 
-        return $connectionFactory->getConnection();
+        /** @var DbalContext $context */
+        $context = $connectionFactory->createContext();
+        return $context->getDbalConnection();
```
lifinsky, line 340: Could you explain why we need to create a new context?

packages/PdoEventSourcing/tests/Fixture/InstantRetry/ConnectionClosingInterceptor.php
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    private function closeAllConnections(DbalConnectionFactory $connectionFactory): void
+    {
+        $context = $connectionFactory->createContext();
+        $connection = $context->getDbalConnection();
+
+        try{
+            $connection->rollBack();
```
lifinsky, line 62: Probably, it makes sense to close the connection before the instant retry, leaving the rollback as the responsibility of the transactional interceptor.

packages/Dbal/src/DbalTransaction/DbalTransactionInterceptor.php
```diff
@@ -133,8 +134,12 @@ public function transactional(MethodInvocation $methodInvocation, Message $messa
         return $result;
     }
 
-    private function isImplicitCommitException(Throwable $exception): bool
+    private function isImplicitCommitException(Throwable $exception, Connection $connection): bool
     {
+        if (!($connection->getDriver() instanceof Driver)) {
```
lifinsky, line 139: This code needs to be removed.

packages/Dbal/src/DbalTransaction/DbalTransactionInterceptor.php
```diff
@@ -92,7 +93,7 @@ public function transactional(MethodInvocation $methodInvocation, Message $messa
                     $this->logger->info('Database Transaction committed', $message);
                 } catch (Exception $exception) {
                     // Handle the case where a database did an implicit commit or the transaction is no longer active
-                    if ($this->isImplicitCommitException($exception)) {
+                    if ($this->isImplicitCommitException($exception, $connection)) {
```
lifinsky, line 96: it’s safer to analyze specifically the PDOException, rather than just checking the text of any exception

packages/Dbal/src/DbalTransaction/DbalTransactionInterceptor.php
lifinsky: MySQL
`PDOException(code: 0): There is no active transaction`

```
dev.INFO: Database Transaction started {"message_id":"35636469-e35a-4790-ae07-636826b52168","correlation_id":"b0d43484-c0c4-47be-af96-e80c77dc4ffe","parent_id":"e36db6d0-003e-424e-ac91-0e08be04da82"} 
dev.INFO: Deduplication table was created  
dev.INFO: Executing Event Handler App\MultiTenant\Application\CustomerService::sendNotificationWhen {"message_id":"35636469-e35a-4790-ae07-636826b52168","correlation_id":"b0d43484-c0c4-47be-af96-e80c77dc4ffe","parent_id":"e36db6d0-003e-424e-ac91-0e08be04da82"} 
dev.INFO: Message was stored in deduplication table. {"message_id":"35636469-e35a-4790-ae07-636826b52168","consumer_endpoint_id":"notifications","routing_slip":"notificationSender.target.execute"} 
dev.INFO: Implicit Commit was detected, skipping manual one. {"message_id":"35636469-e35a-4790-ae07-636826b52168","correlation_id":"b0d43484-c0c4-47be-af96-e80c77dc4ffe","parent_id":"e36db6d0-003e-424e-ac91-0e08be04da82","exception":"[object] (PDOException(code: 0): There is no active transaction at /data/app/vendor/laravel/framework/src/Illuminate/Database/PDO/Connection.php:142)
[stacktrace]
#0 /data/app/vendor/laravel/framework/src/Illuminate/Database/PDO/Connection.php(142): PDO->commit()
```

lifinsky: Just to note, implicit commit caused by DDL operations is also characteristic of Oracle, not only MySQL.
In Oracle, before executing any DDL (CREATE TABLE, ALTER TABLE, etc.), the database automatically commits all open transactions in the session. Then, after the DDL completes, it performs another automatic commit.
This means that any changes before the DDL are permanently saved, and the DDL itself cannot be rolled back - similar to MySQL, but even stricter.

packages/Dbal/src/DbalReconnectableConnectionFactory.php
```diff
@@ -111,4 +112,35 @@ public static function getWrappedConnection(object $connection): Connection
...
+
+    public static function handleImplicitCommitAfterDDLOperation(?Connection $connection): void
+    {
+//        @TODO think whatever to replace transaction interceptor implicit commit handling with this
+        return;
+        if (! $connection) {
+            return;
+        }
+
+        if (
```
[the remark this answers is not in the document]
lifinsky: This code should be kept only for MySQL and Oracle, as both databases trigger an implicit commit when executing DDL operations. However, to make it more reliable, you should catch PDOException and analyze the exception message, similar to how it’s done in TransactionalInterceptor.

packages/Dbal/src/DbalReconnectableConnectionFactory.php
```diff
@@ -111,4 +112,35 @@ public static function getWrappedConnection(object $connection): Connection
...
+            DbalReconnectableConnectionFactory::isConnectionWorkingWithImplicitCommit($connection)
+            && $connection->isTransactionActive()
+        ) {
+            try {
+                $connection->commit();
+            }catch (\Exception $exception) {
+                // ignore
+            }
+            try {
+                $connection->rollBack();
```
lifinsky, line 139: Why try a rollback?

packages/PdoEventSourcing/src/Prooph/LazyProophEventStore.php
dgafka: In case of dbal, context holds reference to same connection, no matter as how much of those we will create. 
Therefore this code being called either way should have no difference

jlabedo:
Hello @lifinsky @dgafka. This approach feels a bit hacky to me. We should avoid all implicit commits: as @lifinsky mentioned, we could end up with part of the business transaction committed while another part is not, which may lead to an inconsistent state. This actually happened to me a month ago, and it was very hard to restore consistency afterwards.

In a basic event-sourcing application, tables are usually created before any other inserts in the event store, so that could work since there is no INSERT before the autocommit. However, if we also use state aggregates or perform other operations on the database, the issue would surface.

My proposal would be: if Ecotone is configured to auto-create tables, we should perform all table creations before opening the transaction. For example, we could introduce an attribute like `#[DbalSetup]` or `#[DbalInit]` on methods that would be triggered **before** Ecotone opens a transaction. By default, I would initialize event-sourced tables and all projection tables. This way, Ecotone itself does not create implicit commits; only the user would be responsible for avoiding DDL operations on MySQL inside a transaction.

WDYT?

dgafka:
@lifinsky @jlabedo the implicit commit already exists for mysql side of things, it's nothing new. You've probably have not noticed because it's only for Mysql, and will happen only for first execution when tables need to be created. 

I've just tried to work around it within this PR (for mysql). Anyways for Ecotone 2.0, I would like to drop it completely, and expose some CLI which will execute all migrations or will dump raw SQLs needed for Ecotone.

packages/Dbal/src/DbalTransaction/DbalTransactionInterceptor.php
dgafka: Like mentioned in the main comment, implicit commit is there for mysql already. It was just trying to go around the limitations it makes during retries in tests.

lifinsky:
@jlabedo I completely agree with you, but I think that such initialization, convenient for execution in migration scripts, I would suggest doing separately in the future so as not to block the most important fix for a critical situation with different connections.

packages/Dbal/src/DbalTransaction/DbalTransactionInterceptor.php
dgafka: Is there some specific code that we can depend on?

lifinsky: It’s possible to detect that an exception is caused by an implicit commit (e.g. during COMMIT or ROLLBACK after a DDL statement) by inspecting the $errorInfo array of PDOException. However, the values inside errorInfo are database-driver specific, so the error code can differ between MySQL, Oracle, etc.

A simpler and more portable approach is to just catch PDOException and analyze its message text, since the message always contains the information about implicit commit behavior regardless of the driver.

dgafka:
Ok guys, I thin we are good to go with this. 

It will fix db connections, and transaction nested in retries

PR #530, opened 2025-10-02
Fix: ignore errors while parsing array shapes

jlabedo:

`TypeResolver` is not able to parse multiline phpdoc types: it will just take the first line of type.
With the changes introduced by #523 , we can now parse array shapes, which are the most common types that are multiline, and could break some existing code.
For backward compatibility, this PR ignore malformed array shapes and returns a basic 'array' type.

@lifinsky 


lifinsky:
@dgafka 🙏

dgafka (approved):

PR #531, opened 2025-10-03
Default async config in tests

dgafka:

This ensures that default async configuration for Message Consumers is set up even for Asynchronous Projections (as currently it works only for Async Event Handlers)


jlabedo (approved):

lifinsky:
Thanks, guys!

Issue #532, opened 2025-10-04
Converter receives UuidV4 instead of string after update to 1.268.2


lifinsky:
**Ecotone version(s) affected**: latest

**Description**  
After upgrading Ecotone from 1.255 to 1.268.2, a previously working #[Converter] stopped functioning correctly.
It now receives a Symfony\Component\Uid\UuidV4 instance instead of a string, causing a TypeError.

```php
final class CardTransactionDisplayIdConverter
{
    #[Converter]
    public static function convertToString(CardTransactionDisplayId $displayId): string
    {
        return (string) $displayId;
    }

    #[Converter]
    public static function convertFromString(string $value): CardTransactionDisplayId
    {
        return CardTransactionDisplayId::create($value);
    }
}
```

```php
TypeError: 
App\Converter\ValueObject\CardTransaction\CardTransactionDisplayIdConverter::convertFromString(): 
Argument #1 ($value) must be of type string, Symfony\Component\Uid\UuidV4 given, 
called in /app/vendor/ecotone/ecotone/src/Messaging/Conversion/StaticCallConverter.php on line 23
#136 /src/App/Converter/ValueObject/CardTransaction/CardTransactionDisplayIdConverter.php(17)
#135 /vendor/ecotone/ecotone/src/Messaging/Conversion/StaticCallConverter.php(23)
#134 /vendor/ecotone/ecotone/src/Messaging/Conversion/AutoCollectionConversionService.php(64)
#133 /vendor/ecotone/ecotone/src/Messaging/MessageConverter/DefaultHeaderMapper.php(179)
#132 /vendor/ecotone/ecotone/src/Messaging/MessageConverter/DefaultHeaderMapper.php(125)
#131 /vendor/ecotone/ecotone/src/Messaging/MessageConverter/DefaultHeaderMapper.php(84)
#130 /vendor/ecotone/ecotone/src/Modelling/AggregateFlow/SaveAggregate/SaveAggregateServiceTemplate.php(153)
```

**Where it happens**

When sending a command that contains both a UuidV4 and a custom value object (CardTransactionDisplayId):

```php
$this->commandBus->send(
    new CreateSuccessfulAuthorizationDecisionCommand(
        aggregateId: new UuidV4(),
        displayId: CardTransactionDisplayId::generate(),
        // ...
    ),
    metadata: [
        'customerAccountId' => $card->customerAccount->id, // UuidV4
         // ...
    ]
);
```
[...]

PR #534, opened 2025-10-06
Fix StaticCallConverter

lifinsky:
Fixes [issue #532](https://github.com/ecotoneframework/ecotone-dev/issues/532).

## Description of Changes
The bug occurred in StaticCallConverter, where a converter parameter typed as string was incorrectly treated as compatible with Stringable. This behavior is not allowed in strict mode, since Stringable objects cannot be implicitly cast to string.

This fix ensures that type matching is handled correctly when comparing an object type against a non-object type.

### DIFF
--- a/packages/Ecotone/src/Messaging/Conversion/StaticCallConverter.php
+++ b/packages/Ecotone/src/Messaging/Conversion/StaticCallConverter.php
@@ -6,6 +6,7 @@ namespace Ecotone\Messaging\Conversion;
 
 use Ecotone\Messaging\Handler\Type;
 use Ecotone\Messaging\Handler\Type\ObjectType;
+use Ecotone\Messaging\Handler\Type\UnionType;
 
 /**
  * licence Apache-2.0
@@ -35,7 +36,10 @@ class StaticCallConverter implements Converter
             return false;
         }
 
-        if ($sourceType instanceof ObjectType && ! $this->sourceType instanceof ObjectType) {
+        if ($sourceType instanceof ObjectType
+            &&  ! $this->sourceType instanceof ObjectType
+            && ! $this->sourceType instanceof UnionType
+        ) {
             return false;
         }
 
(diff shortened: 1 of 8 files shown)


packages/Ecotone/src/Messaging/Conversion/StaticCallConverter.php
```diff
@@ -28,9 +29,21 @@ public function convert($source, Type $sourceType, MediaType $sourceMediaType, T
...
+            return false;
+        }
+
+        if ($sourceType instanceof ObjectType && ! $this->sourceType instanceof ObjectType) {
+            return false;
+        }
+
+        if ($this->targetType instanceof ObjectType && ! $targetType instanceof ObjectType) {
+            return false;
+        }
```
dgafka, line 44: Are those two if statement meant to handle not passing Stringable to string and vice versa? 

We also handle union types as UnionType, therefore I don't think instanceof usage will be safe here.

dgafka: If strict mode doesn't allow for conversion between stringable and string, then I don't think we should allow for defining equality between those in Ecotone's type system.

Therefore we could drop representation of stringable as string here:
image

@jlabedo wdyt?

dgafka: If strict mode doesn't allow for conversion between stringable and string, then I don't think we should allow for defining equality between those in Ecotone's type system.

Therefore we could drop representation of stringable as string here:
image

@jlabedo wdyt?

lifinsky: @dgafka Yes, I was thinking about that too.
But the `BuiltinType` looks like a class that’s too generic to contain fixes for the acceptType and accepts methods.

I want to point out that this strict typing issue would also occur with float/integer and string/scalar types.

lifinsky: I’d also like to note that intersection types are not supported by the current type system, which could be critical for converter target types.

lifinsky: @dgafka Updated, I suggest staying on this version until there is more time to add tests for conversion via StaticCallConverter and cover all edge cases.

jlabedo: Hey, the static converter is just a converter that do not need to be registered in the dependency container. It should have the exact same behaviour as `ReferenceServiceConverter` regarding `match` method. The fix is in #536

PR #535, opened 2025-10-06
feat: docker image upgrade

dgafka:

This upgrades docker images to use latest base PHP images and install new sockets extensions (needed for amqp lib)


lifinsky:
Maybe [php:8.4.13](https://hub.docker.com/layers/library/php/8.4.13-alpine/)?

PR #536, opened 2025-10-06
fix stringable conversion for static converters

jlabedo:
Fixes issue #532 

## Description of Changes
Allow implicit php type conversion in `StaticCallConverter.php`

### DIFF
--- a/packages/Ecotone/src/Messaging/Conversion/StaticCallConverter.php
+++ b/packages/Ecotone/src/Messaging/Conversion/StaticCallConverter.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Ecotone\Messaging\Conversion;
 
 use Ecotone\Messaging\Handler\Type;
--- a/packages/Ecotone/tests/Messaging/Unit/Conversion/ReferenceServiceConverterBuilderTest.php
+++ b/packages/Ecotone/tests/Messaging/Unit/Conversion/ReferenceServiceConverterBuilderTest.php
@@ -9,6 +9,8 @@ use Ecotone\Messaging\Attribute\Converter;
 use Ecotone\Messaging\Config\ServiceConfiguration;
 use Ecotone\Messaging\Handler\ServiceActivator\ServiceActivatorBuilder;
 use Ecotone\Messaging\Support\InvalidArgumentException;
+use Ecotone\Modelling\Attribute\CommandHandler;
+use Ecotone\Modelling\Attribute\EventHandler;
 use Ecotone\Test\ComponentTestBuilder;
 use PHPUnit\Framework\TestCase;
 use stdClass;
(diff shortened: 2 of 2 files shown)


lifinsky:
Hi! What will happen if there are two converters — one for string and another for a value object type that implements Stringable? I don’t see any prioritization logic in the code that would choose the more specific type in such a case.

lifinsky:
```php
final class UuidV4Converter
{
    #[Converter]
    public static function convertFromString(string $uuid): UuidV4
    {
        $factory = new UuidFactory();

        /** @var UuidV4 $uuidV4 */
        $uuidV4 = $factory->fromString($uuid);

        return $uuidV4;
    }

    #[Converter]
    public static function convertToString(UuidV4 $uuid): string
    {
        return $uuid->toString();
    }
}
```

@jlabedo When passing UuidV4 in a header, if we leave your code unchanged, both converter methods are triggered — which isn’t correct.

jlabedo:
Only one converter would be triggered. And yes, there is no priority logic for converters but this has always been the case.

In your example, if you are requesting an UUidV4 from a string, `public static function convertFromString(string $uuid): UuidV4` will be called (uuidV4 is also a string, but a string is not an uuidv4, so conversion is required).
If you are requesting a string and you pass an UUidV4, no converter will be called, as UUIV4 is stringable.
Does it makes sense ?

lifinsky:
```php
    public function matches(Type $sourceType, MediaType $sourceMediaType, Type $targetType, MediaType $targetMediaType): bool
    {
        if (! $sourceMediaType->isCompatibleWithParsed(MediaType::APPLICATION_X_PHP)
            || ! $targetMediaType->isCompatibleWithParsed(MediaType::APPLICATION_X_PHP)
        ) {
            return false;
        }

        if ($sourceType instanceof ObjectType
            &&  ! $this->sourceType instanceof ObjectType
            && ! $this->sourceType instanceof UnionType
        ) {
            return false;
        }

        if ($this->sourceType instanceof BuiltinType && $sourceType instanceof ObjectType) {
            return false;
        }

        if ($this->targetType instanceof ObjectType && ! $targetType instanceof ObjectType) {
            return false;
        }

        \var_dump($this->method, $this->sourceType->toString(), $this->targetType->toString());

        return $sourceType->isCompatibleWith($this->sourceType)
            && $targetType->acceptType($this->targetType);
    }
```

@jlabedo This code allows this case to work without removing strict typing in StaticCallConverter.
It uses only one converter (UuidV4 -> string) and doesn’t invoke the reverse converter when writing headers to the stream.
Those who need to work with Stringable should take responsibility for that explicitly by using a union type string|Stringable.
Additionally, the documentation should highlight that there’s no prioritization mechanism for selecting the more specific converter based on the argument or return type.

lifinsky:
So, as I understand it, you’re suggesting that we adopt your approach and remove the convertToString method in such converters when the value object already implements `Stringable`, correct?

So instead of

```php
class CardTransactionDisplayIdConverter
{
    #[Converter]
    public static function convertToString(CardTransactionDisplayId $displayId): string
    {
        return (string) $displayId;
    }

    #[Converter]
    public static function convertFromString(string $value): CardTransactionDisplayId
    {
        return CardTransactionDisplayId::create($value);
    }
}
```


You’re suggesting keeping only this:

```php
class CardTransactionDisplayIdConverter
{
    #[Converter]
    public static function convertFromString(string $value): CardTransactionDisplayId
    {
        return CardTransactionDisplayId::create($value);
    }
}
```

lifinsky:
At this point, any option works for me as long as it resolves the critical production issue with the connection that you’re aware of. However, I’m in favor of keeping strict types (Stringable is not the same as string).

lifinsky:
Thank you very much, we are waiting for the tag and will do the regression.

PR #537, opened 2025-10-06
Conversion service with strings

dgafka:

Trying to reproduce issue with string serialization

### DIFF
--- a/packages/Ecotone/tests/Messaging/Unit/Conversion/ConversionServiceTest.php
+++ b/packages/Ecotone/tests/Messaging/Unit/Conversion/ConversionServiceTest.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace Test\Ecotone\Messaging\Unit\Conversion;
 
+use Ecotone\Lite\EcotoneLite;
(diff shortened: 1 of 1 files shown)


dgafka:
@lifinsky I took example from [closed pr](https://github.com/ecotoneframework/ecotone-dev/pull/534/files#diff-a481cd5fe494444d3f4eb801d4efb317d7cb0007a12a821c2b903bb9846c873d), but it does work on master.

So I've create multiple converters for object : string, but it still passes. Can you point me, how to reproduce that?

lifinsky:
When some converter is missing, I still don’t understand why there’s an attempt to use a converter that doesn’t match the signature - we’ll look into that later when there’s a bit more time.

dgafka:
@lifinsky I got confused now. Is there an issue with wrong Converters being called, or all good? :)

lifinsky:
> @lifinsky I got confused now. Is there an issue with wrong Converters being called, or all good? :)

The new version of Ecotone is currently running in two production services. We plan to update the other services soon as well. Thank you for your support!

PR #541, opened 2025-10-17
MRE: Static converters are applied in incorrect order

lifinsky:
We created a minimal test that registers two static converters:
`ConverterOne` — a generic converter (string → `SomeStringableDataOne`)
`ConverterTwo` — a specific converter (`SomeStringableDataTwo` → string)

When Ecotone is bootstrapped with these converters in different order, conversion behavior changes — proving that converter resolution depends on registration order instead of type specificity.


jlabedo:
Fix implemented in #545

PR #543, opened 2025-10-17
Introduce AMQP Stream Support

dgafka:

This introduces amqp stream support as new Enterprise feature.
image

- Adds support for both Amqp Lib and Amqp Ext
- Adds ability to register new Streaming Channel for RabbitMQ
- Adds ability to track position for streaming channel consumption

### DIFF
--- a/packages/Dbal/src/Deduplication/DeduplicationInterceptor.php
+++ b/packages/Dbal/src/Deduplication/DeduplicationInterceptor.php
@@ -3,11 +3,11 @@
 namespace Ecotone\Dbal\Deduplication;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Types;
 use Ecotone\Dbal\DbalReconnectableConnectionFactory;
 use Ecotone\Enqueue\CachedConnectionFactory;
-use Ecotone\Laravel\Config\PDO\PostgresDriver;
 use Ecotone\Messaging\Attribute\AsynchronousRunningEndpoint;
 use Ecotone\Messaging\Attribute\Deduplicated;
 use Ecotone\Messaging\Attribute\IdentifiedAnnotation;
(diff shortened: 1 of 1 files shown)


dgafka:
This is quite big PR, and I had to add support for both Amqp lib and Amqp Ext at the same time. 

@jlabedo or @unixslayer can you take a look, if all looks right (to avoid breaking existing amqp integration)?

jlabedo:
Hey @dgafka , I am not an expert of amqp but I will definitely have a look at your PR. 
Could we use amqp streams as an event store in the future ?

packages/Ecotone/src/Messaging/Consumer/ConsumerPositionTracker.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace Ecotone\Messaging\Consumer;
+
+/**
+ * Tracks consumer position/offset for resumable message consumption
+ * licence Apache-2.0
+ */
+interface ConsumerPositionTracker
```
jlabedo: @dgafka Do you have a test showing how to use a consumer with position tracking ? A projection in the projecting namespace seems to be exactly the same as a consumer with position tracking :) Maybe we should use the same abstraction for both

dgafka: Yep, there are multiple tests, you can take a look in AmqpStreamChannelTest

packages/Amqp/src/AmqpReconnectableConnectionFactory.php
```diff
@@ -22,14 +26,18 @@
 class AmqpReconnectableConnectionFactory implements ReconnectableConnectionFactory
 {
     private string $connectionInstanceId;
-    private AmqpConnectionFactory $connectionFactory;
+    private AmqpExtConnectionFactory|AmqpLibConnectionFactory $connectionFactory;
     private ?SubscriptionConsumer $subscriptionConsumer = null;
 
-    public function __construct(AmqpConnectionFactory $connectionFactory, ?string $connectionInstanceId = null)
+    public function __construct(AmqpExtConnectionFactory|AmqpLibConnectionFactory $connectionFactory, ?string $connectionInstanceId = null, private bool $publisherAcknowledgments = false)
```
unixslayer: Have you considered typehinting `Interop\Amqp\AmqpConnectionFactory`? both `AmqpExt` and `AmqpLib` implements it.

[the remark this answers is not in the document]
dgafka: By consumer acknowledgements you mean standard message acking? 

As this is only for publish confirms, standard consumer message acking is not configurable as internal Ecotone process.

PR #545, opened 2025-10-19
Fix conversion handling

jlabedo:
#526 introduced a regression in how converters are resolved

## Description of Changes
The target type of a custom converter should match exactly the requested type (or a be an exact member of an union type).

### DIFF
--- a/packages/Ecotone/tests/Messaging/Unit/Conversion/ConversionServiceTest.php
+++ b/packages/Ecotone/tests/Messaging/Unit/Conversion/ConversionServiceTest.php
@@ -210,6 +210,63 @@ class ConversionServiceTest extends TestCase
             )
         );
     }
+
+    /**
+     * @dataProvider provideConvertors
+     */
+    public function test_it_wrong_conversion_calling(array $convertors): void
+    {
+        $ecotone = EcotoneLite::bootstrapFlowTesting(
+            classesToResolve: $convertors,
+            containerOrAvailableServices: [],
+        );
+
+        /** @var ConversionService $conversionService */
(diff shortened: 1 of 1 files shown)


packages/Ecotone/tests/Messaging/Unit/Conversion/ConversionServiceTest.php
```diff
@@ -210,6 +210,63 @@ classesToResolve: [
...
+
+        $converterTwo = new class () {
+            #[Converter]
+            public static function convertToString(SomeStringableDataTwo $value): string
+            {
+                return $value->value;
+            }
+        };
+
+        yield 'Passed' => [[
```
lifinsky: maybe Forward order, Reverse order?

packages/Ecotone/src/Messaging/Conversion/CustomConverter.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * @inheritDoc
+     */
+    public function matches(Type $sourceType, MediaType $sourceMediaType, Type $targetType, MediaType $targetMediaType): bool
+    {
+        return $sourceMediaType->isCompatibleWithParsed(MediaType::APPLICATION_X_PHP)
+            && $targetMediaType->isCompatibleWithParsed(MediaType::APPLICATION_X_PHP)
+            && $sourceType->isCompatibleWith($this->sourceType)
+            && (
+                $targetType->equals($this->targetType)
```
dgafka: Do I understand correctly that we solve this by expecting now equal, meaning Converter have to explictly state what it does convert?

Issue #409, opened 2024-11-13
Automatic entity manager flush method call during ES projection reset


lifinsky:
**Description**  
For some projections, it is convenient to use the doctrine entity manager instead of low-level inserts or updates (using db connection). But in this case, without calling the flush method, the projection can not be restored after resey, but calling flush explicitly duplicates the transaction block during the synchronous operation of the projection during the handler commands, especially since there can be several synchronous projections for one event.

**Example**  
```php
   /**
     * @throws OptimisticLockException
     * @throws ORMException
     */
    #[EventHandler]
    public function onCardWasClosed(CardWasClosed $event): void
    {
        /** @var Card $card */
        $card = $this->entityManager->find(Card::class, (string) $event->id);
        $card->markAsClosed($event->closedAt());
        $this->entityManager->flush(); // redundant during handler commands
    }
```

dgafka:
For ProjectionV2 doctrine flush will automatically hook after each batch.
The ObjectManagerInterceptor automatically calls flush() + clear() on the Doctrine EntityManager after each projection batch.                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
  The flow is:                                                                                                                                                                                                                                          
  1. ProjectingManager processes events in a batch (within a transaction)            
  2. After the batch, it calls projectorExecutor->flush()                
  3. This triggers the #[ProjectionFlush] method on the projection
  4. The ObjectManagerInterceptor wraps that call, doing flush() + clear() on all registered ObjectManagers

PR #548, opened 2025-10-23
Projecting: flush method and performance improvements

jlabedo:

- Introduce a new attribute `ProjectionFlush` that is triggered before projection manager commits the state. This is handy when using doctrine ORM or any other sort of query batching inside projections, drastically reducing DB roundtrips.
- EventStoreGlobalStreamSource build its own optimized queries to event store tables. This is a major performance improvement.
- Minor hot paths performance improvements on `GatewayHeaderConverter`, `HeaderConverter` and `EcotoneSpanBuilder`

By introducing those changes and query batching on user side, I managed to achieve **20x projection speed**.

### DIFF
--- a/packages/PdoEventSourcing/tests/Projecting/ProophIntegrationTest.php
+++ b/packages/PdoEventSourcing/tests/Projecting/ProophIntegrationTest.php
@@ -447,7 +447,7 @@ class ProophIntegrationTest extends ProjectingTestCase
             ->sendQueryWithRouting('getTicketsCount');
 
         self::assertSame(5, $ticketsCount, 'Batch projection should process all events in batches');
-        self::assertSame(4, $projection->flushCallCount, 'Flush should be called 4 times (10 events / batch size 3)');
+        self::assertSame(4, $projection->flushCallCount, 'Flush should be called 4 times (10 events / batch size 3) = 4 rounded up');
     }
 
     public function test_it_handles_custom_name_stream_source(): void


lifinsky:
@jlabedo Hi!
Will this improvement resolve this issue?
 https://github.com/ecotoneframework/ecotone-dev/issues/409

packages/PdoEventSourcing/src/Projecting/StreamSource/EventStoreGlobalStreamSource.php
```diff
@@ -7,69 +7,75 @@
...
+
     public function __construct(
-        private EventStore      $eventStore,
+        DbalConnectionFactory|ManagerRegistryConnectionFactory  $connectionFactory,
         private EcotoneClockInterface  $clock,
         private string          $streamName,
         private int             $maxGapOffset = 5_000,
         private ?Duration       $gapTimeout = null,
     ) {
+        $this->proophStreamTable = '_' . \sha1($streamName);
```
dgafka: Can we extract building stream table name to some shared method?

packages/Ecotone/src/Projecting/ProjectingManager.php
```diff
@@ -45,6 +45,8 @@ public function execute(?string $partitionKey = null, bool $manualInitialization
                 foreach ($streamPage->events as $event) {
                     $userState = $this->projectorExecutor->project($event, $userState);
                 }
+                $this->projectorExecutor->flush();
```
dgafka: I don't see a test case for this. 
Can we add Ecotone Lite which will ensure that this is called after each batch (and with different batch sizes)

packages/PdoEventSourcing/src/Projecting/StreamSource/EventStoreGlobalStreamSource.php
```diff
@@ -7,69 +7,75 @@
...
+        $query = $this->connection->executeQuery(<<<SQL
+            SELECT no, event_name, payload, metadata, created_at
+                FROM {$this->proophStreamTable}
+                WHERE no > :position {$gapQueryPart}
+            ORDER BY no
+            LIMIT {$count}
+            SQL, [
+            'position' => $tracking->getPosition(),
+            ...$gapQueryPartParams
+        ], $gapQueryPartParamTypes);
```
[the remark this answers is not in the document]
jlabedo: Agree. I hesitated, but my rule of thumb is to avoid adding abstractions when in doubt, as anticipating abstractions leads to bad abstractions :) Maybe we could had it when the new projections are more mature ?

jlabedo:
> @jlabedo Hi! Will this improvement resolve this issue? #409
> 
> Will this also improve the projections related to the Prooph event store?

Hello @lifinsky , this won't affect existing projections, this is for a work in progress non documented and unstable feature on a new projecting system :)

packages/PdoEventSourcing/src/Projecting/StreamSource/EventStoreGlobalStreamSourceBuilder.php
```diff
@@ -34,12 +35,17 @@ public function compile(MessagingContainerBuilder $builder): Definition|Referenc
...
                 Reference::to(EcotoneClockInterface::class),
-                $this->streamName,
+                self::getProophTableName($this->streamName),
                 5_000,
                 new Definition(Duration::class, [60], 'seconds'),
             ],
         );
     }
+
+    public static function getProophTableName($streamName): string
```
dgafka: Can we add test case that is does work with [custom stream name](https://docs.ecotone.tech/modelling/event-sourcing/event-sourcing-introduction/persistence-strategy/making-stream-immune-to-changes#custom-stream-name)?

jlabedo: done

packages/Ecotone/src/Projecting/ProjectingManager.php
jlabedo: I've added a `ProjectionBatchSize` attribute to configure the batch size (it was hard coded), and added a test

packages/Ecotone/src/Messaging/Scheduling/DatePoint.php
```diff
@@ -42,7 +42,7 @@ public function setUnixTime(Duration $duration): static
     /**
      * @throws DateMalformedStringException When $format or $datetime are invalid
      */
-    public static function createFromFormat(string $format, string $datetime, ?DateTimeZone $timezone = null): static
+    public static function createFromFormat(string $format, string $datetime, ?DateTimeZone $timezone = null): static|false
```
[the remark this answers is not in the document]
jlabedo: done, we throw an exception.

packages/Ecotone/src/Projecting/Config/ProjectingConsoleCommands.php
```diff
@@ -37,6 +38,7 @@ public function initProjection(?string $name = null, #[ConsoleParameterOption] b
     }
 
     #[ConsoleCommand('ecotone:projection:backfill')]
+    #[WithoutDbalTransaction]
```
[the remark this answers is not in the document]
dgafka: Make sense

packages/PdoEventSourcing/tests/Projecting/ProophIntegrationTest.php
```diff
@@ -402,4 +410,71 @@ public function init(): void
...
+        $basketProjection = new #[Projection(self::NAME), FromStream(Basket::BASKET_STREAM)] class {
+            public const NAME = 'basket_projection';
+            public int $basketCount = 0;
+
+            #[EventHandler(BasketWasCreated::EVENT_NAME)]
+            public function onBasketCreated(): void
+            {
+                $this->basketCount++;
+            }
+        };
```
dgafka: This is so satisfying, seeing how easy is to set up those tests with anoynomous classes combined with attributes

packages/PdoEventSourcing/tests/Projecting/ProophIntegrationTest.php
```diff
@@ -402,4 +404,43 @@ public function init(): void
...
+            $ecotone->sendCommand(new CreateTicketCommand($ticketId))
+                ->sendCommandWithRoutingKey(Ticket::ASSIGN_COMMAND, metadata: ['aggregate.id' => $ticketId]);
+        }
+
+        // Trigger projection processing
+        $ticketsCount = $ecotone->triggerProjection($projection::NAME)
+            ->sendQueryWithRouting('getTicketsCount');
+
+        self::assertSame(5, $ticketsCount, 'Batch projection should process all events in batches');
+        self::assertSame(4, $projection->flushCallCount, 'Flush should be called 5 times (10 events / batch size 3)');
```
dgafka: > Flush should be called 4 times (10 events / batch size 3)

nit

packages/PdoEventSourcing/tests/Projecting/ProophIntegrationTest.php
dgafka: Would have to see, how such configuration builder would look like. 
But what I can imagine for sure is automatic DI set up, so we don't need to deal with wiring Services :)

PR #569, opened 2025-12-02
Disable Signal Handling During Test Execution to Prevent SIGINT Interference

jlabedo:
When running PHPUnit, Ecotone’s `SignalInterceptor` enables `pcntl_async_signals()` and installs a custom SIGINT handler.  
While this is correct for long-running asynchronous workers, it causes issues during test execution:

- PHPUnit relies on the default SIGINT behavior to stop execution with **Ctrl+C**
- Ecotone’s signal handling intercepts SIGINT, preventing PHPUnit from terminating
- After running a single test involving running asynchronous queues, **Ctrl+C no longer stops the test run**, forcing manual process termination

To provide a smooth developer experience and avoid interfering with test runners, signal handling should be disabled when Ecotone is executed in a test environment.

## Description of Changes

- Added a new `onShutdown` method for consumer interceptors, providing a way for `SignalInterceptor` to restore previous handlers when consumer stops
- Added new option in ExecutionPollingMetadata to disable signals handling, set by default to `disabled` with default testing setup (`ExecutionPollingMetadata::withTestingSetup`)
- Fully backward compatible; no behavior changes in production or worker contexts

### DIFF
--- a/packages/Ecotone/src/Messaging/Endpoint/ExecutionPollingMetadata.php
+++ b/packages/Ecotone/src/Messaging/Endpoint/ExecutionPollingMetadata.php
@@ -14,7 +14,6 @@ final class ExecutionPollingMetadata
     private ?bool $stopOnError = null;
     private ?bool $finishWhenNoMessages = null;
     private ?int $executionAmountLimit = null;
-    private ?bool $signalInterceptorsEnabledIfAvailable = null;
 
     private function __construct()
     {
@@ -82,7 +81,6 @@ final class ExecutionPollingMetadata
     public function withTestingSetup(int $amountOfMessagesToHandle = 1, int $maxExecutionTimeInMilliseconds = 100, bool $failAtError = true): self
     {
         return $this
-            ->withSignalInterceptorIfAvailable(false)
             ->withHandledMessageLimit($amountOfMessagesToHandle)
             ->withStopOnError($failAtError)
             ->withExecutionTimeLimitInMilliseconds($maxExecutionTimeInMilliseconds);
@@ -120,14 +118,6 @@ final class ExecutionPollingMetadata
         return $self;
     }
 
-    public function withSignalInterceptorIfAvailable(bool $value = true): self
-    {
-        $self = clone $this;
-        $self->signalInterceptorsEnabledIfAvailable = $value;
-
-        return $self;
-    }
-
     public function getCron(): ?string
     {
         return $this->cron;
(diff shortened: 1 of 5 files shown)


dgafka (dismissed):

packages/Ecotone/src/Messaging/Endpoint/Interceptor/SignalInterceptor.php
```diff
@@ -36,57 +36,48 @@ public function onStartup(): void
...
-    /**
-     * @inheritDoc
-     */
-    public function preRun(): void
+    public function onShutdown(): void
     {
-    }
+        if ($this->beforeHandlers !== null) {
+            foreach ($this->beforeHandlers as $signal => $handler) {
+                pcntl_signal($signal, $handler);
```
dgafka: If I understand correctly this will trigger stopConsumer method, which will then mark `shouldBeStopped=true`, but at the point of shutdown it's already closed, so it won't have any effect?

jlabedo: No. Stopconsumer method may have been called before, or any other interceptor may have asked the consumer to stop.
Here we are in consumer shutdown phase. we are just restoring previous signal handlers.

packages/Ecotone/src/Messaging/Endpoint/PollingConsumer/InterceptedConsumer.php
```diff
@@ -85,6 +85,10 @@ public function run(): void
                 [],
                 OnConsumerStop::CONSUMER_STOP_CHANNEL_NAME
             );
+
+            foreach ($this->consumerInterceptors as $consumerInterceptor) {
+                $consumerInterceptor->onShutdown();
+            }
```
jlabedo: This is where the shutdown phase is called @dgafka

packages/Ecotone/src/Messaging/Endpoint/ExecutionPollingMetadata.php
```diff
@@ -118,6 +120,14 @@ public function withExecutionAmountLimit(int $limit): self
...
     }
 
+    public function withSignalInterceptorIfAvailable(bool $value = true): self
+    {
+        $self = clone $this;
+        $self->signalInterceptorsEnabledIfAvailable = $value;
+
+        return $self;
+    }
+
```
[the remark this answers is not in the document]
dgafka: Ok, let's remove

PR #573, opened 2025-12-05
Multiple message handlers

unixslayer:

Consider following scenario. 

- When product becomes billable (`ProductBecameBillable`), it should be added to `InvoiceOrder` identified by `Debtor` of billable product. `InvoiceOrder` can be created when this occurs.
- `Debtor` of a product can change overtime (`ProductDebtorWasChanged`). When it happens, following outcome should take place:
- - product should be removed from `InvoiceOrde` identified by old `Debtor`
- - product should be added to `InvoiceOrder` identified by new `Debtor`. `InvoiceOrder` of new `Debtor` may not yet exists, therefore it can be created when this occurs.

Above scenario require aggregate to have multiple factory methods and action channels:

`ProductBecameBillable` - factory method of `InvoiceOrder`
`ProductDebtorWasChanged` - factory method + two action channels, one to remove and second to add product to `InvoiceOrder`

Proper identifier mapping is defined in `EventHandler` definition as shown in `packages/Ecotone/tests/Modelling/Fixture/MultipleMessageHandlers/InvoiceOrder.php`

However, current implementation require that Message Handlers on Aggregate and Saga can be used either for single factory method and single action method together, or for multiple actions methods and forbid that implementation.

## Description of Changes

tbd

### DIFF
--- a/packages/Ecotone/tests/Modelling/Fixture/MultipleMessageHandlers/InvoiceOrder.php
+++ b/packages/Ecotone/tests/Modelling/Fixture/MultipleMessageHandlers/InvoiceOrder.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Test\Ecotone\Modelling\Fixture\MultipleMessageHandlers;
+
+use Ecotone\Messaging\Attribute\Parameter\Payload;
+use Ecotone\Modelling\Attribute\EventHandler;
+use Ecotone\Modelling\Attribute\Identifier;
+use Ecotone\Modelling\Attribute\Saga;
+
+#[Saga]
+class InvoiceOrder
+{
+    private array $products;
+
+    public function __construct(
+        #[Identifier] private string $debtorId,
+    ) {
+        $this->products = [];
+    }
+
+    public function products(): array
+    {
+        return $this->products;
+    }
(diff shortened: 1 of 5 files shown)


packages/Ecotone/tests/Modelling/Fixture/MultipleMessageHandlers/InvoiceOrder.php
```diff
@@ -0,0 +1,61 @@
...
+        #[Payload] ProductBecameBillable $event
+    ): self {
+        $invoiceOrder = new self($event->debtor);
+        $invoiceOrder->products[] = $event->productId;
+
+        return $invoiceOrder;
+    }
+
+    #[EventHandler(endpointId: 'createWhenDebtorWasChanged', identifierMapping: ['debtorId' => 'payload.newDebtor'])]
+    public static function createWhenDebtorWasChanged(
```
dgafka: If we would go this path, then there should be still guard that there is only one factory:action for given identifier mapping (either payload or headers). 
Otherwise it become confusing whatever they do relate to each other or not, and then it's not obvious if we should create instance or redirect to action.

PR #586, opened 2025-12-17
Feat/new projecting system dx

dgafka:

This is next step into making new projecting system open to wide public.
New projecting system provides much more advanced features, and we need to come with Developer Experience, that will make it intent driven and easy to follow. Therefore this PR, introduces high level usage for new projecting system, that is based on declarative configuration using attributes.

### Global tracking Projection (default)

```php
#[ProjectionV2]
class CountTicketsProjections
```

### Partitioned tracking Projection 

```php
#[Partitioned]
#[ProjectionV2]
class TicketsProjections
```

### Polling Projection

```php
#[Polling("tickets_projection")]
#[ProjectionV2]
class TicketsProjections
```

### Streaming Projection

```php
#[Streaming("streaming_channel")]
#[ProjectionV2]
class TicketsProjections
```

### DIFF
--- a/packages/PdoEventSourcing/tests/EventSourcingMessagingTestCase.php
+++ b/packages/PdoEventSourcing/tests/EventSourcingMessagingTestCase.php
@@ -17,6 +17,7 @@ abstract class EventSourcingMessagingTestCase extends TestCase
 {
     private ConnectionFactory $tenantAConnection;
     private ConnectionFactory $tenantBConnection;
+    private static ?DbalConnectionFactory $defaultConnection = null;
 
     protected static function getSchemaManager(Connection $connection): \Doctrine\DBAL\Schema\AbstractSchemaManager
     {
@@ -28,6 +29,12 @@ abstract class EventSourcingMessagingTestCase extends TestCase
     {
         self::clearDataTables($this->connectionForTenantA()->createContext()->getDbalConnection());
         self::clearDataTables($this->connectionForTenantB()->createContext()->getDbalConnection());
+    }
+
+    protected function tearDown(): void
+    {
+        $this->connectionForTenantA()->createContext()->getDbalConnection()->close();
+        $this->connectionForTenantB()->createContext()->getDbalConnection()->close();
     }
 
     protected function connectionForTenantB(): ConnectionFactory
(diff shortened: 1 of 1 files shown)


packages/Ecotone/src/Projecting/Attribute/PartitionedProjection.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+#[Attribute(Attribute::TARGET_CLASS)]
+class PartitionedProjection extends Projection
+{
+    public function __construct(
+        string $name,
+        string $partitionHeaderName = MessageHeaders::EVENT_AGGREGATE_ID,
+        string $runningMode = self::RUNNING_MODE_EVENT_DRIVEN,
+        ?string $endpointId = null,
+        ?string $streamingChannelName = null,
```
dgafka: That's for streaming channels, I may think how to extract that. For now it works as configurable option when message channel is passed here

dgafka:
@jlabedo looks good? :)

packages/PdoEventSourcing/tests/Projecting/Global/GapDetectionInSynchronousProjectionTest.php
```diff
@@ -0,0 +1,37 @@
...
+ * Original tests covered:
+ * - test_detecting_gaps_without_detection_window
+ * - test_detecting_gaps_with_detection_window
+ * - test_detecting_gaps_without_gap_detection
+ */
+final class GapDetectionInSynchronousProjectionTest extends TestCase
+{
+    public function test_skipped_gap_detection_handled_differently_in_new_system(): void
+    {
+        $this->markTestSkipped(
```
jlabedo: Useless ai ?

dgafka:
> 3. Could misguide users
Users will ask: “Should I use Global or Partitioned?”
That’s the wrong question.
The real question is: “What is the projection’s consistency and isolation boundary?”.

Hmm, ye global or partitioned are from solution space, not problem space. 
Even so I agree, my concern here is different. 

1. Assuming that people have not much knowledge, if that will be simple config change within same attribute, then it may create feeling that Ecotone's internals will take care of the switch. 
In reality that's not what going to happen, each one has separate tracking system, therefore projection will be considered fresh (and will start from the beginning).

2. What about streaming projection, the tracking there is no concern of Ecotone, as it will be part of the Streaming platform itself (partitions on Kafka streaming channel). It will also require passing the streaming channel name to the Projection (you can take a look now on Partitioned attribute as I've not extracted separate StreamingProjection attribute). 
So having single attribute creates design that leads to confusion, because we are exposing configs that are not part of given Projection type strategy (autoinitialization is not needed for partitioned, streaming channel is only needed for streaming projections). So even so we can guard that and disallow, that just shows that the design is simply wrong and now we need to compensate to avoid wrong set ups (with friendly design, wrong set ups are not even possible).

PR #615, opened 2026-01-15
data protection

unixslayer:

Sensitive data should be obfuscated when leaving application via transport channels.

## Description of Changes

- use `Ecotone\DataProtection\Attribute\Sensitive` to define messages with sensitive data
- attribute can be defined with `message`, `endpoint`, or globally for `channel`
- define encryption keys with `Ecotone\DataProtection\Configuration\DataProtectionConfiguration`
- sensitive data will be encrypted right before its sended to queue and decrypted right after message is being retrieved from queue
- data protection require JMSConverter module to be enabled
- whole message payload will be encrypted with defined headers

### DIFF
--- a/packages/DataProtection/composer.json
+++ b/packages/DataProtection/composer.json
@@ -43,7 +43,7 @@
     },
     "scripts": {
         "tests:phpstan": "vendor/bin/phpstan",
-        "tests:phpunit": "vendor/bin/phpunit",
+        "tests:phpunit": "vendor/bin/phpunit --no-coverage --testdox",
         "tests:ci": [
             "@tests:phpstan",
             "@tests:phpunit"
@@ -51,10 +51,10 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-main": "1.62-dev"
+            "dev-main": "1.298.0-dev"
         },
         "ecotone": {
-            "repository": "DataProtection"
+            "repository": "data-protection"
         },
         "merge-plugin": {
             "include": [
--- a/packages/DataProtection/tests/Integration/ObfuscateChannelTest.php
+++ b/packages/DataProtection/tests/Integration/ObfuscateChannelTest.php
@@ -24,7 +24,7 @@ use Test\Ecotone\DataProtection\Fixture\ObfuscateChannel\TestEventHandler;
 use Test\Ecotone\DataProtection\Fixture\SomeMessage;
 use Test\Ecotone\DataProtection\Fixture\TestClass;
 use Test\Ecotone\DataProtection\Fixture\TestEnum;
-use Test\Ecotone\Messaging\Unit\Channel\TestQueueChannel;
+use Test\Ecotone\DataProtection\TestQueueChannel;
 
 /**
  * @internal
--- a/packages/DataProtection/tests/Integration/ObfuscateEndpointsTest.php
+++ b/packages/DataProtection/tests/Integration/ObfuscateEndpointsTest.php
@@ -30,11 +30,10 @@ use Test\Ecotone\DataProtection\Fixture\ObfuscateEndpoints\EventHandlerWithAnnot
 use Test\Ecotone\DataProtection\Fixture\ObfuscateEndpoints\EventHandlerWithAnnotatedEndpointWithSecondaryEncryptionKey;
 use Test\Ecotone\DataProtection\Fixture\ObfuscateEndpoints\EventHandlerWithAnnotatedMethodWithoutPayload;
 use Test\Ecotone\DataProtection\Fixture\ObfuscateEndpoints\EventHandlerWithAnnotatedPayloadAndHeader;
-use Test\Ecotone\DataProtection\Fixture\ObfuscateEndpoints\NoPollableEventHandler;
 use Test\Ecotone\DataProtection\Fixture\SomeMessage;
 use Test\Ecotone\DataProtection\Fixture\TestClass;
 use Test\Ecotone\DataProtection\Fixture\TestEnum;
-use Test\Ecotone\Messaging\Unit\Channel\TestQueueChannel;
+use Test\Ecotone\DataProtection\TestQueueChannel;
 
 /**
  * @internal
--- a/packages/DataProtection/tests/Integration/ObfuscateMessagesTest.php
+++ b/packages/DataProtection/tests/Integration/ObfuscateMessagesTest.php
@@ -24,7 +24,7 @@ use Test\Ecotone\DataProtection\Fixture\ObfuscateMessages\TestCommandHandler;
 use Test\Ecotone\DataProtection\Fixture\ObfuscateMessages\TestEventHandler;
 use Test\Ecotone\DataProtection\Fixture\TestClass;
 use Test\Ecotone\DataProtection\Fixture\TestEnum;
-use Test\Ecotone\Messaging\Unit\Channel\TestQueueChannel;
+use Test\Ecotone\DataProtection\TestQueueChannel;
 
 /**
  * @internal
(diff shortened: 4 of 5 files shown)


packages/DataProtection/src/Configuration/DataProtectionModule.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        $obfuscators = [];
+
+        $messagesUsingSensitiveData = $annotationRegistrationService->findAnnotatedClasses(UsingSensitiveData::class);
+
+        foreach ($messagesUsingSensitiveData as $messageUsingSensitiveData) {
+            /** @var UsingSensitiveData $attribute */
+            $usingSensitiveDataAttribute = $annotationRegistrationService->getAttributeForClass($messageUsingSensitiveData, UsingSensitiveData::class);
+
+            $reflectionClass = new \ReflectionClass($messageUsingSensitiveData);
```
dgafka: You can use `$interfaceToCallRegistry->getClassDefinitionFor()`, this is cached way of resolving classes :)

packages/DataProtection/src/Configuration/DataProtectionModule.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $pollableMessageChannels = ExtensionObjectResolver::resolve(MessageChannelWithSerializationBuilder::class, $extensionObjects);
+
+        foreach ($pollableMessageChannels as $pollableMessageChannel) {
+            $messagingConfiguration->registerChannelInterceptor(
+                new OutboundEncryptionChannelBuilder($pollableMessageChannel->getMessageChannelName())
+            );
+            $messagingConfiguration->registerChannelInterceptor(
+                new OutboundDecryptionChannelBuilder($pollableMessageChannel->getMessageChannelName())
+            );
```
dgafka: What was the reason to do it via `Channel Interceptors`, rather than hook in into `Converter mechanism`?

packages/DataProtection/tests/Integration/ObfuscateAnnotatedMessagesTest.php
```diff
@@ -0,0 +1,166 @@
...
+        $channelMessage = $channel->getLastSentMessage();
+        $messagePayload = Crypto::decrypt(base64_decode($channelMessage->getPayload()), $this->primaryKey);
+        $messageHeaders = $channelMessage->getHeaders();
+
+        self::assertEquals('{"class":{"argument":"value","enum":"first"},"enum":"first","argument":"value"}', $messagePayload);
+        self::assertEquals('secret-value', Crypto::decrypt(base64_decode($messageHeaders->get('foo')), $this->primaryKey));
+        self::assertEquals('even-more-secret-value', Crypto::decrypt(base64_decode($messageHeaders->get('bar')), $this->primaryKey));
+        self::assertEquals('non-sensitive-value', $messageHeaders->get('baz'));
+    }
+
```
dgafka: Can we add a test case where we encrypt only headers without payload?

packages/DataProtection/tests/Integration/ObfuscateChannelTest.php
```diff
@@ -0,0 +1,300 @@
...
+        $channelMessage = $channel->getLastSentMessage();
+        $messagePayload = Crypto::decrypt(base64_decode($channelMessage->getPayload()), $this->primaryKey);
+        $messageHeaders = $channelMessage->getHeaders();
+
+        self::assertEquals('{"class":{"argument":"value","enum":"first"},"enum":"first","argument":"value"}', $messagePayload);
+        self::assertEquals($metadataSent['foo'], Crypto::decrypt(base64_decode($messageHeaders->get('foo')), $this->primaryKey));
+        self::assertEquals($metadataSent['bar'], Crypto::decrypt(base64_decode($messageHeaders->get('bar')), $this->primaryKey));
+        self::assertEquals($metadataSent['baz'], $messageHeaders->get('baz'));
+    }
+
```
dgafka: Can we add a test case where we encrypt only headers without payload?

dgafka (commented):
Just some more coverage comments and potential API improvements. 

I do think we can start with this and iterate over that (e.g. discuss how we want to handle Event Streams) :)

packages/DataProtection/src/OutboundEncryptionChannelInterceptor.php
```diff
@@ -0,0 +1,76 @@
...
+            return null;
+        }
+
+        if ($messageObfuscator = $this->findMessageObfuscator($message)) {
+            return $messageObfuscator->encrypt($message);
+        }
+
+        if ($routingObfuscator = $this->findRoutingObfuscator($message)) {
+            return $routingObfuscator->encrypt($message);
+        }
```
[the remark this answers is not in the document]
unixslayer: There is a test that asserts that message annotation will always take precedence.

packages/DataProtection/composer.json
```diff
@@ -0,0 +1,81 @@
...
+            "Test\\Ecotone\\DataProtection\\": [
+                "tests"
+            ]
+        }
+    },
+    "require": {
+        "ext-openssl": "*",
+        "ecotone/ecotone": "~1.298.0",
+        "ecotone/jms-converter": "~1.298.0",
+        "defuse/php-encryption": "^2.4"
```
dgafka: Image

This one seems to be abandoned. Is there some other lib we can rely on?

packages/DataProtection/src/Attribute/Sensitive.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+/**
+ * licence Enterprise
+ */
+namespace Ecotone\DataProtection\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_CLASS | Attribute::TARGET_PARAMETER)]
```
dgafka:
```suggestion
Attribute::TARGET_CLASS | Attribute::TARGET_PARAMETER)]
```

As far as I see method and parameter are interchangeable. 
I think we can drop it from method level, keeping one sensible api on the parameter level. This will be in touch with other features that do follow this practice

packages/DataProtection/src/Configuration/DataProtectionModule.php
```diff
@@ -0,0 +1,214 @@
...
+            $messageObfuscatorReferences[$messageClass] = Reference::to($id);
+        }
+
+        foreach (ExtensionObjectResolver::resolve(MessageChannelWithSerializationBuilder::class, $extensionObjects) as $pollableMessageChannel) {
+            if (! $pollableMessageChannel->isPollable()) {
+                continue;
+            }
+
+            $messagingConfiguration->registerChannelInterceptor(
```
dgafka: Can we only register if encryption is meant to be used? Otherwise each polling channel will be decorated in the call stack

packages/DataProtection/src/Attribute/Sensitive.php
unixslayer: resolved in #625

packages/DataProtection/src/Obfuscator/Obfuscator.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+namespace Ecotone\DataProtection\Obfuscator;
+
+use Defuse\Crypto\Crypto;
+use Defuse\Crypto\Key;
+use Ecotone\Messaging\Message;
+use Ecotone\Messaging\Support\Assert;
+use Ecotone\Messaging\Support\MessageBuilder;
+
+readonly class Obfuscator
```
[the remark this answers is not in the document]
unixslayer: @dgafka in follow up PR I've dropped obfuscation related naming in favour of emphasizing on Message Encryption #626

PR #616, opened 2026-01-16
feat(amqp): Allow configurable delay strategy for backed channel

bakhritdinov:
feat(Amqp): Allow custom DelayStrategy in AmqpBackedMessageChannel

This commit introduces the ability to provide a custom `DelayStrategy` for AMQP-backed message channels.

The `AmqpBackedMessageChannelBuilder` now includes a `withDelayStrategy` method, which passes the strategy down to the `AmqpOutboundChannelAdapter`. This allows users to extend or modify the default delaying behavior.

An integration test has been added to verify that the custom strategy is correctly wired and invoked.

Previously, the `AmqpOutboundChannelAdapter` utilized `HeadersExchangeDelayStrategy` as the standard implementation for message delaying.

This change introduces a configurable way for users to provide their own `DelayStrategy` implementation, increasing the flexibility of the component. This is beneficial for scenarios requiring custom delay logic, such as integrating with specific AMQP delay plugins or implementing specific business rules for message redelivery.

## Description of Changes

- **`packages/Amqp/src/AmqpOutboundChannelAdapter.php`**
  The constructor was updated to accept an optional `?DelayStrategy $delayStrategy = null` parameter. The `handle` method now utilizes this injected `delayStrategy` if provided, while maintaining `HeadersExchangeDelayStrategy` as the default behavior to ensure backward compatibility.

- **`packages/Amqp/src/AmqpOutboundChannelAdapterBuilder.php`**
  A new fluent method, `withDelayStrategy(DelayStrategy $delayStrategy): self`, was added to store the user-defined `DelayStrategy`. The `compile` method was updated to pass this strategy to the `AmqpOutboundChannelAdapter` definition.

- **`packages/Amqp/src/AmqpBackedMessageChannelBuilder.php`**
  A new fluent method, `withDelayStrategy(\Enqueue\AmqpTools\DelayStrategy $delayStrategy): self`, was added. This delegates the provided strategy to the underlying `AmqpOutboundChannelAdapterBuilder`, completing the configuration chain.

- **`packages/Amqp/tests/Fixture/AmqpChannel/SpyDelayStrategy.php` (New File)**
  A helper class `SpyDelayStrategy` was created for testing purposes. It implements `Enqueue\AmqpTools\DelayStrategy` and tracks method invocations while delegating to the default strategy to ensure correct behavior during tests.

- **`packages/Amqp/tests/Integration/AmqpChannelAdapterTest.php`**
  A new integration test `test_using_custom_delay_strategy_from_channel_builder()` was added. It registers a `SpyDelayStrategy`, configures the channel to use it, and verifies that the custom strategy is correctly invoked within the Ecotone container.

- [x] I have read and agree to the contribution terms outlined in CONTRIBUTING.


packages/Amqp/src/AmqpOutboundChannelAdapterBuilder.php
```diff
@@ -140,6 +149,7 @@ public function compile(MessagingContainerBuilder $builder): Definition
             $outboundMessageConverter,
             new Reference(ConversionService::REFERENCE_NAME),
             Reference::to(AmqpTransactionInterceptor::class),
+            $this->delayStrategy
```
dgafka: This is an object, and the Definition is actually an description on how it should be registered in Dependency Container.

Can we then pass to reference name:

```php
AmqpBackedMessageChannelBuilder::withDelayStrategy(string $delayStrategyReferenceName)
```

Reference name would be id from container, most likely class name.

So in here we would have `Reference::to($this->delayStrategyReferenceName)` instead of `$this->delayStrategy`.  It will be then fetch your Delay Strategy from Dependency Container,, making it possible to define this class once and reuse between different Message Channels. 

Let me know, if make sense :)

bakhritdinov: Yes, of course. I’ve made the changes.

PR #618, opened 2026-01-17
feat(scheduling): Sync Static Global Clock with Clock Service (Quick Solution)

jb-needhelp:
Automatically synchronize `Clock::$globalClock` with the injected clock instance, eliminating the need for manual `Clock::set()` calls in test environments.

### Problem
When using a custom clock (e.g., a static clock for testing) with `\Ecotone\Messaging\Scheduling\Clock`, the global clock remains unsynchronized. This causes incorrect timestamps in message headers created via `MessageHeaders::createMessageHeadersWith()`.
Currently, the workaround requires calling `Clock::set()` explicitly in each test case, which leaks internal implementation details into application tests.
### Solution
Set `$globalClock` automatically when a `Clock` instance is constructed:

If an internal clock is provided → use it as the global clock
Otherwise → default to NativeClock

This allows applications to define their own `PsrClockInterface` implementation without manual global clock management.

### Cleaner Solution
A cleaner long-term solution would be removing the deprecated `Clock::get()` method entirely and replace it with `Clock`instance. However, this creates a cascade of changes, particularly around `DelayableQueueChannel` which calls `MessageBuilder::fromMessage($message)` without access to a `Clock` instance.

## Description of Changes

* **`Ecotone\Messaging\Scheduling\Clock`**: Auto-set `$globalClock` on construction; added `resetGlobalClock()` for test teardown
* **`Test\Ecotone\Messaging\Unit\Scheduling\StaticGlobalClockTest`**: Verify expected behavior


packages/Ecotone/tests/Messaging/Unit/Scheduling/StaticGlobalClockTest.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        $staticClock = $this->createStaticClock('2025-08-11 16:00:00');
+        $clock = new Clock($staticClock);
+
+        $globalClock = Clock::get();
+
+        $this->assertInstanceOf(Clock::class, $globalClock);
+        $this->assertEquals('2025-08-11 16:00:00', $globalClock->now()->format('Y-m-d H:i:s'));
+    }
+
```
dgafka: Can you provide test case with delay message using new clock strategy?

So similar test [to this](https://docs.ecotone.tech/modelling/testing-support/testing-asynchronous-messaging#testing-delayed-messages), but by changing time using clock

jb-needhelp: Added this separate test: `DelayedMessageAgainstGlobalClockTest`

PR #684, opened 2026-07-18
feat: asynchronous message publishing with batching and pre-commit delivery guarantees (Enterprise)

dgafka:

Publishing to a message broker blocks the handler on one full round-trip per message — a
handler emitting many messages pays N sequential broker round-trips, which caps publishing
throughput as systems scale.

Ecotone now provides an Enterprise feature that makes publishing scale, while still keeping
all the delivery guarantees end users rely on: every message is still confirmed by the
broker, a failed delivery still fails the business operation or is routed to the error
channel, and every failure points to the exact message that failed — nothing is ever lost
silently.

This is achieved by firing messages to the broker immediately, batching them natively per
provider, and deferring confirmation awaiting to the last responsible moment — messages
travel to the broker while the handler keeps working, and all confirmations are collected
before the operation completes.

### Benchmarks

Publishing 1,000 messages, relative to each provider's previous message-by-message
synchronous publishing (= 100%). Mode of 10 iterations, all delivery confirmations awaited:

| Provider | Feature used                          | Time vs previous | Speed-up |
|----------|---------------------------------------|------------------|----------|
| Kafka    | Batching + async sending              | 9%               | ~11×     |
| SQS      | Batching + async sending              | 9%               | ~11×     |
| DBAL     | Batching                              | 10%              | ~10×     |
| AMQP     | Batching + async sending              | 32%              | ~3×      |
| Redis    | Batching                              | 36%              | ~3×      |
| Kafka    | Async sending (single messages)       | 36%              | ~3×      |
| SQS      | Async sending (single messages)       | 65%              | ~1.5×    |

Batching + async sending publishes the workload as provider-native batches with
confirmation awaiting deferred; async sending fires individual messages without waiting.
DBAL and Redis write synchronously, so batching is their applicable feature.

### Resulting flow

```mermaid
graph TD
    A[Handler emits messages] --> B[Collector gathers them]
    B -->|one BatchMessage| C[Adapter: provider-native batch publish, no wait]
    C --> D[Registry: PendingDelivery per publish]
    D --> E[Waiter interceptor: await all confirmations]
    E -->|all confirmed| F[Operation completes]
    E -->|some failed| G[Only failed messages to error channel, or operation fails]
```

### Enabling

```php
    ->withExtensionObjects([
        // Message Channel: messages published from handlers are gathered,
        // sent as one batch and confirmed before the operation completes
        AmqpBackedMessageChannelBuilder::create('orders')
            ->withAsyncPublishing(),
        KafkaMessageChannelBuilder::create('orders')
            ->withAsyncPublishing(timeoutInMilliseconds: 12000),

        // Message Publisher: enables MessagePublisher::asyncPublish()
        AmqpMessagePublisherConfiguration::create()
            ->withAsyncPublishing(),
    ]);
```

### Example

```php
#[CommandHandler]
public function place(PlaceOrder $command, EventBus $eventBus): void
{
    // events are batched to the broker and their confirmations
    // awaited automatically before the operation completes
    $eventBus->publish(new OrderWasPlaced($command->orderId));
}

// explicit async publishing with a Future
$future = $messagePublisher->asyncPublish(
    BatchMessage::constructEmpty()
        ->append($firstOrder)
        ->append($secondOrder)
);

$future->resolve(); // throws AsyncPublishingFailedException listing only the messages that failed
```


packages/Sqs/src/SqsOutboundChannelAdapter.php
```diff
@@ -34,4 +58,131 @@ public function initialize(): void
...
+     */
+    private function buildBatchRequest(SqsDestination $destination, string $queueUrl, array $entries, array $trackedMessages): array
+    {
+        $arguments = [
+            '@region' => $destination->getRegion(),
+            'QueueUrl' => $queueUrl,
+            'Entries' => $entries,
+        ];
+
+        $arguments['@http'] = ['timeout' => $this->asyncPublishingTimeout / 1000];
```
dgafka, line 184: When async publishing is disabled, every send — including plain single-message OSS sends — now goes through `sendMessageBatchAsync` with a hard 12s `@http` timeout (set here unconditionally from `asyncPublishingTimeout`), and failures surface as `AsyncPublishingFailedException` instead of the previous Interop/AWS exceptions. That changes behavior for existing flows: a slow SQS endpoint/proxy taking >12s previously succeeded with SDK-default timeouts, and existing `catch` blocks keyed on the old exception types no longer match.

Proposal: keep the previous synchronous `getProducer()->send()` path when `asyncPublishing` is false, and use the batch/async machinery (including this timeout) only when async publishing is enabled — so current flows stay untouched.

packages/Ecotone/src/Messaging/Channel/PollableChannel/SendRetries/SendRetryChannelInterceptor.php
```diff
@@ -65,7 +71,7 @@ public function afterSendCompletion(Message $message, MessageChannel $messageCha
 
         if ($this->deadLetterChannel !== null) {
             $this->errorChannelService->handle(
```
dgafka: When retries exhaust and a `deadLetterChannel` is configured, `$messageToRedeliver` can still carry a `BatchMessage` payload here — `ErrorChannelService::handle` will run `OutboundMessageConverter::prepare` on the internal wrapper and store one batch blob as a single dead letter. That differs from `AsyncPublishingWaiterInterceptor::unpackFailedMessages`, which unpacks entries into individual messages before routing to the error channel. The new `SendRetryOfFailedBatchDeliveriesTest` constructs the interceptor with `deadLetterChannel: null`, so this path is currently untested.

Proposal: add an EcotoneLite test scenario covering dead-letter storage of failed batch deliveries for both paths — async (waiter → error channel) and sync fallback (send retries → dead letter) — asserting per-message granularity in the dead letter store.

packages/Amqp/src/AmqpOutboundChannelAdapter.php
```diff
@@ -79,27 +251,87 @@ public function handle(Message $message): void
...
-        if ($this->publisherConfirms && ! $this->amqpTransactionInterceptor->isRunningInTransaction()) {
+    /**
+     * @param array<int, array{message: Message, deliveryTag: int, correlationId: string}> $publishRecords
+     */
+    private function registerPendingDelivery(array $publishRecords): void
+    {
+        $this->asyncPublishingRegistry->register(
+            $this->channelName,
+            new AmqpPendingDelivery(
+                $this->connectionFactory->createContext(),
```
dgafka: The publish operation re-derives the context from the cached factory at several points (`recordPublishedMessage()`, here, and `awaitPublisherConfirmsSynchronously()`), while the actual publishes happened on the context resolved at the start of `publishMessages()`.

Failure scenario: the connection drops between publish and this call. `createContext()` then reconnects, and `AmqpReconnectableConnectionFactory::createContext()` calls `AmqpPublisherConfirmations::reset()`, bumping the epoch. `AmqpPendingDelivery` captures the epoch **in its constructor — i.e. post-reset** — so its stale-channel detection (`getEpoch() !== $this->confirmationsEpoch`) never fires. The old delivery tags (1..N) can then be settled by confirmations of *new* messages published on the fresh channel as the watermark advances: false success, message lost silently — the exact case the feature promises can't happen.

Proposed fix: resolve `$context` and capture `$confirmations->getEpoch()` once at the top of `handle()`, *before* the first publish, and thread both through `publishMessages()` / `recordPublishedMessage()` / `registerPendingDelivery()` / `awaitPublisherConfirmsSynchronously()`. `AmqpPendingDelivery` should receive the pre-publish epoch as a constructor argument instead of reading it from the confirmations object. A mid-operation reconnect is then detected and reported as a connection-reset failure instead of a false success. This also removes the repeated `createContext()` instanceof check per published record in `recordPublishedMessage()`.

packages/Ecotone/src/Messaging/BatchMessage.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public static function constructEmpty(): self
+    {
+        return new self();
+    }
+
+    /**
+     * @param array<string, mixed> $headers
+     */
+    public function append(mixed $payload, array $headers = []): self
```
dgafka, line 40: `append()` mutates `$entries` *and* returns `$this`, while core callers reassign as if it were immutable (`$batchMessage = $batchMessage->append(...)` in `CollectorSenderInterceptor::combineIntoBatch()` and `SendRetryChannelInterceptor::messageContainingOnlyFailedDeliveries()`). The contract is ambiguous, and a caller keeping a reference can mutate a batch after it has already been sent.

Proposal — make it genuinely immutable:

```php
public function append(mixed $payload, array $headers = []): self
{
    $appended = clone $this;
    $appended->entries[] = ['payload' => $payload, 'headers' => $headers];

    return $appended;
}
```

Chained construction (`constructEmpty()->append()->append()`) keeps working unchanged. Since clone-per-append copies the entries array each time (O(n²) for large batches), also add a bulk constructor for the hot paths:

```php
/** @param array<int, array{payload: mixed, headers: array<string, mixed>}> $entries */
public static function fromEntries(array $entries): self
```

and use it in `combineIntoBatch()` / `messageContainingOnlyFailedDeliveries()`, keeping `append()` for user-facing incremental construction.

packages/Redis/src/RedisOutboundChannelAdapter.php
```diff
@@ -33,4 +67,85 @@ public function initialize(): void
...
+            return;
+        }
+
+        $arguments = [count($immediatePayloads), ...$immediatePayloads];
+        foreach ($delayedEntries as $delayedEntry) {
+            $arguments[] = $delayedEntry['score'];
+            $arguments[] = $delayedEntry['payload'];
+        }
+
+        $pushedMessages = $context->getRedis()->eval(
```
dgafka: Redis is the one provider where "every failure points to the exact message that failed" doesn't hold. If a `redis.call` errors after k of N pushes, the script aborts but the k earlier LPUSH/ZADDs persist (Lua scripts don't roll back), and the return value — the pushed count — is lost. The adapter then throws a plain `RuntimeException`, so `SendRetryChannelInterceptor` has no `AsyncPublishingFailedException` with failed deliveries to trim by and retries the **whole** batch: entries 1..k get duplicated.

Proposed fix: wrap the pushes in `pcall` and, on first error, return the count pushed so far, e.g. `return {pushed, errorMessage}`. Entries are pushed in order, so `pushed = k` identifies exactly which entries made it. The adapter can then throw `AsyncPublishingFailedException` containing only entries k+1..N, and retry / error-channel routing handles just those — same semantics as the other providers.

Proposed test scenario to lock it in (unit-level, faking the client behind `$context->getRedis()`): publish a batch of 3 where `eval` reports 1 pushed and then fails; assert entries 2–3 (and only those) are retried / routed to the error channel, and entry 1 is not published twice. An integration variant can assert the base guarantee independent of the fix: a mid-batch failure must never report success.

PR #686, opened 2026-07-21
Fix tenant-aware connection factory lookup in HeaderBasedMultiTenantConnectionFactory

ralfkuehnel:

Pass the resolved tenant to `getConnectionFactory()` so the connection factory is looked up using the correct tenant context instead of relying solely on the current execution context.

## Description of Changes

Passing the tenant through ensures consistent tenant resolution and correct connection factory selection.


dgafka:
Hey @ralfkuehnel,

Can you add test case using Ecotone Lite, which would fail without this fix? (to avoid regression bugs in future)

dgafka:
Do you need any help from my side @ralfkuehnel ?

PR #689, opened 2026-07-23
Fix scalar message header conversion to nullable enums

lifinsky:
## Summary

Fix scalar message header conversion for nullable enum handler parameters.

## Problem

When a message handler declares a nullable enum header parameter:

```php
#[Header('deliveryMethod')]
?DeliveryMethod $deliveryMethod = null
```

Ecotone represents the target type as the `DeliveryMethod|null` union.

`ScalarToEnumConverter` correctly recognizes the target as an enum-compatible type, but then passes the complete union type string to `ReflectionEnum`. This results in the following error for a non-null scalar header value:

```text
Class "DeliveryMethod|null" does not exist
```

A missing header or a header with a `null` value is not affected because scalar-to-enum conversion is not required in that case.

## Fix

When the target is a union type, remove `null` before creating `ReflectionEnum` and converting the scalar value to the enum case.

## Commit Structure

1. Add a regression test that reproduces the failure without the fix.
2. Add the minimal fix to `ScalarToEnumConverter`.

## Verification

Before the fix, the regression test fails with:

```text
Class "DeliveryMethod|null" does not exist
```

After the fix:

```text
OK (1 test, 1 assertion)
```

The existing Core test suite also passes.

### DIFF
--- a/packages/Ecotone/src/Messaging/Conversion/ScalarToEnum/ScalarToEnumConverter.php
+++ b/packages/Ecotone/src/Messaging/Conversion/ScalarToEnum/ScalarToEnumConverter.php
@@ -19,6 +19,10 @@ class ScalarToEnumConverter implements Converter
      */
     public function convert($source, Type $sourceType, MediaType $sourceMediaType, Type $targetType, MediaType $targetMediaType)
     {
+        if ($targetType instanceof Type\UnionType) {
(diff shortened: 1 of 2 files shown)


lifinsky:
@dgafka Hi, I need your help :)

dgafka (approved):
Make sense, thanks for the fix :)

PR #692, opened 2026-08-04
feat: batched outbox forwarding for combined message channels (Enterprise)

dgafka:
### Why

An outbox in front of a broker channel relays one message per poll cycle, so a single consumer run barely moves a busy queue, and a deep backlog makes each cycle progressively slower. This change turns the outbox into a true relay: an opt-in publishing process drains messages in batches straight from the database and pushes them forward in bulk, without consuming them through a message channel or deserializing payloads. Relaying 10 000 messages into a durable Dbal target drops from ~82s to ~1s (and to ~0.35s as a single in-memory-bound batch). Delivery stays at-least-once: a failed delivery is released for redelivery without duplicating already-delivered messages, and connection failures roll the whole cycle back for a clean retry.

### Resulting flow

```mermaid
graph TD
    A[Polling tick of batch publishing endpoint] --> B[Claim up to batch size rows per outbox channel<br/>FOR UPDATE SKIP LOCKED on PostgreSQL, claim markers elsewhere]
    B --> C[Group rows by routing slip target, payloads stay serialized]
    C --> D{Target supports<br/>high-throughput publishing?}
    D -- yes --> E[Single BatchMessage per target, one multi-row insert]
    D -- no --> F[Per-message publish]
    E --> G[Delete delivered rows, commit per-channel transaction]
    F --> G
    G --> H{Delivery failed?}
    H -- release/ignore strategy --> I[Release only failed rows for redelivery, or drop on ignore]
    H -- stop or connection failure --> J[Rollback restores claims instantly, clean retry]
```

### Out of scope

Batch forwarding for non-Dbal outbox sources — configuring it for any other channel fails at compile time.

### Example

```php
#[ServiceContext]
public function channels(): array
{
    return [
        OutboxForwardingMessageChannel::create('orders', DbalBackedMessageChannelBuilder::create('outbox'), 'orderProcessing')
            ->withMaxForwardingBatchSize(100),
        AmqpBackedMessageChannelBuilder::create('orderProcessing')
            ->withHighThroughputPublishing(),
    ];
}
```

`OutboxForwardingMessageChannel` extends `CombinedMessageChannel` and is the whole definition of the relay: exactly one Dbal backed source (the builder may be embedded, registering the outbox channel along the way and making a non database source unrepresentable) and one target channel. Reusing the outbox inside a plain `CombinedMessageChannel`, pointing an endpoint or output channel at it, or configuring it without a Dbal source all fail at compile time; forwarding channels sharing one outbox must agree on batch size, endpoint id and failure strategy.

One process can relay multiple outboxes — including ones in different databases — by sharing an endpoint id: `OutboxForwardingMessageChannel::create('orders', 'ordersOutbox', 'ordersTarget')->withEndpointId('outboxPublisher')` for each flow, then running the `outboxPublisher` consumer.

With multi-tenancy each tenant already owns an outbox in its own database; the publisher reuses polling consumer tenant propagation, so every run drains the next tenant's outbox in round robin and forwards to that tenant's target — no extra configuration beyond the standard `MultiTenantConfiguration`.

### Failure handling per strategy

The strategy applied to failed deliveries is configured on the forwarding itself; when not set, it inherits from the outbox channel (default `RESEND`):

```php
OutboxForwardingMessageChannel::create('orders', 'outbox', 'orderProcessing')
    ->withFinalFailureStrategy(FinalFailureStrategy::STOP);
```

| Strategy | Behaviour when delivery of a message fails |
|---|---|
| `RESEND` (default) / `RELEASE` | Only the failed rows return to the outbox marked as redelivered and are retried on the next run; successfully delivered rows of the same batch stay delivered — no duplicates |
| `IGNORE` | Failed rows are dropped from the outbox; the rest of the batch stays delivered |
| `STOP` | The exception propagates and the publisher stops; the batch transaction rolls back, so every row returns to the outbox — messages already handed to the target within that batch may be delivered again after restart (at-least-once) |
| any + connection failure | The cycle always rolls back and claims restore instantly regardless of the configured strategy, so a restarted process retries the full batch cleanly |

The error channel is not involved in relay failures — the outbox itself is the retry store, so failed messages never leave the delivery guarantee for a dead-end.

### Benchmark results

**Draining a 10 000 message outbox** (PostgreSQL source, warmed consumer, phpbench mode; all broker targets receive whole batches via high-throughput publishing):

| Scenario | Total time | Per message | Speedup |
|---|---|---|---|
| Message by message (without licence) | 76.6 s | ~7.7 ms | 1× |

| Scenario | Total time | Per message | Speedup |
|---|---|---|---|
| Batched, size 100 → in-memory target | 0.83 s | ~0.083 ms | ~92× |
| Batched, single batch of 10 000 → in-memory target | 0.44 s | ~0.044 ms | ~174× |
| Batched, size 100 → **Dbal** target | 0.91 s | ~0.091 ms | ~84× |
| Batched, size 100 → **RabbitMQ** target | 0.87 s | ~0.087 ms | ~88× |
| Batched, size 100 → **Kafka** target | 0.58 s | ~0.058 ms | ~132× |
| Batched, size 100 → **Redis** target | 0.59 s | ~0.059 ms | ~129× |
| Batched, size 100 → **SQS** target | 3.01 s | ~0.30 ms | ~25× |

Every broker receives 10 000 durable messages in under a second except SQS, which is bound by its API's 10-message batch limit (~1000 HTTP round trips). The message-by-message path also degrades as the backlog deepens (per-message cost grew from ~2.4 ms at 200 messages to ~7.7 ms at 10 000), while the batched path improves with volume as fixed per-run overhead amortizes. Batch size is the throughput knob: one large batch outperforms a hundred 100-row cycles at the cost of holding the whole batch in memory within one transaction.

## Description of Changes

- New opt-in `OutboxForwardingMessageChannel` (Enterprise) — a Combined Message Channel of exactly one Dbal backed outbox and one target — replaces the outbox channel's consumer with a standalone polling endpoint executing SQL directly: claim, group by routing-slip target, publish, delete/release — all inside an explicit per-channel transaction; rows relay in wire format, so serialization happens once on the producing side and format conversion belongs to the target channel
- Compile-time safety: outbox forwarding requires an Enterprise licence; a non-Dbal or unknown source, reuse of the outbox inside a plain Combined Message Channel, usage as execution or output channel, and conflicting settings between flows sharing an outbox all fail bootstrap
- Failure handling per configurable final failure strategy (`withFinalFailureStrategy` on the forwarding configuration, inheriting from the outbox channel when not set): release only failed rows, ignore drops them, stop and connection failures roll back the transaction so claims restore instantly; crashed processes' claims are swept for redelivery
- Channel-level `withAsyncPublishing()` renamed to `withHighThroughputPublishing()` across DBAL, AMQP, Kafka, Redis and SQS channel builders; `finalFailureStrategy` is now correctly passed to DBAL, SQS and Redis inbound message converters (pre-existing gap, AMQP already did)
- Cross-cutting performance: header mapping short-circuits scalars and match-all mappings, batch entries in wire format skip intermediate message construction on the outbound path, and the Dbal batch insert binds parameters directly without the doctrine type registry
- Multi-tenant setups publish in round robin across tenant outboxes per run, routing each batch to the drained tenant's target; the publishing endpoint carries `WithoutDatabaseTransaction`, now honoured consistently by the transaction, object manager and deduplication interceptors (also removing a per-tick deduplication insert every relay paid)
- Warmed 10k-message benchmark comparing message-by-message against batched draining, including batch-size and durable-target variants


packages/Ecotone/src/Messaging/Handler/Bridge/BatchForwardingBridge.php
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /**
+     * @return Message[]
+     */
+    private function drainSourceChannel(): array
+    {
+        $drainedMessages = [];
+        while (count($drainedMessages) < $this->maxBatchSize - 1) {
+            $nextMessage = $this->sourceChannel->receive();
```
dgafka: **Blocking** — the drain loop blocks for the full receive timeout on every cycle that does not fill the batch.

`$this->sourceChannel->receive()` carries no polling metadata, so `EnqueueMessageChannel::receive()` falls back to `PollingMetadata::create('enqueue')` (execution time limit `0`) -> `EnqueueInboundChannelAdapter` uses `receiveTimeoutInMilliseconds`, which defaults to `EnqueueInboundChannelAdapterBuilder::DEFAULT_RECEIVE_TIMEOUT = 10000`. `ConsumerPollingTrait::receive()` then spins with 1s sleeps until that timeout expires.

Concretely: outbox holds 3 messages, batch size 100 -> poll 1, drain 2, and the third drain call waits **10 seconds** (~10 extra SELECTs against the enqueue table, inside the open outbox transaction) before any of the 3 is published. "Fewer messages waiting than the batch size" is the steady state of a healthy relay, so this is the normal path, not the tail. Before this change the cycle published and committed immediately.

It also means `ExecutionPollingMetadata::maxExecutionTimeInMilliseconds` no longer bounds a cycle — the new tests pass `5000`, yet each drain-to-empty still waits 10s.

Direction: drain with a no-wait / zero-timeout receive and stop on the first `null`.

Reproduction to add to `CombinedChannelBatchForwardingTest`:

```php
public function test_single_message_is_forwarded_without_waiting_for_source_receive_timeout(): void
{
    $messaging = EcotoneLite::bootstrapFlowTesting(
        [$orderService::class],
        [DbalConnectionFactory::class => $this->getConnectionFactory(), $orderService],
            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE, ModulePackageList::DBAL_PACKAGE]))
            ->withExtensionObjects([
                CombinedMessageChannel::create('orders', ['outbox', 'orderProcessing']),
                DbalBackedMessageChannelBuilder::create('outbox')->withReceiveTimeout(3000),
                DbalBackedMessageChannelBuilder::create('orderProcessing'),
            ]),
        licenceKey: LicenceTesting::VALID_LICENCE,
    );

    $messaging->sendCommandWithRoutingKey('order.register', 'espresso');

    $startedAt = microtime(true);
    $messaging->run('outbox', ExecutionPollingMetadata::createWithTestingSetup(maxExecutionTimeInMilliseconds: 10000));
    $elapsedInMilliseconds = (microtime(true) - $startedAt) * 1000;

    $this->assertCount(1, $this->receiveAllFrom($messaging->getMessageChannel('orderProcessing')));
    $this->assertLessThan(3000, $elapsedInMilliseconds);
}
```

Today the run spends the full 3000ms `withReceiveTimeout` inside the drain loop while already holding the message.

packages/Ecotone/src/Messaging/Handler/Bridge/BatchForwardingBridge.php
```diff
@@ -0,0 +1,270 @@
...
+            if ($groupedMessage !== $polledMessage) {
+                $this->acknowledge($groupedMessage);
+            }
+        }
+    }
+
+    /**
+     * @param Message[] $failedMessages
+     */
+    private function releaseFailedDelivery(array $failedMessages, Message $polledMessage, string $targetChannelName, Throwable $exception): void
```
dgafka: **Blocking** — drained messages lose the whole `AcknowledgeConfirmationInterceptor` safety net.

That interceptor honours `FinalFailureStrategy` (`IGNORE` -> reject, `RELEASE`, `RESEND`, `STOP`), `RejectMessageException`, `pollingMetadata->isStoppedOnError()`, and retries the acknowledgement itself. Here the only outcome is `release()`, which is `reject($message, requeue: true)` — no backoff, no attempt counter.

Failure scenario: source channel configured with `withFinalFailureStrategy(FinalFailureStrategy::IGNORE)` plus a poison message the target channel rejects on send. Arriving as the **polled** message it is discarded and the relay moves on (unchanged behaviour). Arriving as a **drained** message it is released, re-drained and re-released every cycle, forever. Same message, opposite outcome depending only on its position in the batch. `STOP` is equally ineffective — the consumer no longer stops.

Direction: let the bridge honour the configured strategy for drained messages the same way the interceptor does. If that is not reasonably reachable from here, then make `RESEND` a **requirement** when batched forwarding is enabled and fail configuration for any other strategy, rather than silently degrading it.

Reproduction to add (needs an always-failing variant of `FailOnceOnPayloadChannelInterceptor`, so the poison message is the drained one, not the polled one):

```php
public function test_drained_message_honours_configured_final_failure_strategy(): void
{
    $messaging = EcotoneLite::bootstrapFlowTesting(
        [$orderService::class],
        [
            DbalConnectionFactory::class => $this->getConnectionFactory(),
            $orderService,
            'alwaysFailingDelivery' => new AlwaysFailOnPayloadChannelInterceptor('cappuccino'),
        ],
            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE, ModulePackageList::DBAL_PACKAGE]))
            ->withExtensionObjects([
                CombinedMessageChannel::create('orders', ['outbox', 'orderProcessing']),
                DbalBackedMessageChannelBuilder::create('outbox')
                    ->withFinalFailureStrategy(FinalFailureStrategy::IGNORE),
                DbalBackedMessageChannelBuilder::create('orderProcessing'),
                SimpleChannelInterceptorBuilder::create('orderProcessing', 'alwaysFailingDelivery'),
            ]),
        licenceKey: LicenceTesting::VALID_LICENCE,
    );

    $messaging->sendCommandWithRoutingKey('order.register', 'espresso');
    $messaging->sendCommandWithRoutingKey('order.register', 'latte');
    $messaging->sendCommandWithRoutingKey('order.register', 'cappuccino');

    $messaging->run('outbox', ExecutionPollingMetadata::createWithTestingSetup(maxExecutionTimeInMilliseconds: 5000));

    $this->assertSame(['espresso', 'latte'], $this->payloadsOf($this->receiveAllFrom($messaging->getMessageChannel('orderProcessing'))));
    $this->assertNull($messaging->getMessageChannel('outbox')->receive());
}
```

Today the last assertion fails: `cappuccino` is released instead of discarded, and stays on the outbox for every subsequent cycle. A sibling test pinning `FinalFailureStrategy::STOP` (consumer stops, message left on the outbox) is worth adding alongside it.

packages/Ecotone/src/Messaging/Channel/Collector/MessageCollectorChannelInterceptor.php
```diff
@@ -23,6 +25,12 @@ public function __construct(
 
     public function preSend(Message $message, MessageChannel $messageChannel): ?Message
     {
+        if ($message->getHeaders()->containsKey(MessageHeaders::COLLECTOR_BYPASS)) {
```
dgafka: **Should-fix** — the per-message bypass header is the wrong lever; the collector should simply be disabled for relay channels.

Two problems with the header approach:

1. **It leaks whenever the collector is not registered.** `CollectorModule` only registers `MessageCollectorChannelInterceptor` when the channel's collector is enabled — `GlobalPollableChannelConfiguration::withCollector(false)` or per-channel `PollableChannelConfiguration::withCollector(false)` turn it off. With the collector off nothing strips `collectorBypass`, so it is published to the broker. It is also absent from `MessageHeaders::getFrameworksHeaderNames()`, so `unsetFrameworkKeys()` will not remove it either. Downstream, a service that consumes and re-publishes that message would silently bypass **its own** collector.
2. **It widens the publish-before-commit window.** `CollectorStorage`'s stated purpose is to "delay sending messages so it's done just before transaction is committed". With the bypass, up to 100 messages are published at the *start* of the cycle, then draining and acking run, then commit. Any rollback after the first publish (ack failure, deadlock on the outbox table, DB error) duplicates every already-published message. Still at-least-once, but a measurable increase in duplicate rate — and the PR description frames connection-failure retry as "clean", which reads as duplicate-free.

Direction: ensure the collector is **disabled** for the relay channel at configuration time, rather than setting and stripping a header per message. Then no header exists to leak, and the intent is expressed once in configuration instead of on every send.

Reproduction to add:

```php
public function test_forwarded_message_does_not_carry_internal_collector_header(): void
{
    foreach ([true, false] as $collectorEnabled) {
        $messaging = EcotoneLite::bootstrapFlowTesting(
            [$orderService::class],
            [DbalConnectionFactory::class => $this->getConnectionFactory(), $orderService],
                ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE, ModulePackageList::DBAL_PACKAGE]))
                ->withExtensionObjects([
                    CombinedMessageChannel::create('orders', ['outbox', 'orderProcessing']),
                    DbalBackedMessageChannelBuilder::create('outbox'),
                    DbalBackedMessageChannelBuilder::create('orderProcessing'),
                    GlobalPollableChannelConfiguration::createWithDefaults()->withCollector($collectorEnabled),
                ]),
            licenceKey: LicenceTesting::VALID_LICENCE,
        );

        $messaging->sendCommandWithRoutingKey('order.register', 'espresso');
        $messaging->sendCommandWithRoutingKey('order.register', 'latte');

        $messaging->run('outbox', ExecutionPollingMetadata::createWithTestingSetup(maxExecutionTimeInMilliseconds: 5000));

        foreach ($this->receiveAllFrom($messaging->getMessageChannel('orderProcessing')) as $forwardedMessage) {
            $this->assertFalse($forwardedMessage->getHeaders()->containsKey(MessageHeaders::COLLECTOR_BYPASS));
        }
    }
}
```

Today the `withCollector(false)` pass fails — every forwarded message still carries `collectorBypass`.

packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
```diff
@@ -536,7 +541,17 @@ private function configureAsynchronousEndpoints(InterfaceToCallRegistry $interfa
...
              */
-            $this->messageHandlerBuilders[$asynchronousChannel] = BridgeBuilder::create()
+            $channelBuilder = $this->channelBuilders[$asynchronousChannel];
+            $this->messageHandlerBuilders[$asynchronousChannel] = ServiceActivatorBuilder::createWithDefinition(
+                new Definition(BatchForwardingBridge::class, [
+                    new ChannelReference($asynchronousChannel),
+                    new Reference(ChannelResolver::class),
+                    new Reference(LoggingGateway::class),
+                    $this->isRunningForEnterpriseLicence,
+                    $channelBuilder instanceof ForwardingBatchSizeAware ? $channelBuilder->getMaxForwardingBatchSize() : self::DEFAULT_FORWARDING_BATCH_SIZE,
```
dgafka: **Should-fix** — batched forwarding is switched on far more widely than the scenario it was built for.

The `: self::DEFAULT_FORWARDING_BATCH_SIZE` fallback gives batching (at 100) to **every** pollable channel builder that does not implement `ForwardingBatchSizeAware` — in-memory queue channels, `DynamicMessageChannelBuilder`, Kafka channels, third-party builders. Kafka is the sharpest case: it is pollable, it does get batch forwarding, but it has no `withMaxForwardingBatchSize()` to turn it down, so the PR description's "configurable max forwarding batch size on message channel builders" is not true for it.

Direction: gate this to the scenario it was designed and benchmarked for — a combined channel whose source is a DBAL outbox forwarding into a message broker channel. There is no reason to enable it for other channel types, and every one of them is an untested surface today.

Reproduction to add:

```php
public function test_batched_forwarding_does_not_apply_to_non_outbox_source_channels(): void
{
    $messaging = EcotoneLite::bootstrapFlowTesting(
        [$orderService::class],
        [$orderService],
            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE]))
            ->withExtensionObjects([
                CombinedMessageChannel::create('orders', ['inMemoryOutbox', 'inMemoryProcessing']),
                SimpleMessageChannelBuilder::createQueueChannel('inMemoryOutbox'),
                SimpleMessageChannelBuilder::createQueueChannel('inMemoryProcessing'),
            ]),
        licenceKey: LicenceTesting::VALID_LICENCE,
    );

    $messaging->sendCommandWithRoutingKey('order.register', 'espresso');
    $messaging->sendCommandWithRoutingKey('order.register', 'latte');
    $messaging->sendCommandWithRoutingKey('order.register', 'cappuccino');

    $messaging->run('inMemoryOutbox', ExecutionPollingMetadata::createWithTestingSetup());

    $this->assertCount(1, $this->receiveAllFrom($messaging->getMessageChannel('inMemoryProcessing')));
    $this->assertCount(2, $this->receiveAllFrom($messaging->getMessageChannel('inMemoryOutbox')));
}
```

Today all 3 are forwarded in one run. A Kafka-source equivalent is worth adding under `packages/Kafka/tests` for the same assertion.

packages/Enqueue/src/EnqueueMessageChannelBuilder.php
```diff
@@ -45,6 +49,19 @@ public function isStreamingChannel(): bool
         return false;
     }
 
+    public function withMaxForwardingBatchSize(int $maxForwardingBatchSize): self
```
dgafka: **Should-fix** — this puts batched forwarding into the shared Enqueue/DBAL channel builder, which is what makes findings 1, 5, 8 and 9 possible at all.

Because the knob lives on the builder that every DBAL/AMQP/Redis/SQS channel inherits from, batching leaks into channels that were never meant to relay: plain `#[Asynchronous]` DBAL channels, and non-auto-acked sources such as `SimpleMessageChannelBuilder::createQueueChannel(isAutoAcked: false)`. In the latter case `BatchForwardingBridge::acknowledge()` silently no-ops, so drained messages are published but never removed from the source — guaranteed duplicates on the next cycle, with no signal.

Direction: introduce a **dedicated high-throughput inbound channel adapter** rather than extending the existing DBAL/Enqueue one. Register it only for combined channels, adapt it fully (batch receive, acknowledgement and failure semantics owned in one place), and revert the default DBAL channel behaviour to what it was. That keeps the fast path opt-in and confined, instead of changing a builder that thousands of existing non-relay channels already use.

Reproductions to add:

**(a) plain DBAL async channel is untouched under an Enterprise licence**

```php
public function test_plain_asynchronous_dbal_channel_keeps_one_message_per_handled_message(): void
{
    $messaging = EcotoneLite::bootstrapFlowTesting(
        [$orderService::class],
        [DbalConnectionFactory::class => $this->getConnectionFactory(), $orderService],
            ->withSkippedModulePackageNames(ModulePackageList::allPackagesExcept([ModulePackageList::ASYNCHRONOUS_PACKAGE, ModulePackageList::DBAL_PACKAGE]))
            ->withExtensionObjects([
                DbalBackedMessageChannelBuilder::create('orders'),
            ]),
        licenceKey: LicenceTesting::VALID_LICENCE,
    );

    $messaging->sendCommandWithRoutingKey('order.register', 'espresso');
    $messaging->sendCommandWithRoutingKey('order.register', 'latte');

    $messaging->run('orders', ExecutionPollingMetadata::createWithTestingSetup(amountOfMessagesToHandle: 1));

    $this->assertSame(['espresso'], $messaging->sendQueryWithRouting('order.getRegistered'));
}
```

**(b) non-auto-acked source is never silently drained**

```php
public function test_non_auto_acked_source_does_not_duplicate_forwarded_messages(): void
{
    // CombinedMessageChannel::create('orders', ['outbox', 'orderProcessing'])
    // SimpleMessageChannelBuilder::createQueueChannel('outbox', isAutoAcked: false)
    // -> run twice, assert each payload reaches 'orderProcessing' exactly once
}
```

**(c) failure raised from the adapter, not from `preSend`**

`FailOnceOnPayloadChannelInterceptor` throws in `preSend`, i.e. before the channel's own send, so the current failure tests never exercise the real outbound path (including how `ConnectionException` is actually detected in `EnqueueInboundChannelAdapter`). Worth a variant where the target channel's `send()` itself throws mid-batch — `FailingPollableChannel` already does this for the whole channel and only needs a payload-scoped version.

Today (a) forwards nothing extra only by accident of the routing-slip target not being pollable; (b) publishes and then re-publishes on the next run.

dgafka: Addressed in 40d3d097 and ebddf028:

- **Confined to combined channels**: the forwarding bridge is now registered only for channels that are a non-final hop of a combined channel chain; every other pollable channel got the plain `Bridge` back, so plain `#[Asynchronous]` DBAL channels are structurally untouched — proven by `test_plain_asynchronous_dbal_channel_keeps_one_message_per_handled_message`.
- **Knob moved off the shared builder**: `withMaxForwardingBatchSize()` was removed from `EnqueueMessageChannelBuilder` (and the `ForwardingBatchSizeAware` interface deleted); it now lives on `CombinedMessageChannel` itself and reaches the bridge via a registered `CombinedChannelForwardingConfiguration`.
- **(b) non-auto-acked sources are never drained**: the bridge falls back to one-message-per-cycle when the polled message's acknowledgement is not auto-acked — `test_non_auto_acked_source_is_not_drained` covers it, so silent publish-without-ack duplicates cannot occur.
- **(c) failure raised from the real send path**: added `FailOnceOnPayloadPollableChannel` (payload-scoped `send()` failure) with an explicit `PollableChannelConfiguration` retry template — `test_failed_send_of_single_message_on_target_channel_releases_only_that_message` exercises the actual outbound send including the channel-level send retries, not `preSend`.

The full dedicated inbound-adapter direction (owning batch receive) is left for the bulk-receive follow-up; with registration confinement the fast path is now opt-in by combined-channel declaration and no shared builder behavior changed.

packages/Ecotone/src/Messaging/Handler/Bridge/BatchForwardingBridge.php
dgafka: Fixed in 5b9b6ffb — the drain now uses `receiveWithTimeout` with a 1ms execution budget, so it does one no-wait fetch per drained message and stops on the first `null` instead of sitting out the channel's receive timeout inside the open transaction. Your reproduction is added as `test_single_message_is_forwarded_without_waiting_for_source_receive_timeout` (with `withReceiveTimeout(3000)`, the run completes well under 3000ms; before the fix it measured ~3025ms).

packages/Ecotone/src/Messaging/Handler/Bridge/BatchForwardingBridge.php
dgafka: Fixed in 5b9b6ffb — failed drained deliveries now honour the acknowledgement's configured `FinalFailureStrategy` exactly like the interceptor does: IGNORE rejects (discards), RELEASE releases, RESEND resends, and STOP releases then rethrows so the consumer stops and the transaction rolls back without loss. Added your reproduction as `test_drained_message_honours_ignore_final_failure_strategy` plus the sibling `test_drained_message_with_stop_failure_strategy_stops_consumer_without_message_loss`, with a new `AlwaysFailOnPayloadChannelInterceptor` fixture. While wiring the tests this surfaced a pre-existing gap: DBAL, SQS and Redis inbound channel adapter builders never passed `withFinalFailureStrategy` into their acknowledgement callbacks (AMQP did) — all three now forward the configured strategy.

packages/Ecotone/src/Messaging/Channel/Collector/MessageCollectorChannelInterceptor.php
dgafka: Fixed in ac9cdfd5 — the `collectorBypass` header is gone entirely. The intent is now expressed once at configuration time: `CollectorModule` skips registering the collector for channels that are non-first hops of a combined channel chain (the relay targets), so relay forwards are sent directly with no header to set, strip, or leak, while first-hop outbox channels and all other channels keep their collector (publish-side batching and discard-on-failure untouched). Your reproduction is added as `test_forwarded_message_does_not_carry_internal_collector_header`, running both with and without `GlobalPollableChannelConfiguration::withCollector`.

packages/Ecotone/src/Messaging/Config/MessagingSystemConfiguration.php
dgafka: Fixed in ac9cdfd5 (building on the confinement in 40d3d097) — batched forwarding now requires two things at configuration time: the channel must be the first hop of a combined channel chain, and its builder must implement the new `BatchForwardingSourceChannel` marker, which only the DBAL channel builder does. In-memory, Kafka, and any third-party builders get the plain bridge — one message per handled message. Your reproduction is added as `test_batched_forwarding_does_not_apply_to_in_memory_source_channels`, and the Kafka-source equivalent as `packages/Kafka/tests/Integration/CombinedChannelForwardingTest.php`. The `ForwardingBatchSizeAware` fallback this comment pointed at was already removed when the knob moved to `CombinedMessageChannel`.

